PR #119, opened 2013-09-07
Adding toggle for attachments on compose view

obsidev:
Hi,

In order to have more width in the composer, I added a toggle for the attachments.
I set this toggle reduced by default but it can be easily changed.
The impact of the modification is (logically) small, just changing the width of two elements.

Hope it could be useful.

Kind regards,
Jerome @ Obsidev


thomascube:
Deferred

PR #125, opened 2013-09-16
allow plugins to be loaded before session stuff

rodrigo-castillo:
Currently plugins are loaded at the end of rcmail::startup() - after the session has begun (among other things). Being able to load plugins earlier in the process allows plugins to touch more functionality, for my purposes, session handling.

This change allows, and sets, plugins to load earlier, just after rcmail::init() (though I imagine it's now possible to get it even earlier by putting the plugin loading within that method).

Because plugin loading and plugin init has been separated, existing plugins should continue to function normally because they're initialized in the same sequence as before.

Plugin authors will now be able to extend the rcube_plugin::__construct() method to make changes that should happen before that init(). Naturally if rcmail task or output are not set, they will not be accessible at that time. 

Conflicts:
    program/include/rcmail.php

### DIFF
--- a/plugins/managesieve/managesieve.php
+++ b/plugins/managesieve/managesieve.php
@@ -96,8 +96,8 @@ class managesieve extends rcube_plugin
         }
 
         // add 'Create filter' item to message menu
-        $this->api->add_content(html::tag('li', null, 
-            $this->api->output->button(array(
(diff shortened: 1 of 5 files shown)


alecpl:
A complete solution in PR #259.

PR #183, opened 2014-05-07
dont render plugin blocks if none selected

corbosman:
I noticed that even when I dont select my plugin in the preferences section, it was rendering the blocks. I could not stop it from doing that, because the current code expects there to be content in the section blocks even if your plugin is not showing. 

This PR prevents that. It should not change anything for existing plugins, but now you can modify your plugin to check $args['current'] instead of $args['section'] and only render when your plugin is selected.

### DIFF
--- a/program/steps/settings/func.inc
+++ b/program/steps/settings/func.inc
@@ -1235,7 +1235,7 @@ function rcmail_user_prefs($current = null)
             $data['blocks']['advanced'] = $adv;
         }
 
-        if (!$found)
+        if ($current && !$found)
             unset($sections[$idx]);
         else
             $sections[$idx]['blocks'] = $data['blocks'];


corbosman:
Hi Thomas, I looked into this again. I have created a branch pr in my plugin message_highlight to show the problem (https://github.com/corbosman/message_highlight/tree/pr)

This plugin adds its own section. The section has blocks. 

For efficiency reasons I only want to create the blocks if im the currently selected section. 

But the first time you go to url /?_task=settings, $current isnt set.  So the code goes through the sections, checks each section, and if it's the current section, it renders the blocks as well. This all works fine for the default sections.

Then it goes through the plugin provided sections, and there things to wrong. The plugin part goes through the blocks, and if there is no $block['content'] or $block['options'] it sets found to false. But we have not created those blocks on purpose in the plugin to prevent unnecessary code execution. Why render blocks if you're not showing them. 

When found is false, the code then removes the whole section, because it thinks the section is empty. 

Since this happens in an ajax environment, this then stays that way.

So my fix, although admittedly not that clean, says that if current is empty as well, dont remove any section blocks. 

Looking at that now, i guess thats not the right solution, because a plugin provided section could in fact really be empty. 

This is probably a better fix:  https://github.com/roundcube/roundcubemail/pull/258

That PR says, if there are no options blocks set for the section, THEN remove it. If it sees any options block, dont remove it.

Otherwise, I dont know how to fix this. It's not efficient to render section blocks if you're not the current section.

alecpl:
I don't think we need any fix here. It might look strange but there is a reason for that. Take two points into consideration:
1. Roundcube has dont_override option with which you can disable some configuration options.
2. We don't want to show sections that have disabled all options inside (there are no blocks or all blocks are empty).

So, that are the reasons why we check sections content when listing sections. That's is indeed a performance issue if your section is slow. But there's a simple solution for this. If your section is not the current one, just return 'content' => true.

corbosman:
Ah yes, makes sense. I'll close it.

PR #205, opened 2014-07-22
Add session_garbage plugin hook call

Takika:
Add session_garbage plugin hook call to the rcube_session class. This call needed if you want to collect the user authentication informations to a database. You can handle the session start and close via other hooks (login_after, login_failed, session_destroy) but if a session isn't closed properly you can't log it. With this change you get the chance to do other session_id related tasks before the session fully deleted.


alecpl:
Also, sessions can be deleted by gc.sh script.

PR #213, opened 2014-08-20
update rcube_plugin_api::exec_hook() to support new hooks that have been added mid-run

dhoffend:
This pull request fixes 2 things

1) fix tracking of the active_hook state

2) during the hook execution loop check for newly added hooks mid-run

Example: Plugin A dynamically loads plugin B during exec_hook:startup. Without this fix the startup hook of plugin B won't get executed anymore. (In this case kolab_auth plugin during kolab_auth_role_plugins execution)

Additional information to the while loop: 
- foreach won't see react on new added hooks
- using array pointers with while(next()) will not work cause adding new hooks will reset the current array pointer position.

I tried to keep it lightweight and hope this check doesn't affect too much of the performance.

What do yo think?

### DIFF
--- a/program/lib/Roundcube/rcube_plugin_api.php
+++ b/program/lib/Roundcube/rcube_plugin_api.php
@@ -46,7 +46,7 @@ class rcube_plugin_api
     protected $actionmap = array();
     protected $objectsmap = array();
     protected $template_contents = array();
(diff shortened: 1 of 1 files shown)


alecpl:
I fixed that in a much simpler way. See commit [0a5841702fdc].

PR #214, opened 2014-08-26
Don't hardcode the INBOX redirect

Takika:
If we remove this expression and didn't hardcode the INBOX redirect the app will return to the previously opened mbox when switch back to email from Preferences or Addressbook.


thomascube:
Deferred

PR #244, opened 2014-10-30
Add Dockerfile and Default Configuration Changes

alunduil:
Hey Guys,

I'm sure this is going to require some work before it's merged into master but here is an initial attempt at Dockerizing roundcube (Docker image can be found on hub.docker.com).  I added a Dockerfile that properly initializes the roundcube environment and runs apache when invoked. 

I also added the ability for default configuration values to come from environment variables.  The helper methods I've created probably belong somewhere else (I'll move them if necessary), but it seems to properly parse the values if they exist and does not impact current behavior.  It only changes behavior if these environment variables are set.

Please, let me know what changes if any are necessary for this to be merged properly and if it would be prudent to setup a roundcube hub.docker.com organization to keep the docker images up to date with the development of this project.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM debian:latest
+MAINTAINER Alex Brandt <alunduil@alunduil.com>
+
+EXPOSE 80 443
+
+RUN apt-get -qq update
+RUN apt-get install -qq apache2-mpm-event
+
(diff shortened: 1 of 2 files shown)


minad:
hi @alunduil, it seems that the .htaccess is ignored because 000-default.conf doesn't have the following:

~~~
	<Directory /var/www/public_html/>
		AllowOverride All
	</Directory>
~~~

alunduil:
@minad, you would be correct.  Good catch!  Mind submitting a pull request that injects that statement into the default configurations in the Dockerfile?  Should be just another sed expression.  Let me know via email if you need any assistance with it.

PR #246, opened 2014-11-04
Plugin password : Add custom function feature to validate new password

brenard:
Hello,

To implement my own method to validate new password submit by user, I had possibility to define a custom function to validate password. This new feature is control by two parameters : 
- $config['password_check_function'] : the name of the custom function. This function have to accept the new password as parameter and return a boolean to validate or invalidate the new password.
- $config['password_check_function_invalid_message'] : the error message in case of password invalidate by the function. If this parameter is not define, a standard error message is define in localized messages.

Thank's for this great plugin.


brenard:
Hello,

I still using this functionality and I upgrade my code to be compliant with actual password plugin. There is a chance that this pull request will be approved ?

Thank's for this great plugin.

alecpl:
A config file might not be the best place to put validation code. Maybe it would be better to add a plugin API hook, so you can write a separate plugin for validation code.

alecpl:
Extended the API. Now you can add a driver for password strength checking. Added also zxcvbn driver.

brenard:
Very nice feature ! Thank you !

PR #247, opened 2014-11-08
Support BCRYPT generated by Dovecot (doveadm pw).

ghost:
To generate BCRYPT hash with 'doveadm pw', you must use 'BLF-CRYPT' as scheme name, but you must replace it by '{CRYPT}' before storing the hash in SQL, so that Dovecot can verify it with system libc.


alecpl:
Anyway, you can use password_dovecotpw_with_method=false to remove the prefix and add it in the sql query using concat().

PR #254, opened 2014-12-31
bring %s variable support in setting smtp_server

arnisoph:
%s is available in defaut_host so providing %s variable support in smtp_server should be also possible since when you're using imap.domain.de you often have smtp.domain.de, too.

### DIFF
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -47,6 +47,7 @@ $config['default_host'] = 'localhost';
 // %t - hostname without the first part
 // %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
 // %z - IMAP domain (IMAP hostname without the first part)
+// %s - domain name after the '@' from e-mail address provided
 // For example %n = mail.domain.tld, %t = domain.tld
 $config['smtp_server'] = '';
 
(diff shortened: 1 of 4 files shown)


thomascube:
@alecpl what's your take on this?

alecpl:
Not bad, but there's small problem. %s might be resolved to different values on login and on smtp connect. This might be confusing. Anyway, %t, %d, %z is really not enough?

bechtoldt:
@alecpl 

Well there are people on this planet whose smtp server 

* isn't running on the same host/ ip address as the webmail server
* isn't running on the same host as the imap server

Since it's possible to have %s in ``$config['default_host']`` why not providing it in ``$config['smtp_server']``, too?

alecpl:
I still think this can be done with other variables, but if we really need this I'd go with different variable name, maybe using a capital S.

bechtoldt:
@alecpl Which variable could be used for the case explained above?

alecpl:
If your imap server is imap.domain.de and you want smtp_server to be smtp.domain.de set smtp_server to "smtp.%z".

bechtoldt:
@alecpl That seems to work. Looks like I misunderstood %z. 

Sorry for the confusion!

Closing PR since there isn't an use case left for this.

PR #257, opened 2015-01-05
session refactor

corbosman:
This is a rewrite of the rcube_session class. It introduces 3 new classes, rcube_session_db, rcube_session_memcache and rcube_session_native. 
- The native class is necessary to inherit functionality from the abstract class, and to make the code not have to go through all kinds of ugly hoops to exclude the native php sessions. These exceptions are now maintained in the native class.
- I had to move the plugin initialisation up in the execution order. Originally plugins were initialised after session_init, giving you the dummy plugin api class. Since I implemented a plugin hook for session storage, this had to be moved to before session_init. I am not sure if there are any large issues with this, but a brief check didnt seem to show any. 
- I made the write save handler a bit different from the rest. Some of it is handled in the abstract class. The original code had all kinds of calculations to see if a session should be written, or if we could use a cached version. I obviously wanted to keep that logic. But I belief some of that could be moved to the abstract class, making the implementation classes a little cleaner and allowing the abstract class to make some of these decisions.  But, i did leave some of it in the implementation class because the db and memcache implementations do things slightly different. I did not want to mess with those decision processes, so I left them in the implementation classes. Maybe sometime in the future we can unify these calculations and make the implementation classes cleaner. 
- I wrote a redis implementation plugin that can be found here: https://github.com/corbosman/roundcube_redis_session .  With your permission I would like to bring that into core with another PR once this has been (hopefully) approved. Redis is a much better choice as a session cache than memcache because of two reasons.
  1. memcache is not persistent. A restart of the memcache server will kill all your sessions. This makes updates of the memcache server software a huge pain.
  2. memcache can actually delete session on its own. It allows for key deletion whenever it feels like it really. This makes memcache not an ideal choice for sessions. 

Redis has none of these problems. 

I did quite a bit of testing to make sure everything works with all the implementation classes, but I can not rule out that I made a mistake. So please review carefully.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -374,9 +374,14 @@ $config['session_auth_name'] = null;
 // Session path. Defaults to PHP session.cookie_path setting.
 $config['session_path'] = null;
 
-// Backend to use for session storage. Can either be 'db' (default), 'memcache' or 'php'
+// Backend to use for session storage. Can either be 'db' (default), 'redis', 'memcache', or 'php'
+//
 // If set to 'memcache', a list of servers need to be specified in 'memcache_hosts'
 // Make sure the Memcache extension (http://pecl.php.net/package/memcache) version >= 2.0.0 is installed
+//
+// If set to 'redis', a server needs to be specified in 'redis_hosts'
+// Make sure the Redis extension (http://pecl.php.net/package/redis) version >= 2.0.0 is installed
+//
 // Setting this value to 'php' will use the default session save handler configured in PHP
 $config['session_storage'] = 'db';
 
@@ -396,6 +401,13 @@ $config['memcache_timeout'] = 1;
 // Setting this parameter to -1 disables automatic retry.
 // See http://php.net/manual/en/memcache.addserver.php
 $config['memcache_retry_interval'] = 15;
+
+// use this for accessing redis
+// currently only one host is supported. cluster support may come in a future release.
+// you can pass 4 fields, host, port, database and password.
+// unset fields will be set to the default values host=127.0.0.1, port=6379, database=0, password=  (empty)
+
+$config['redis_hosts'] = null; // e.g. array( 'localhost:6379' );  array( '192.168.1.1:6379:1:secret' );
 
 // check client IP in session authorization
 $config['ip_check'] = false;
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -522,9 +522,12 @@ class rcube
         ini_set('session.cookie_httponly', 1);
 
         // use database for storing session data
-        $this->session = new rcube_session($this->get_dbh(), $this->config);
+        $storage = $this->config->get('session_storage', 'db');
+        $this->session = $this->get_session($storage);
 
+        // register default gc handler
         $this->session->register_gc_handler(array($this, 'gc'));
+
         $this->session->set_secret($this->config->get('des_key') . dirname($_SERVER['SCRIPT_NAME']));
         $this->session->set_ip_check($this->config->get('ip_check'));
 
@@ -534,8 +537,30 @@ class rcube
 
         // start PHP session (if not in CLI mode)
         if ($_SERVER['REMOTE_ADDR']) {
-            $this->session->start();
+            $this->session->start($this->config);
         }
+    }
+
+    /**
+     * get an rcube_session instance
+     *
+     * @return rcube_session
+     */
+    private function get_session($storage)
+    {
+        // class name for this storage
+        $class = "rcube_session_" . $storage;
+
+        // try to instantiate class
+        if(class_exists($class)) {
+            return new $class();
+        }
+
+        // no storage found, raise error
+        rcube::raise_error(array('code' => 604, 'type' => 'session',
+                               'line' => __LINE__, 'file' => __FILE__,
+                               'message' => "Failed to find session driver. Check session_storage config option"),
+                           true, true);
     }
 
 
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -521,48 +521,17 @@ class rcube
         ini_set('session.use_only_cookies', 1);
         ini_set('session.cookie_httponly', 1);
 
-        // use database for storing session data
-        $storage = $this->config->get('session_storage', 'db');
-        $this->session = $this->get_session($storage);
+        // get storage driver from config
+        // $storage = $this->config->get('session_storage', 'db');
 
-        // register default gc handler
-        $this->session->register_gc_handler(array($this, 'gc'));
-
-        $this->session->set_secret($this->config->get('des_key') . dirname($_SERVER['SCRIPT_NAME']));
-        $this->session->set_ip_check($this->config->get('ip_check'));
-
-        if ($this->config->get('session_auth_name')) {
-            $this->session->set_cookiename($this->config->get('session_auth_name'));
-        }
+        // get session driver instance
+        $this->session = rcube_session::factory($this->config);
 
         // start PHP session (if not in CLI mode)
         if ($_SERVER['REMOTE_ADDR']) {
-            $this->session->start($this->config);
+            $this->session->start();
         }
     }
-
-    /**
-     * get an rcube_session instance
-     *
-     * @return rcube_session
-     */
-    private function get_session($storage)
-    {
-        // class name for this storage
-        $class = "rcube_session_" . $storage;
-
-        // try to instantiate class
-        if(class_exists($class)) {
-            return new $class();
-        }
-
-        // no storage found, raise error
-        rcube::raise_error(array('code' => 604, 'type' => 'session',
-                               'line' => __LINE__, 'file' => __FILE__,
-                               'message' => "Failed to find session driver. Check session_storage config option"),
-                           true, true);
-    }
-
 
     /**
      * Garbage collector - cache/temp cleaner
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -520,10 +520,7 @@ class rcube
         ini_set('session.use_cookies', 1);
         ini_set('session.use_only_cookies', 1);
         ini_set('session.cookie_httponly', 1);
-
-        // get storage driver from config
-        // $storage = $this->config->get('session_storage', 'db');
-
+        
         // get session driver instance
         $this->session = rcube_session::factory($this->config);
 
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -520,7 +520,7 @@ class rcube
         ini_set('session.use_cookies', 1);
         ini_set('session.use_only_cookies', 1);
         ini_set('session.cookie_httponly', 1);
-        
+
         // get session driver instance
         $this->session = rcube_session::factory($this->config);
 
--- a/program/lib/Roundcube/rcube_session.php
+++ b/program/lib/Roundcube/rcube_session.php
@@ -15,6 +15,7 @@
  +-----------------------------------------------------------------------+
  | Author: Thomas Bruederli <roundcube@gmail.com>                        |
  | Author: Aleksander Machniak <alec@alec.pl>                            |
+ | Author: Cor Bosman <cor@roundcu.be>                            |
  +-----------------------------------------------------------------------+
 */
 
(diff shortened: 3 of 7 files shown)


corbosman:
Unfortunately moving plugins higher up in the execution path does cause problems. I dont immediately see a way to fix this. The way things are initialised right now im not sure we could have plugin provided sessions.

alecpl:
There is a request for that issue, but is not finished - https://github.com/roundcube/roundcubemail/pull/125 - you can however review the code changes and solutions discussed there.

corbosman:
Ok, i'll have a look at those. Thanks for pointing that out.

corbosman:
I have created a PR roundcube/roundcubemail#259 which hopefully solves the problem of loading plugins earlier in startup. I need this code for this PR to function, so I wouldnt mind just including it in this PR. But I created a separate PR so you can see the changes in isolation. 

That is, if you agree with the chosen solution at all of course :)

corbosman:
Alec, how would you want me to proceed? As far as I can tell in testing redis sessions now work using PR259. Other plugins also seem to function normally as that PR should be fully backwards compatible.  

This PR would not apply clean anymore after merging 259. 

Do you want me to combine 259 and this one into one PR? Or clean this one up after you hopefully merge 259?

alecpl:
I think you should remove all plugins-related changes from this PR. Please, give us some time to review and merge these requests. We can't do this now - after RC release, anyway. We have to branch off the 1.1 first.

program/lib/Roundcube/rcube.php
```diff
@@ -531,10 +533,40 @@ public function session_init()
...
+            return new rcube_session_db($this->get_dbh());
+        }
+        elseif ($storage == 'memcache') {
+            return new rcube_session_memcache();
+        }
+        elseif ($storage == 'php') {
+            return new rcube_session_native();
+        }
+        else {
+            $plugin = $this->plugins->exec_hook('session_get', array());
```
alecpl: I don't see a need for this hook. Plugins can just set the session_storage option. The code above should just create an object of rcube_session_<session_storage> class. Constructor should be unified to accept the same arguments. That would make it much simpler.

corbosman:
I agree with you that i can just try to instantiate rcube_session_<session_storage>. I will change that.

For now it looks like the only non-unified argument is the database object. You dont need that in any other storage class probably. 

So are you saying we should just pass the database object anyways, also to other storage classes, or should rcube_session_db() instantiate its own database connection just like the memcache and redis driver do their own connection as well?  

I guess we could do an rcube::get_instance()->get_dbh() inside the db driver, but that is probably less efficient than passing an existing object to the storage class. 

Or maybe we just pass $this to the storage classes? 

It's not easy for an outsider to code these kinds of refactors because things are so interwoven :)

alecpl:
Get rid of the argument and set $this->db in rcube_session_db constructor by calling get_dbh(). BTW, rename rcube_session_native to rcube_session_php.

corbosman:
Yeah, i know I have to rename it to php if I load it like that :)  I just thought it looked silly :) Most frameworks use the word 'native' as well. But in this case we have no choice so i'll make it _php ofcourse. 

Ok, no problem, easy changes.

program/lib/Roundcube/rcube.php
```diff
@@ -519,8 +519,10 @@ public function session_init()
         ini_set('session.cookie_httponly', 1);
 
         // use database for storing session data
-        $this->session = new rcube_session($this->get_dbh(), $this->config);
+        $storage = $this->config->get('session_storage', 'db');
+        $this->session = $this->get_session($storage);
 
+        $this->session->register_session_handler();
```
alecpl: This is very session-driver specific. I think this should be done in driver classes.

corbosman:
Why is it session driver specific?  For session_set_save_handler() you have to always provide a set of functions. Why not unify the function names by forcing abstract function implementations? It also allows you to make some decisions about session management in core (rcube_session.php). Else you're going to get a lot of duplication in each driver. They're all going to have to register handlers, and they're all going to have to make some decisions about using cached data. Why not pull that out of the drivers and have the abstract function handle it? Have the drivers only worry about the actual IO of the data. I was already not happy that the write functions in the driver classes has some logic in it. 

Im also looking at this from the viewpoint of plugins providing session drivers.  The less they need to know about cached sessions the better right?

alecpl:
We should just do it in session-driver class constructor not in rcube class. Also, native session driver does not need them ;)

corbosman:
Ah ok, i misunderstood. I thought you meant the driver implementation classes.  So ok, move it to construct. I solved the issue of native session not needing them by providing that class with an empty function. is that ok? Else the driver class needs knowledge about the storage driver setting.

corbosman:
Id also like your input about something else. 

Im building a plugin for redis, mainly to test the plugin functionality as well. If I require_once() the redis implementation class (rcube_session_redis), at the top of the  plugin, rcube.php can actually find that class (if I use class_exists() and not file_exists() or something).  Is that ok? Or should rcube_php never itself instantiate plugin provided implementation classes and let the plugin itself return the instantiated object?  In that case I need to prevent rcube.php from finding plugin provided classes, and do a file_exists or something..

corbosman:
Ah! I remember now why I moved  $this->session->register_session_handler(); out of _construct().  Else the driver implementation classes need to always call parent::__construct(). Fine for roundcube's own classes, but I think it's not nice for plugins to have to do that.  If you disagree, I will have all driver implementation classes call parent..

alecpl:
Just do this in driver class constructor. As for the redis plugin, it's fine to just require the class file and allow rcube to detect the class existence.

corbosman:
ok, just to be sure. When you say "Just do this in driver class constructor" you mean implementation classes must call parent::__construct() right?

alecpl:
No, they should call $this->register_session_handler().

corbosman:
Oh ok.. it's a matter of terminology confusion. :) Ok, i understand now.

corbosman:
Sorry to keep bothering you, although I really appreciate the help. 

If I let rcube.php instantiate plugin created sessions drivers, I see no easy way to pass plugin config settings to that driver to pass host/port/password/whatever config options. 

I cant use the constructor, because the config options are not loaded for plugins. 

This is why I used a hook, but if I use a hook, I cant just let rcube.php find the class. 

What do you think would be the best way to tackle this?

alecpl:
Do $this->load_config() and require in onload() method of the plugin you discussed in #259.

corbosman:
It doesnt really matter where you require right? 

If we use load_config() in the plugin, I believe we need to pass $config to the rcube_session driver constructor. Correct?

alecpl:
No. You don't need to pass any arguments. Configuration can be read at any moment from rcube::get_instance()->config.

corbosman:
ugh, i keep forgetting you can get an instance. thanks for your patience.

corbosman:
I refactored PR based on the comments here. I moved the plugin init back to where it originally was, which means this PR alone doesnt function properly for session plugins. 

This needs PR#259 which is meant to move plugin init to before session init.

If you prefer a combined PR, let me know. It is a little difficult maintaining these 2 PRs separately while testing if it all works :)

corbosman:
I added that last commit to make it separate from the onload PR. That commit mistakenly came along.

corbosman:
And this is what the redis plugin looks like now using onload:  https://github.com/corbosman/roundcube_redis_session

Works great.

I'll wait and see when you guys get to this, and i'll perhaps just run our 1.0 branch with    these patches applied in the meantime so we can move to redis.  DB sessions are killing us.

config/defaults.inc.php
```diff
@@ -384,6 +389,18 @@
...
+// currently only one host named 'default' is supported. cluster support may come in a future release.
+// $config['redis_hosts'] = array(
+//     'default' => array(
+//         'host'     => '127.0.0.1',         // host to connect to
+//         'port'     => 6379,                // port to connect to
+//         'database' => 0,                   // default database, should be an integer
+//         'password' => false,               // password when using auth
+//     )
+// );
+$config['redis_hosts'] = null;
```
alecpl: It might be convenient to also support format compatible with memcache_hosts, i.e. a list of host specifications. If database and password aren't set we use defaults. Note: password=null, database=null, should be handled correctly in rcube_session_redis, it isn't now.

corbosman: How about we just support 1 format just like memcache, and any missing values will be a default?  This may be what you meant, but since you said 'also support' it suggests you want to support both formats. Im fine with just simply supporting the format below. Just let me know if just one is fine and i'll do that.

```
$config['redis_hosts'] = array( 'localhost:7379:0:foobar, '192.168.1.12:11211:1:foobar2')
```

For now you cant give multiple hosts because we dont support clustering yet. php-redis does not have cluster support. If we'd want cluster support we'd have to bundle https://github.com/nrk/predis which is php 5.3+. But php-redis already has a branch with cluster development so im assuming that'll come soon.

corbosman:
I just changed it so it's like memcache_hosts. I assume thats what you meant.

corbosman:
Shall I rebase this PR into 1 commit, or are there still some modifications necessary?

corbosman:
Hi guys, I made an error trying to fix some merge conflicts after recent master changes. Forgot to rebase so it added a huge merge commit.

I fixed it by re-playing the patch on a fresh master, and forcing a push. Hope things worked out ok.

program/lib/Roundcube/rcube.php
```diff
@@ -534,8 +537,30 @@ public function session_init()
...
+     * @return rcube_session
+     */
+    private function get_session($storage)
+    {
+        // class name for this storage
+        $class = "rcube_session_" . $storage;
+
+        // try to instantiate class
+        if(class_exists($class)) {
+            return new $class();
```
thomascube: I'd still like to pass `$this->config` to the constructor rather than to `start()` later on.

program/lib/Roundcube/rcube.php
```diff
@@ -534,8 +537,30 @@ public function session_init()
...
+
+    /**
+     * get an rcube_session instance
+     *
+     * @return rcube_session
+     */
+    private function get_session($storage)
+    {
+        // class name for this storage
+        $class = "rcube_session_" . $storage;
```
thomascube: What about moving this `get_session()` into a `rcube_session::factory()` method as we have it for `rcube_db`?

program/lib/Roundcube/rcube.php
corbosman: Every driver class has a constructor as well. This means I have 2 options. 

1. Every driver class has to pass $config to a parent::__construct() call, and the abstract class __construct($config) sets $this->config.   Personally, I think this is a bit ugly and means every driver class has to know to pass $config to the parent class. 

2. I rename __construct() in all the driver classes to init(), and call init() from rcube.php. This is probably a better option, especially with plugin provided drivers in mind. They know they'll have to provide an init() function to setup their driver. 

Another option we have, is that we call start() with the config options directly, just like the other method calls before it. Then start() doesnt need to know about $config at all, and we dont need to pass $config. Thats probably how I should have done it, seeing as the calls before it do it that way. 

But, the redis driver does need $config as well. Right now, it gets the config instance itself, which is an artifact from it being a plugin. I could keep it like that, so we dont need to pass $config.

Looking at the code, we could even take this a step further. In rcube.php, we do some calls to $this->session (like set_secret, set_ip_check etc). We could move all of this to the abstract rcube_session.php constructor. It doesnt seem like they're used anywhere else in the code. Combine this with renaming __construct in the driver classes to init().

Which route would you like me to take?

program/lib/Roundcube/rcube.php
corbosman: No problem. Done locally, just waiting for your input on the previous comment.

program/lib/Roundcube/rcube.php
thomascube: I don't see 1. as a problem. This is just how inheritance works. And the rcube_session subclasses may want to pick some data from $config as well. If we make it ->init($config), the situation is more or less the same. A derivate may override init() and in that case also needs to call parent::init($config). So what's the difference?

But I agree to move these set_* classs directly into the rcube_session base class.

corbosman: The abstract base class wouldnt have an implementation init() method, just an abstract init method. 

But im fine with just calling parent::__construct($config) from each driver. So i'll do that.

corbosman:
Thomas, I changed it to use a factory method and pass $config via constructor.

Let me know what you think, and if things are ok I can rebase into one commit.

thomascube:
Looks good! I'll give it a test ride anytime soon.

PR #259, opened 2015-01-06
Plugin preload

corbosman:
I think this may be a solution for allowing plugins to load before sessions.

Ive introduced an rcube_plugins variable name $preload, which tells the plugin code to load this plugin early. Default is false, and then it gets loaded as it is now. If you set $preload in your plugin, you accept that all kinds of variables are not yet initialised, including $task and $app->output. 

I also had to account for $force, so ive also introduced a variable named $initialised, which tells the second phase of plugin init which plugins were already initialised in round 1. 

I think my logic with regards to noajax/noframe etc is ok, but please check. 

This PR would replace PR roundcube/roundcubemail#125

Since I need this code in my PR roundcube/roundcubemail#257 please let me know if you prefer this PR separate or if I should add it there since it's kind of related.

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -93,6 +93,10 @@ class rcmail extends rcube
             $this->filename = $basename;
         }
 
+        // load all configured plugins
+        $this->plugins->load_plugins((array)$this->config->get('plugins', array()),
+                                     array('filesystem_attachments', 'jqueryui'));
+
         // start session
         $this->session_init();
 
@@ -124,10 +128,8 @@ class rcmail extends rcube
             $GLOBALS['OUTPUT'] = $this->load_gui(!empty($_REQUEST['_framed']));
         }
 
-        // load plugins
+        // run init method on all the plugins
         $this->plugins->init($this, $this->task);
-        $this->plugins->load_plugins((array)$this->config->get('plugins', array()),
-            array('filesystem_attachments', 'jqueryui'));
     }
 
     /**
--- a/program/lib/Roundcube/rcube_plugin_api.php
+++ b/program/lib/Roundcube/rcube_plugin_api.php
@@ -34,6 +34,8 @@ class rcube_plugin_api
     public $dir;
     public $url = 'plugins/';
     public $task = '';
+    public $initialized = false;
+
     public $output;
     public $handlers              = array();
     public $allowed_prefs         = array();
@@ -117,12 +119,20 @@ class rcube_plugin_api
     {
         $this->task   = $task;
         $this->output = $app->output;
-
         // register an internal hook
         $this->register_hook('template_container', array($this, 'template_container_hook'));
-
         // maybe also register a shudown function which triggers
         // shutdown functions of all plugin objects
+
+        foreach ($this->plugins as $plugin) {
+            // ... task, request type and framed mode
+            if (!$this->filter($plugin)) {
+                $plugin->init();
+            }
+        }
+
+        // we have finished initializing all plugins
+        $this->initialized = true;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


corbosman:
I see what you mean now.. :)

I think we need to unfortunately add a parameter to load_config.. 

something like:

```
public function load_plugin($plugin_name,  $force = false, $require = true, $delay=false);
```

And load->plugins() then uses $delay=true;  

I dont see another way without breaking the load_plugin() api call..

thomascube:
Hmm what about:
```php
public function load_plugin(...)
{
    ...
    if (method_exists($plugin, 'onload')) {
        $plugin->onload();
    }
    if ($this->initialized) {
        $plugin->init();
    }
    $this->plugins[$plugin_name] = $plugin;
   ...
}

public function init_plugins(...)
{
    ...
    $this->initialized = true;
}
```

corbosman:
Ohh, I see now what you mean. Yes of course, you said move it to plugin_api. I was still thinking in my head about per-plugin status. 

You mean, you keep an initialized variable just to remember we did init_plugins and all further plugin loads should call init itself. Yup, no problem.

corbosman:
I have implemented the onload method check, and also refactored that noajax/noframe check into its own function to prevent code duplication and to make it more readable.

thomascube:
Thanks a lot! One last thing ;-)

`rcube_plugin_api::init()` became almost empty while `init_plugins()` has become pretty much of a replacement for it. So why not restore `init()` to what it was before and just add the plugin initialization loop to it. Then call `init()` after `load_plugins()`.

Just to keep the change as simple as possible and existing methods and signatures intact.

corbosman:
Sure no problem :)

corbosman:
Ok, that should do it. Seems to work :)

program/include/rcmail.php
```diff
@@ -93,6 +93,10 @@ protected function startup()
             $this->filename = $basename;
         }
 
+        // load all configured plugins
+        $this->plugins->load_plugins((array)$this->config->get('plugins', array()),
```
alecpl: You moved this code before $task is being set. So, filtering by task in load_plugin() will not work as expected.

program/lib/Roundcube/rcube_plugin.php
```diff
@@ -59,7 +59,7 @@
      * @var boolean
      */
     public $noframe = false;
-
+
```
alecpl: Please, clean this up.

program/lib/Roundcube/rcube_plugin_api.php
```diff
@@ -235,6 +255,42 @@ public function load_plugin($plugin_name, $force = false, $require = true)
...
+
+    /**
+     * check if we should prevent this plugin from initialising
+     *
+     * @param $plugin
+     * @return bool
+     */
+    private function filter($plugin)
+    {
+        return (    ($plugin->noajax  && !(is_object($this->output) && $this->output->type == 'html') )
```
alecpl: I'd like to see also the task filter moved here.

program/include/rcmail.php
corbosman: Im not sure what you mean. Isnt that the whole point? 

Load plugins is called early, but with the understanding that if you use an onload() method it wont have access to $task and other variables.  Then initialisation of the plugins, where task is checked, is defered until a later step. So $plugin->init(); in load_plugin is actually not called during the load_plugins() phase of the plugin system initialisation. 

Then later in rcmail.php $this->plugins->init($this, $this->task); is called where plugins have access to $task. 

I probably just dont understand your comment, so please elaborate which code path is going to fail and i'll be happy to fix it if possible.

corbosman:
I removed the spurious space (phpstorm added that somehow), moved $task check to filter. Please let me know what you meant with the first comment.

program/include/rcmail.php
alecpl: OK, my mistake. The code is not simple to follow here. Maybe too much magic in it. Still init_plugins() method can be removed.

corbosman: Good catch, didnt notice that unnecessary method. Committed. 

I totally agree with you the code is not easy to follow, but there are so many dependancies that I couldnt figure out a better way to do this without having to continuously modify more and more of the code.

alecpl:
I think this is now ready to merge. Can we have this PR rebased as one commit (instead of eight)? Also, there are still some small CS issues like no space after 'foreach' or redundant spaces in 'if'.

corbosman:
I have rebased it into one commit, and hopefully found all the space issues.

PR #260, opened 2015-01-09
use one line default configuration stub in Dockerfile

alunduil:
Docker doesn't process the -e option on echo properly and mangles the
default configuration file resulting in a white screen when running
roundcube.


alecpl:
I saw examples that work with -e, but there are double quotes used. Could you try that?

alunduil:
Would that be preferred to the one line stub configuration file?  The problem arose due to one build getting the -e in the file itself and I'd rather just avoid that being possible.  I can use double quotes if it's preferred just let me know.

macropin:
-e is not required. The newlines are expanded without it.

macropin:
btw, I fixed this in #292

PR #261, opened 2015-01-09
modify pear dependency net_smtp2 to net_smtp in Dockerfile

doublerr:
Currently installing `Net_SMTP2` results with the following error:

``` bash
[09-Jan-2015 21:15:14 UTC] PHP Fatal error:  Class 'Net_SMTP' not found in /var/www/program/lib/Roundcube/rcube_smtp.php on line 111
```

Switching back to Net_SMTP fixes this issue


alecpl:
You should also fix Auth_SASL, I suppose.

doublerr:
@alecpl auth_sasl2-beta has been changed to auth_sasl

PR #263, opened 2015-01-29
Fix unregister midflight

dhoffend:
1) Renumber handlers after unregister_hook() otherwise the exec_hook() for-loop can fail.

2) Postpone real unset of hooks mid-exection to a later point.

If a hook is beeing unregistered during it's exection, the real unset() should be postponed until the exec_hook() command is done, otherwise it might screw up the for-loop

This pullrequest is a follow up on #213.


dhoffend:
If you think that the 2nd commit is too complex, please at least take the first one (which is a real bug)

alecpl:
Good point about array keys issue, fixed in 20740a9650. I think removing the handler which is already executed would be so rare case that we should not care and keep the code simpler.

PR #265, opened 2015-02-11
Allow the timeout and retry interval for memcached to be configured

kanarip:
As per the title, no change to default behaviour.


thomascube:
Also described memcache-related config options in the defaults.inc.php for reference [de3fc1a]

PR #266, opened 2015-02-14
Use 'skins' rather than 'themes' for the sake of consistency

kanarip:
The paths used by the jqueryui plugin are fixed with 'themes' rather than 'skins', whereas the rest of roundcubemail consistently refers to its 'skins' as 'skins' not 'themes'.

### DIFF
--- a/plugins/jqueryui/README
+++ b/plugins/jqueryui/README
@@ -24,7 +24,7 @@ Some jquery-ui modules provide localization. One example is the datepicker modul
 If you want to load localization for a specific module, then set up config.inc.php.
 Check the config.inc.php.dist file on how to set this up for the datepicker module.
 
-As of version 1.8.6 this plugin also supports other themes. If you're a theme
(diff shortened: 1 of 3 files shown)


thomascube:
This is because jquery UI calls them themes and what the plugin does is to map jquery UI themes to roundcube skins. Please note that these are not necessarily the same! Some default jquery UI themes (e.g. redmond) can be mapped to be used in various Roundcube skins. Therefore I'd like to stay with the current naming conventions.

kanarip:
The only occurrences of the literal string "theme" though is only in roundcube's code, or at least so it seems. Would you not rather stay consistent within the roundcube codebase rather than diverge whenever a plugin calls a skin by a different term?

thomascube:
Changing this would affect existing installations/configurations as well as custom skins. I hesitate a little bit to deliberately break these just for the sake of better naming.

dereks:

```
git fetch upstream
git rebase upstream/master
```

PR #267, opened 2015-02-23
Make display_message() always work

dtaylor:
In some cases, say with custom plugins, the message container DOM element may not be present. When this happens the messages simply disappear - no JS errors, etc. This patch ensures that the container is always setup. It should work for both framed and unframed cases.

It works for our $work installation including custom plugins, but I haven't tested it with a stock installation. The concept should be helpful even if my newbie PHP skills are lacking.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -557,8 +557,10 @@ function rcube_webmail()
     this.env.lastrefresh = new Date();
 
     // show message
-    if (this.pending_message)
+    if (this.pending_message) {
+      this.init_messagestack();
       this.display_message.apply(this, this.pending_message);
+    }
 
     // init treelist widget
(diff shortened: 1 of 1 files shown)


thomascube:
The message container is defined by the skin, also its type/tag and ID. Adding `<div id="messagestack">` to app.js breaks the separation of view and application logic and would break things with other skins not naming the container #messagestack.

dtaylor:
Thank you for letting me know - I'm very new to PHP and Roundcube. I've already fixed our non-iframe plugin but I'm still trying to figure out where to fix the skin/plugin template for our iframe-based plugin.

dtaylor:
Scratch my last comment. I just figured it out and now everything works perfectly. Thank you for the tip!

PR #272, opened 2015-04-06
plugins/password: Add config option for password schemes to SQL driver

paulmenzel:
Often the password scheme is prepended to the hash. Dovecot is doing
that and the VirtualMailManager (VMM).

Add a boolean option to enable prepending the scheme.

The `doveadm pw` cannot be used, especially if Roundcube is not
installed on the mail server. Actually `password_dovecotpw_with_method`
could be renamed, if that is allowed.

### DIFF
--- a/plugins/password/config.inc.php.dist
+++ b/plugins/password/config.inc.php.dist
@@ -72,6 +72,9 @@ $config['password_query'] = 'SELECT update_passwd(%c, %u)';
 // your operating system supports the other hash functions.
 $config['password_crypt_hash'] = 'md5';
 
+// Enables use of password with crypt method prefix in %D, e.g. {MD5}$1$LUiMYWqx$fEkg/ggr/L6Mb2X7be4i1/
+$config['password_with_scheme'] = false;
+
 // By default domains in variables are using unicode.
 // Enable this option to use punycoded names
 $config['password_idn_ascii'] = false;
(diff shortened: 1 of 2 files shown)


paulmenzel:
Please don’t merge this yet. This is more a request for comments.

For example, I’m unsure what scheme should be used for *DES*.

alecpl:
You can just add (concat) the scheme string in password_query. There's no need for an additional option.

paulmenzel:
@alecpl, thank you for your reply. Unfortunately, I do not see how this can easily be done.

Could youl please tell me, how I can easily add `{SHA512-CRYPT}` before the `%c`?

    $rcmail_config['password_query'] = 'UPDATE users SET passwd=%c WHERE local_part=%l and gid=(SELECT gid FROM domain_name WHERE domainname=%d LIMIT 1);';

As the macros are escaped, I don’t see how I can easily prepend the string.

alecpl:
For mysql: passwd=CONCAT('{SHA512-CRYPT}', %c), and because the configuration string above is using single quotes you'll need to change to double quotes or escape single-quotes inside concat().

paulmenzel:
@alecpl, thank you. PostgreSQL is used here, so looking at [String Functions and Operators](https://www.postgresql.org/docs/current/static/functions-string.html) I first saw the `||` and finally got that to work, after having to find out how quoting and escaping in PHP works.

    'UPDATE users SET passwd=\'{SHA512-CRYPT}\' ||' . '%c WHERE local_part=%l and gid=(SELECT gid FROM domain_name WHERE domainname=%d LIMIT 1);';

Only now I discovered, that PostgreSQL also provides the function `concat()`.

All in all, this is not easy for a non-PHP and non-SQL person. I don’t know what Roundcube’s target audience is, but I thought that such things should be easily configurable.

PR #275, opened 2015-04-29
Update index.php

yainel:
To enable it again, set $config['enable_installer'] = true; but in RCUBE_CONFIG_DIR/defaults.inc.php


thomascube:
Nope, the message is correct. `defaults.inc.php` shall not be changed at any time!

From http://trac.roundcube.net/wiki/Howto_Config:

    If the local config file doesn't already contain one of the listed options
    simply copy it from the config/defaults.inc.php file.

PR #276, opened 2015-05-07
Update bootstrap.php (config setting fix)

thecodeassassin:
Fixed config setting. $array += doesn't actually do anything.


alecpl:
Where did you find it does not work? Works for me.

thecodeassassin:
Ahh sorry i did it a little bit wrong. But if you already defined suhosin.session.encrypt it will NOT add it. With array_merge it will :)

thomascube:
How would it possibly be already defined? `$config` is just set a few lines above.

PR #278, opened 2015-05-27
Update fr_FR.inc

jjmz:
Obviously inverted French translation for vacation status.


alecpl:
Please, use Transifex service at https://www.transifex.com/roundcube/roundcube-webmail/

PR #284, opened 2015-06-16
full path to cat needs to specified in order for this to work

prashantjois:
Running the container with the default site settings will fail because apache can't find the path to `cat`.  Here we specify it explicitly


alunduil:
@prashantjois can you rebase your changes off of current master?  Otherwise, the change looks good and I'll +1 it for merging ASAP.

prashantjois:
@alunduil rebased

PR #286, opened 2015-07-09
Fix wash_style() to accept style declarations with new lines.

dsoares:
Some(?) email clients wrap long lines in HTML messages. With some _bad luck_, the wrapping occurs in the middle of a style declaration. This is an example of a style declaration after wrapping:

``` html
  (...)<div style="padding: 0px
            20px;border:1px solid #000;">(...)
```

With the `s` modifier in `preg_match()`, the dot `.` in the regular expression accepts the `\n` and the style declaration remains correct.


alecpl:
Thanks, but I think it will be better to remove these redundant white-space characters. Fixed in f4c512336d.

PR #290, opened 2015-08-07
Fix group creation to work with Net_LDAP3.

eehakkin:
Net_LDAP3::add_entry does not pass empty attributes to ldap_add, thus
let's replace an empty attribute with a non-empty array containing
an empty string as that is supported by ldap_add, too.

### DIFF
--- a/program/lib/Roundcube/rcube_ldap.php
+++ b/program/lib/Roundcube/rcube_ldap.php
@@ -1947,7 +1947,7 @@ class rcube_ldap extends rcube_addressbook
         $new_entry   = array(
             'objectClass' => $this->prop['groups']['object_classes'],
             $name_attr    => $group_name,
-            $member_attr  => '',
+            $member_attr  => array(''),
         );
 
         if (!$this->ldap->add_entry($new_dn, $new_entry)) {


alecpl:
Hmm... Why it is a problem?

eehakkin:
It is a problem because it breaks group creation.

A typically configuration for LDAP contact groups, I suppose, is to use
 * array('groupOfNames') as object_classes,
 * 'cn' as name_attr and
 * 'member' as member_attr

thus $new_entry will be

    array('objectClass' => array('groupOfNames'), 'cn' => $group_name, 'member' => '')

without this patch.

Then Net_LDAP3::add_entry changes that to

    array('objectClass' => array('groupOfNames'), 'cn' => $group_name)

Then ldap_add sends an add operation to an LDAP server, which refuses to complete the operation as the groupOfNames entry is missing a mandatory member attribute.

With this patch, the the groupOfNames entry will have an empty member attribute (instead of none at all). So this patch restores the behaviour which RoundCube had when it used Net_LDAP2.

alecpl:
Indeed, per https://tools.ietf.org/html/rfc4519#section-3.5 the member attribute is required. I didn't notice that because the ldap server I'm using has this class redefined in a way that member attribute is not required.

Anyway, your fix is not good. On my server setting 'uniquemember' attribute to an empty array or array with an empty element on group creation gives "Invalid syntax" or "Protocol error". So, maybe we need some config option.

dereks:

```
git fetch upstream
git rebase upstream/master
```

PR #291, opened 2015-08-20
Add support for :from in managesieve/vacation plugin

mlocati:


alecpl:
Nice, thank you. Maybe you would like to work also on http://trac.roundcube.net/ticket/1489221 ?

mlocati:
I have dovecot 2.2.9 (standard for Ubuntu 14.04.3 LTS) and if the `:from` field is not set, the sender address of the vacation is correct (I don't have that `system_user@server_hostname` problem) for me.

I mean, if the vacation rule gets fired because someone writes to the address X (and X is in the `:addresses` field), the automatic reply is sent using X as the sender.

So, IMHO http://trac.roundcube.net/ticket/1489221 refers to an older dovecot version, and it could be closed...

alecpl:
An option to fill the field automatically or autocompletion/selection for user emails might be a good addition still, but if you don't want to spend time on this it is fair enough.

mlocati:
Sorry, but I won't have any spare time for some time...

alecpl:
Committed with small improvements in 162bc85be9ded

PR #292, opened 2015-08-24
Dockerfile cleanup

macropin:
This is functionally identical to the previous Dockerfile - it has just been optimised:
- Minimise layers
- Clearly identify build and configuration steps
- Allow shell entry

NB, I'll get to tweaking the configuration contextualisation later.


thomascube:
I'm not a Docker expert so please, can @alunduil or @swarren83 verify this change?

swarren83:
The Dockerfile syntax is valid but I have some concerns regarding the style and scope of changes.

I like the addition of the .dockerignore file but we need to remove "installer" from this file and only ignore the Dockerfile and .git directory.

@macropin The Dockerfile is designed to build Apache > PHP > PERL > Roundcube to satisfy dependencies. Your goal of documenting the build/config steps does not require the Dockerfile to be restructured. 

How are you executing bash in the roundcube docker container? 

I used the below method and it worked without a change to the existing Dockerfile:

```
$docker run -d -p 80:80 --name="rc" alunduil/roundcube
$docker exec -ti rc bash
root@b4b17e8b984d:/#
```

@thomascube The .dockerignore file and "echo fix" are changes we need to work on.

Dockerfile
```diff
@@ -1,34 +1,37 @@
-FROM debian:8.0
+FROM debian:jessie
```
alunduil: I chose to pin the version to ensure that changes in interface and base image didn't impact the image in the future.  Do you have an argument for making the version less restrictive?

alunduil:
I do like the NONINTERACTIVE and the drop of '-e' on echo.

The over-arching structure I was going for  was to setup apache, then PHP, then roundcube.  In discrete sections so they could be changed independently without too much cache invalidation (apache withstanding of course).  This also makes it easier to use other base images for apache and php at some point if we so choose.  What argument do you have for re-structuring the file like this?

I specifically chose to use an entrypoint because docker images are commands and this one just happens to require apachectl.  You can still run bash if necessary either with the trick @swarren83 mentioned with exec or using `--entrypoint` on the command line.  I would much prefer keeping the entrypoint defined and the sections as they were unless there is a good reason to change it.

That all being said I do want:

* the dockerignore
* the new echo command
* the noninteractive environment variable

I'm alright with adding `rm -rf /var/lib/apt/lists` to all apt commands (to minimize individual layers).

Let me know what you think and thanks for the submission.

Dockerfile
macropin: No real argument, other than it seems a more common convention to track the upstream tag.

macropin:
To be honest, I didn't give all the details of this change set as much thought as I should have. 

What I need is a working container that is sane enough for production use. And this one didn't seem suitable (based on the SSL configuration, perceived lack of support for linked containers, and large number of layers), so I started hacking on it to get it to a point where it would work as a base. Minor details eg entry point, upstream image, were just me moving things around to what I believe is closer to Docker standards (if there is such a thing), or less likely to cause confusion for a user (eg splitting entrypoint / cmd args).

My needs were urgent, so I've gone and scratched my itch: [macropin/roundcube](https://github.com/macropin/docker-roundcube) and a corresponding service [unit](https://github.com/macropin/docker-units/blob/master/roundcube.service). It's based on debian:jessie, has support for linked mariadb, and imap services, and production ready SSL configuration. It also has 14 fewer layers. It's not perfect, and certainly not completely generic, but it's close. It could probably make use of `ONBUILD` Docker magic to allow complete override of the `config.inc.php`, or other tweaks to make it useful for development purposes. And then there is the minor issue of bootstrapping the initial schema creation... (I'd appreciate some suggestions here)

Perhaps we have different requirements, but for now at least mine are fulfilled. 

Cheers,

Andrew

alunduil:
No worries.  My goals with this Dockerfile were to follow the conventions docker enables (not necessarily the way people have started using it).  This means twelve factor application.  I ensured that all configuration parameters were configurable via environment variables (the docker way) and dropped the need for a configuration file completely.  Links won't work because docker's variable names are quite specific to the remote service and we haven't added variable expansion in environment variable configurables.  This could be added if we needed it.  Splitting entrypoint and command is actually the recommended way to run docker containers because they should be a command (a single executable that runs).  Therefore, I put in sensible apache requirements into entrypoint and left things that one might want to change when running in command.  SSL is handled by snakebite by default but by mounting a volume with your certificates you override that fairly easily.

I think we can go ahead and close this pull request but do you mind if we use a couple of the concepts you added to your changes?  Ideally, I'd love for you to update this pull request and contribute them directly but if you're satisfied I understand not having the desire to do that.  Please, let us know so we can resolve this.

macropin:
No worries. We'll leave it here then. You're welcome to pull my ideas in. No attribution necessary. Cheers.

PR #295, opened 2015-09-08
Update spellcheck regex to ignore URIs, emails, and additional quote char

lasley:
Hello,

Customers that utilize the spellcheck have been complaining that URIs and emails are not properly handled with the spellcheck function (as well as the ` char). 

This patch modifies the `$seperator` regex to treat URIs and Emails the same way as any other separator char. This allows for filtering of everything without the need to modify downstream code.

Suggestions of better ways to accomplish this are welcome, I am new to this community and don't know whether this was intended or not.

### DIFF
--- a/program/lib/Roundcube/rcube_spellcheck_engine.php
+++ b/program/lib/Roundcube/rcube_spellcheck_engine.php
@@ -31,7 +31,7 @@ abstract class rcube_spellcheck_engine
     protected $lang;
     protected $error;
     protected $dictionary;
-    protected $separator = '/\w+:\/\/\S*|[\s\r\n\t\(\)\/\[\]{}<>\\"`]+|[:;?!,\.](?=\W|$)/';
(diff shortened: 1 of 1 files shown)


alecpl:
Fine for URIs, but your email address regex definitely does not catch all addresses. I'm not sure it should filter out email addresses.

lasley:
Why would we want to spellcheck email addresses? Won't they primarily be names/pronouns, which tend to cause issues with the spellcheck? I'll definitely yield to your experience here if you feel we shouldn't do it, I'm just trying to think of a use case where emails need to be checked, particularly without the option for users to add words to their own dictionary. 

Do you maybe have an example of an email that doesn't work? I tested this with a good amount of the emails in my contact list and it seemed fine. For the most part, I just hacked together the one from http://www.regular-expressions.info/email.html - but there are still the issues presented in that article. Maybe it would be best to break this into its own regex, but that wouldn't really fit with the dynamics of the pre-existing code.

dereks:

```
git fetch upstream
git rebase upstream/master
```

PR #296, opened 2015-09-08
Filter quoted reply text from spellcheck $text

lasley:
Hello,

Customers that utilize the spellcheck have been complaining that previous emails (quoted replies) are being checked with the spellchecker enabled. 

This patch adds a `remove_quoted_reply` method in the abstract class, and then the appropriate `$text = $this->remove_quoted_reply($text)`s in the child `check` methods.

Suggestions of better ways to accomplish this are welcome, I am new to this community and don't know whether this was intended or not. I also pondered whether this should be a setting, but couldn't really find a use case for wanting to scan previous emails.

This PR is dependant on #295. I submitted them separately because I feel this one may generate more discussion/input than #295, as well as them being two different features

### DIFF
--- a/program/lib/Roundcube/rcube_spellcheck_enchant.php
+++ b/program/lib/Roundcube/rcube_spellcheck_enchant.php
@@ -82,7 +82,10 @@ class rcube_spellcheck_enchant extends rcube_spellcheck_engine
         if (!$this->enchant_dictionary) {
             return array();
         }
-
+        
+        // get rid of quoted reply text
+        $text = $this->remove_quoted_reply($text);
+        
         // tokenize
         $text = preg_split($this->separator, $text, NULL, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_OFFSET_CAPTURE);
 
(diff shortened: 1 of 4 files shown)


alecpl:
We have a ticket for this (http://trac.roundcube.net/ticket/1488514) and we'll accept such improvement. However:
1. please, do not mix commits from other PRs, create separate branch for each feature/PR.
2. I think quoted text removal should be done out of the spellchecker class(es) or, when I think about the point below, it should be there but optional.
3. Does it work at all? Without testing I feel it works only in HTML mode, we'd need this to work with both HTML and plain text editor (get() and get_xml() results should handle this correctly).

lasley:
Hi @alecpl 

Oops, sorry about the PR mixing. I'll rebase if the other one isn't merged by the time we finish the open action items.

It does work, but you are correct about that only being in HTML mode. I honestly didn't realize the `get()` and `get_xml()` methods existed.

I'll push a fix in a week or so, thanks for the pointers.

dereks:

```
git fetch upstream
git rebase upstream/master
```

PR #299, opened 2015-09-18
attachment_reminder plugin: config file with locale-independent global keywords

emtiu:
Currently, two keywords are hardcoded into the plugin to indicate a missing attachment: '.doc' and '.pdf'. This patch remopves that hardcoding and allows the user to set custom global keywords in a config file.

My personal use case is that I frequently compose messages in a language other than the set locale. I use this global config file to include keywords from other locales.

### DIFF
--- a/plugins/attachment_reminder/attachment_reminder.js
+++ b/plugins/attachment_reminder/attachment_reminder.js
@@ -33,7 +33,7 @@ function rcmail_get_compose_message()
 
 function rcmail_check_message(msg)
 {
-  var i, rx, keywords = rcmail.gettext('keywords', 'attachment_reminder').split(",").concat([".doc", ".pdf"]);
+  var i, rx, keywords = rcmail.gettext('keywords', 'attachment_reminder').split(",").concat(rcmail.env.global_keywords);
 
   keywords = $.map(keywords, function(n) { return RegExp.escape(n); });
   rx = new RegExp('(' + keywords.join('|') + ')', 'i');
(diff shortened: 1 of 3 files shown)


thomascube:
One minor suggestion: set the default values for the new config option right in the code and cast to array:
```
(array)$rcmail->config->get('global_keywords', array('.doc','.pdf'));
```
This makes sure the plugin still works with an old config that doesn't provide the new option.

emtiu:
Wouldn't the now-included config.inc.php.dist take care of that? Anyway, I can do that no problem :)

thomascube:
Nope, `config.inc.php.dist` is only a template one needs to copy into `config.inc.php`. This is likely not done when updating the plugin. There's no config merging routine in place but maybe that's something you might want to look into :-)

emtiu:
Okay, this should conform to your comments now.

alecpl:
Well, this is useful if you're the only user of Roundcube. To make this really useful users should be able to define the list by their own in Preferences.

emtiu:
I would contend that the current patch can also be useful for company-wide installations, or other situations where users share specific usecases. In any case, it is more useful than the current situation, which requires changing the Roundcube code to set any custom attachment reminder keyword at all.

Still, you make a good point about a user-specific option. Unfortunately, I don't (yet) have the knowledge to implement this outright.

alecpl:
Many plugins add new preferences, e.g. emoticons, archive, newmail_notifier. See how they handle preferences_list and preferences_save hooks.

emtiu:
Thank you for your advice, Alex. I will consult some equivalent code and open a new pull request in time.

PR #300, opened 2015-10-07
Make user_domain setting to be used as the mail_domain

shemgp:
I'm proposing a feature to make the config user_domain be used to set the address book mail_domain.  This is better than using mail_domain config as the email domain can be set to a different domain per host.

An example usage is that of Google Apps Email with custom domains.   If I have a custom domain my imap server is still imap.googlemail.com but my email addresses should not be googlemail.com but my custom domain's.  mail_domain will fix it if I have only one domain, but what if I have to custom domains?


alecpl:
It does not look right for me.
1. username_domain array is indexed by host name not domain name.
2. I suppose this should be made optional and moved to the mail_domain() method.

PR #307, opened 2015-11-08
Update language_in option of Closure Compiler to match current code

johndoh:
The JavaScript Promise code added in 7b8a0af and 2965a98 is not supported under the ES3 spec. Update the language_in setting of Closeure Compiler to ES5 to avoid warnings like:

```
./bin/../program/js/app.js:3396: WARNING - Keywords and reserved words are not allowed as unquoted property names in older versions of JavaScript. If you are targeting newer versions of JavaScript, set the appropriate language_in option.
    }).catch(function(err) {
       ^
```

also

```
./bin/../program/js/publickey.js:8: ERROR - Parse error. IE8 (and below) will parse trailing commas in array and object literals incorrectly. If you are targeting newer versions of JS, set the appropriate language_in option.
        "https://keys.fedoraproject.org/",
        ^
```

though the warnings in publickey.js could also be fixed just be removing the trailing commas.

### DIFF
--- a/bin/jsshrink.sh
+++ b/bin/jsshrink.sh
@@ -2,7 +2,7 @@
 PWD=`dirname "$0"`
 JS_DIR="$PWD/../program/js"
 JAR_DIR='/tmp'
-LANG_IN='ECMASCRIPT3'
+LANG_IN='ECMASCRIPT5'
 # latest version requires Java 7, we'll use an older one
 #CLOSURE_COMPILER_URL='http://dl.google.com/closure-compiler/compiler-latest.zip'
 CLOSURE_COMPILER_URL='http://dl.google.com/closure-compiler/compiler-20131014.zip'


thomascube:
Indeed, we can probably move forward to ES5 with Roundcube master. We already dropped support for IE < 9 for the upcoming versions.

alecpl:
We still support IE7-8 and FF3.6 with legacy_browser plugin. So, would be good to not break the old browsers support. Of course, new features can be ignored on these browsers, but we don't want to see js errors there. Maybe we could drop legacy_browser in 1.3, but I think we should keep it in 1.2.

thomascube:
Good point! At least for now we can get away with a slightly modified but not regressive syntax change as submitted in 310d4939.

thomascube:
FWIW: there's one `.catch()` call in `plugins/enigma/enigma.js:109`. @alecpl maybe you can check whether that can be changed, too?

PR #309, opened 2015-11-12
Download attachments on saved drafts

sakanaou:
I often happen to use mail drafts as simple temporary file storage. Unfortunately roundcube doesn't allow to download attachments on saved drafts. With this change the attachment list is extended to offer download links for each attachment.
The draft has to be saved and the mail editor re-opened in order to work.

The change might be a bit too simple, please bear with me.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -294,7 +294,7 @@ function rcube_webmail()
         }
         else if (this.env.action == 'compose') {
           this.env.address_group_stack = [];
-          this.env.compose_commands = ['send-attachment', 'remove-attachment', 'send', 'cancel',
+          this.env.compose_commands = ['send-attachment', 'remove-attachment', 'download-attachment', 'send', 'cancel',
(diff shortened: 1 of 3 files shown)


alecpl:
Of course it allows it in draft mail preview, but not in compose. Yes. Your change is fine, but not complete. E.g. it won't work when you modify the message and hit "save draft" button. It's because the message UID will change.

sakanaou:
Thanks for pointing that out, alecpl. I have adjusted the patch so that it does not depend on $MESSAGE anymore but rather creates the download url via Javascript upon request. This way it also works if the draft gets saved in between.

alecpl:
Looks better. However, I think we'd need a different approach. What if you attach a new attachment or delete one (other attachment IDs will change)? I think we'd need to use session attachments not the draft message.

alecpl:
Implemented in 8a13615e9

PR #313, opened 2015-12-28
add .log suffix to all log file names

remicollet:
As lot of files are used (sieve, ldap, errors, password, smtp, imap, session, ...) having such a suffix will make much easier to configure logrotate, using

```
 /var/log/roundcubemail/*.log {
     ...
 }
```

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -77,6 +77,9 @@ $config['log_date_format'] = 'd-M-Y H:i:s O';
 // set to 0 to avoid session IDs being logged.
 $config['log_session_id'] = 8;
 
+// Default extension used for log file name
+$config['log_file_ext'] = '.log';
+
 // Syslog ident string to use, if using the 'syslog' log driver.
 $config['syslog_id'] = 'roundcube';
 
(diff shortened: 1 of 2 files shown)


remicollet:
Rebased.

Can you please consider this really minor change for 1.3 ?

alecpl:
We're after the beta release. So, I say no. This is be possible with a plugin using 'write_log' hook.

remicollet:
@alecpl ok, so please consider for master as soon as 1.3 will be branched

FYI this patch is applied in downstream packages (Fedora / EPEL) since version 1.1...

remicollet:
As 1.3 is realeased and branched, can you please consider this PR ?

alecpl:
Yes, but we'd like to have it optional. To make the transition easy for everyone. Could you add log_file_ext option that defaults to '.log' if not set?

remicollet:
> Could you add log_file_ext option that defaults to '.log' if not set?

I will try shortly, I'm not a big roundcubemail expert, but will try.

remicollet:
@alecpl done (I think ;) )

remicollet:
Thanks

PR #314, opened 2016-01-02
Update rcube_sieve_vacation.php

bevand10:
Problem encountered where $auto_addr was true and vacation['addresses'] was undefined. This caused a null/empty addresses value. The above check ensures that the addresses value is only used if it is defined and populated.

Adding the !empty check ensures what I believe to be the expected behaviour where $auto_addr is true.


alecpl:
But... If the vacation rule already exists and has no addresses defined we should not put user emails there.

bevand10:
Based on my limited experience with sieve so far, a vacation rule without an address is simply ignored and does not fire.

Perhaps its just my environment that works this way (RC in SME Server 9.1)

alecpl:
This is not required by standard and there are systems that do not require :addresses in vacation. So, it looks like we'd need an additional config option.

PR #315, opened 2016-01-06
Release 1.1

yano0takashi:
[patch2charset.zip](https://github.com/roundcube/roundcubemail/files/79279/patch2charset.zip)
This is the modified program/lib/Roundcube/rcube_charset.php against 1.1.4.
The reason is:
Some Japanese characters are not shown as expected while mail clients (ex. ThunderBird) can show them as expected.

The mails in ISO-2022-JP encoding can include characters not in ISO-2022-JP.
These characters are included in the encoding ISO-2022-JP-MS.

This patch use ISO-2022-JP-MS with the MINE's ISO-2022-JP.


alecpl:
You did some mistake when creating the PR. Please, create a new one that contains only the change you want using master branch.

PR #317, opened 2016-01-07
Enable use of TLSv1.1 and TLSv1.2 for IMAP.

FrancisRussell:
Currently, Roundcube is only able to use TLSv1 to connect to IMAP. The use TLSv1 is becoming increasingly discouraged. This patch enables IMAP connections using TLSv1.1 and TLSv1.2. Net_SMTP uses the same flag handling for its support of TLSv1.1 and TLSv1.2.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@ CHANGELOG Roundcube Webmail
 - Fix PHP7 warning "session_start(): Session callback expects true/false return value" (#1490624)
 - Fix XSS issue in SVG images handling (#1490625)
 - Fix missing language name in "Add to Dictionary" request in HTML mode (#1490634)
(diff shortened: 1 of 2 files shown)


alecpl:
Looks good. We could make the method configurable like Net_SMTP did. In our case it would be $this->prefs['socket_options']['ssl']['crypto_method'].

FrancisRussell:
I've written the obvious code but I don't really have an idea how to test it. I'm not really sure if there's much benefit in being able to control this option from higher up. A possible benefit might be to force use of higher TLS versions in the client, but I don't think this is supported elsewhere, e.g. for SMTP. I'm inclined to keep things just as the first commit, but I'm happy to test the second if you can give some more info on how to do this.

PR #318, opened 2016-01-08
Encode HTML special chars in additional headers plugin

mcnewton:
Otherwise added headers that e.g. contain email@example.com will send that directly to the browser.


mcnewton:
...and noting github comments where "<" and ">" have been removed and handled specially!

alecpl:
Thanks, but here a better fix 9aaeb23d9d

mcnewton:
That's cool - thanks. You know the code far better than I do! :-)

PR #319, opened 2016-01-09
master

yano0takashi:
[patch2charset.zip](https://github.com/roundcube/roundcubemail/files/83538/patch2charset.zip)

This is the modified program/lib/Roundcube/rcube_charset.php against master.
The reason is:
Some Japanese characters are not shown as expected while mail clients (ex. ThunderBird) can show them as expected.

The mails in ISO-2022-JP encoding can include characters not in ISO-2022-JP.
These characters are included in the encoding ISO-2022-JP-MS.

This patch use ISO-2022-JP-MS with the MINE's ISO-2022-JP.


thomascube:
Please create a pull request from your forked master to our master branch. This one is totally broken as it tries to merge from release-1.1 branch to master. Please try again.

PR #320, opened 2016-01-17
Removed start of line in matches

ghost:
Fails on my system and really unuseful


alecpl:
I'm pretty sure these are fixed in git-master now.

PR #321, opened 2016-01-19
Added try catch for 'Access is denied' error in IE11

jbuncle:
Occurs when using cross-domain iframes. Issue identified and fixed with IE11 on Win7, may effect other versions.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -612,11 +612,14 @@ function rcube_webmail()
       .mouseup(body_mouseup)
       .keydown(function(e){ return ref.doc_keypress(e); });
 
-    $('iframe').on('load', function(e) {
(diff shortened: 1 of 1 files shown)


jbuncle:
Problem only occurs when an iframe references another domain to that of the parent. Specifically we added an embedded youtube video which caused the error, and subsequently broke the javascript on the page. Recreated using Window7 & IE11 VM provided by Microsoft.

Not normally an issue as roundcube doesn't have any such iframes (as they reference itself) by default.

alecpl:
Confirmed, the error is thrown in jQuery's contents() method which uses contentDocument as we do in onload handler.

alecpl:
Thanks for the report. It's fixed in 64763853c9c7.

PR #322, opened 2016-02-05
autoselect_host(): support pop. prefix

kylon:
Some hosts use pop. for both IMAP and POP.
Add a flag to fix auto-selection for these hosts.


alecpl:
I'm pretty sure we don't need this. i.e. can be achieved with existing configuration options.

kylon:
ok, can please point to me to this options?
the wiki is not clear and i cannot login in a user-friendly way to these hosts

thanks

PR #323, opened 2016-02-13
add optional security headers as comment to .htaccess

Rotzbua:
Hi,
I added the most used optional header to .htaccess as a comment, so it do not have any effect.
I think it is helpful for people who like to harden their installation.

Source: https://www.owasp.org/index.php/List_of_useful_HTTP_headers
Testing: https://securityheaders.io/

Best

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -46,7 +46,34 @@ SetOutputFilter DEFLATE
 <IfModule mod_headers.c>
 # replace 'append' with 'merge' for Apache version 2.2.9 and later
 #Header append Cache-Control public env=!NO_CACHE
+
+# Optional security header
+# Only increased security if the browser support those features
+# Be careful! Testing is required! They should be adusted to your intallation / user environment
+
+# HSTS - HTTP Strict Transport Security
+#Header always set Strict-Transport-Security "max-age=31536000; preload" env=HTTPS
+
+# HPKP - HTTP Public Key Pinning
+# Only template - fill with your values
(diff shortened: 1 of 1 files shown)


sebix:
HSTS and HPKP should be done per-domain (system-wide), not per-application.

I'd like to see the others enabled by default, as they totally make sense on application basis (to prevent attacks on users of the application). But I'm not sure if they should be set via PHP or `.htaccess`. Setting them via PHP enables them for all webservers, not only Apache.

Rotzbua:
I agree that HSTS should be done by system config and per (sub)domain. But shared hosters only allow custom headers by .htaccess or php.

I agree with your .htaccess only for Apache argument, but then the other comment for CSP (Content-Security-Policy) should be removed.

.htaccess
```diff
@@ -46,7 +46,34 @@ SetOutputFilter DEFLATE
...
+#Header set X-Frame-Options SAMEORIGIN
+
+# X-Content-Type-Options
+# It prevents Google Chrome and Internet Explorer from trying to mime-sniff the content-type of a response away from the one being declared by the server.
+#Header set X-Content-Type-Options: "nosniff"
+
+# CSP - Content Security Policy
 # for better privacy/security ask browsers to not set the Referer
+# more flags for script, stylesheets and images available, read RFC for more information
 #Header set Content-Security-Policy "referrer no-referrer"
```
sebix: I can't find `no-referrer` in the RFC: https://www.w3.org/TR/CSP/

alecpl: https://w3c.github.io/webappsec-referrer-policy/ It's a working draft, but some browsers support it. https://developer.mozilla.org/en/docs/Web/Security/CSP/CSP_policy_directives

PR #324, opened 2016-02-18
Support MESSAGE QUOTA

ionescu-bogdan:
Update for: Show Quota Messages


alecpl:
As it's IMAP generics class, I'd rather return both STORAGE and MESSAGES quota information here and decide what to display in rcmail.

PR #326, opened 2016-03-05
Add Google Larry skin

rajbdilip:


nikku:
Googie Larry is published under no / proprietary license.

Don't think you can easily include this in roundcube core.

Issue #5139, opened 2015-04-29
scroll by keyboard


rcubetrac:
_Reported by nerijus on 29 Apr 2015 11:40 UTC as Trac ticket #1490373_

If I open a message by doubleclicking it I cannot scroll it down by pressing down arrow. I have to press mouse first (focus) somewhere in the message.

_Migrated-From: http://trac.roundcube.net/ticket/1490373_

guitarmanusa:
See PR:  https://github.com/roundcube/roundcubemail/pull/5184

alecpl:
PR merged.

PR #5184, opened 2016-04-02
Implements 'scroll by keyboard #5139'

guitarmanusa:
On message load, changes focus to messagecontent div which makes it scrollable by the keyboard.

### DIFF
--- a/skins/larry/ui.js
+++ b/skins/larry/ui.js
@@ -172,6 +172,10 @@ function rcube_mail_ui()
         if (get_pref('previewheaders') == '1') {
           toggle_preview_headers();
         }
+
+        if (rcmail.env.action == 'show') {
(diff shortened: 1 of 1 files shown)


skins/larry/ui.js
```diff
@@ -172,6 +172,10 @@ function rcube_mail_ui()
         if (get_pref('previewheaders') == '1') {
           toggle_preview_headers();
         }
+
+        if (rcmail.env.action == 'show') {
+            document.getElementById('messagecontent').focus()
```
thomascube: We prefer jQuery over native DOM functions. Please use $('#messagecontent').focus() is that also works

guitarmanusa:
That worked.  Changed from native DOM call to jQuery.

PR #5191, opened 2016-04-05
Switch some defaults to be more secure (ex: TLS on)

seanm:
Also improved comments regarding these settings.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -123,7 +123,9 @@ $config['apc_debug'] = false;
 // The mail host chosen to perform the log-in.
 // Leave blank to show a textbox at login, give a list of hosts
 // to display a pulldown menu or set one host as string.
-// To use SSL/TLS connection, enter hostname with prefix ssl:// or tls://
+// To use TLS connection (most secure), enter hostname with prefix tls://
+// To use SSL connection (insecure as of 2014), enter hostname with prefix ssl://
+// To use no encryption (insecure), enter hostname with no prefix.
 // Supported replacement variables:
 // %n - hostname ($_SERVER['SERVER_NAME'])
 // %t - hostname without the first part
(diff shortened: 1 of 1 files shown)


CRtEurope:
You wrote down: $config['default_host'] = 'tls://localhost';
In this case it won't work on every host. It'd be better to use the following config:
$config['default_host'] = 'ssl://localhost'; // This item works on every host and uses the correct TLS connection
The config for the smtp connection as you have described works on every host host.

seanm:
I'm new to roundcube, so frankly I know little about it.  I'm just setting it up and therefore going through the possible settings. Many of the default settings struck me as not at all ideal, hence this patch.

So you're saying specifying "ssl://" will in fact use TLS?  Then why have the "tls://" option at all?

CRtEurope:
It depends on what kind of software you use. Eg dovecot or something. And it depends on your cypher settings. If you deactivated ssl and activated tls only on dovecot for example.
dovecot cypher settings for example:
ssl_cipher_list = EECDH+AESGCM:EDH+AESGCM:EECDH+ECDH+AESGCM:DH+AESGCM:EECDH+ECDH+AES256:DH+AES256:EECDH+ECDH+AES128:DH+AES:EECDH+ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS:!RC4:!DES:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-CBC3-SHA

ssl_protocols = !SSLv2 !SSLv3 TLSv1 TLSv1.1 TLSv1.2

This will enable TLS only and works with the roundcube setting $config['default_host'] = 'ssl://localhost'; or $config['default_host'] = 'ssl://localhost:993';

seanm:
Sorry for being thick, but I don't fully follow... :)

Could you explain the current comment in defaults.inc.php which says: "To use SSL/TLS connection, enter hostname with prefix ssl:// or tls://".

When it says "ssl:// or tls://" does it mean those are synonymous?

My assumption was that specifying "tls://" would forcibly exclude SSL3, regardless of server configuration.  (I understand that one can disable older protocols on one's mail server, sure.)

CRtEurope:
As I said before. It just depends on what kind of third party software you use. Dovecot, UW IMAP, Cyrus or something similar and on your protocol and cypher settings in those softwares.

seanm:
So roundcube passes the "default_host" setting down to the mail server (dovecot, or whichever) as a string, and some mail servers recognize the "tls://" scheme, and some don't, is that it?

Anyway, I guess this part of my change is wrong:

```
-// To use SSL/TLS connection, enter hostname with prefix ssl:// or tls://
+// To use TLS connection (most secure), enter hostname with prefix tls://
+// To use SSL connection (insecure as of 2014), enter hostname with prefix ssl://
+// To use no encryption (insecure), enter hostname with no prefix.
```

How would you suggest to word it?  My goal here is to improve the comments so that people setting it up for the first time (like me) can clearly understand what to put.

alecpl:
TLS/SSL on some systems may require setting additional socket options (e.g. to accept self-signed certs). Also, it highly depends on server configuration. I think not using protocol prefix is good enough for a software like Roundcube.

ps. port 993 is usually for SSL-secured connections, and 143 for non-secured, but with possibility to use STARTTLS.

sebix:
@seanm As far as I understood `tls://` does mean enforced StartTLS.

PR #5193, opened 2016-04-08
Controlled logging in the browser

lisandroTSilva:


alecpl:
What exactly is this supposed to improve?

lisandroTSilva:
Hello. The idea is to eliminate residue and / or unnecessary processing.

alecpl:
Oh, indeed. I remember in old days if browser development tools were disabled window.console was undefined. It looks like these days it's not true. However, instead of a new config option we should probably just use devel_mode option.
I'm curious what perf impact it has. Probably browsers optimize somehow use of console.log() when the console is not used.

alecpl:
Or we should get rid of it at all. These days browser development tools are good enough so we don't need this.

lisandroTSilva:
First, thank you for your attention.

I understand and agree. The procedure should be almost negligible.
I've seen some libraries and frameworks that use the console to display information that is generally relevant to the developer. Therefore I believe not be elegant display this information.
You suggest using devel_mode? The same can be defined within the config.inc.php if major problems?

alecpl:
I suggest to not use console in a way it's used now and get rid of rcmail.log method. I.e. we don't need to log request data nor server responses.

thomascube:
I wouldn't go as far as removing `rcmail.log` as it can be useful for debugging. Adding another config option is also a bit too much so what about binding to it the `devel_mode` option:
```php
$this->devel_mode = $this->config->get('devel_mode');
$this->set_env('enable_console_log', (bool)$this->devel_mode);
```

PR #5198, opened 2016-04-12
Update labels.inc

freecris:
Both "From" and "sender" were translated to "Mittente".
This is confusing when creating filters.


alecpl:
Please, use our Transifex service.

freecris:
Just asked to join the community.

Thank you to point me there.

Issue #5240, opened 2016-05-05
Map Area- Persistent XSS Vulnerability in mail content page,please confirm.


fantasy7082:
Steps to reproduce:
-Compose email content via HTML mode.You can use email sending tool or other webmail to send it.The HTML content is below:
` <img src=# usemap=#foo width=100%><map name="foo"><area href=javascript:alert(document.domain) shape=default>`
-Send the mail to roundcube mail system
-Log in to roundcube mail system and open the mail received
-Click the area near the image in the content.The xss will be triggered

The vulnerability can be  triggered in Chrome,Firefox,Safari.
The testing roundcube webmail version is 1.1.5.Link is below:
https://github.com/roundcube/roundcubemail/releases/download/1.1.5/roundcubemail-1.1.5-complete.tar.gz
The test screenshot link:http://pan.baidu.com/s/1nvQt9Eh

fgeek:
No CVE assigned?

thomascube:
That's likely [CVE-2016-4552](https://www.cvedetails.com/cve/CVE-2016-4552/)

thomascube:
... and also CVE-2016-5103

fgeek:
@thomascube Thank you and have a nice weekend.

PR #5241, opened 2016-05-05
Fix map area persistent XSS vulnerability

dsoares:
Fix Map Area- Persistent XSS Vulnerability in mail content page. (#5240)


dsoares:
I tested with success the found vulnerability. This PR fixes the vulnerability (at least, with my test case).

alecpl:
Thank you, we fixed the issue.

PR #5252, opened 2016-05-14
Password plugin, cPanel driver

julianmatz:
Add support for hash auth (cPanel Remote Access Key) in the Password plugin's cPanel driver and call the $xmlapi->listaccts() method to retrieve the cPanel account user that owns the domain (required for cPanel reseller accounts).

### DIFF
--- a/plugins/password/drivers/cpanel.php
+++ b/plugins/password/drivers/cpanel.php
@@ -46,12 +46,12 @@ class rcube_cpanel_password
         $this->xmlapi = new xmlapi($rcmail->config->get('password_cpanel_host'));
         $this->xmlapi->set_port($rcmail->config->get('password_cpanel_port'));
         // Hash auth
-        if (!empty($rcmail->config->get('password_cpanel_hash'))) {
-            $this->xmlapi->hash_auth( $this->cuser, $rcmail->config->get('password_cpanel_hash'));
+        if (!empty($cpanel_hash = $rcmail->config->get('password_cpanel_hash'))) {
+            $this->xmlapi->hash_auth( $this->cuser, $cpanel_hash);
         }
         // Pass auth
-        else if (!empty($rcmail->config->get('password_cpanel_password'))) {
-            $this->xmlapi->hash_auth( $this->cuser, $rcmail->config->get('password_cpanel_password'));
+        else if (!empty($cpanel_password = $rcmail->config->get('password_cpanel_password'))) {
+            $this->xmlapi->hash_auth( $this->cuser, $cpanel_password);
         }
         else {
             return false;
--- a/plugins/password/drivers/cpanel.php
+++ b/plugins/password/drivers/cpanel.php
@@ -45,22 +45,19 @@ class rcube_cpanel_password
         // Setup the xmlapi connection
         $this->xmlapi = new xmlapi($rcmail->config->get('password_cpanel_host'));
         $this->xmlapi->set_port($rcmail->config->get('password_cpanel_port'));
-		
-		// Hash auth
-		if ( ! empty ( $rcmail->config->get( 'password_cpanel_hash' ) ) ) {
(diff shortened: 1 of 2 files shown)


alecpl:
Could you keep our code style and indentation, please?

julianmatz:
Sure. I've fixed it in commit [8042955](https://github.com/roundcube/roundcubemail/pull/5252/commits/80429559271996f9ac7534c02ed8da29c24cb828)

plugins/password/drivers/cpanel.php
```diff
@@ -45,7 +45,18 @@ public function save($curpas, $newpass)
...
         $this->xmlapi = new xmlapi($rcmail->config->get('password_cpanel_host'));
         $this->xmlapi->set_port($rcmail->config->get('password_cpanel_port'));
-        $this->xmlapi->password_auth($this->cuser, $rcmail->config->get('password_cpanel_password'));
+        // Hash auth
+        if (!empty($rcmail->config->get('password_cpanel_hash'))) {
+            $this->xmlapi->hash_auth( $this->cuser, $rcmail->config->get('password_cpanel_hash'));
+        }
+        // Pass auth
+        else if (!empty($rcmail->config->get('password_cpanel_password'))) {
+            $this->xmlapi->hash_auth( $this->cuser, $rcmail->config->get('password_cpanel_password'));
```
alecpl: Here password_hash() should be used. Also, I'd like to see config->get() use once per option.

plugins/password/drivers/cpanel.php
```diff
@@ -45,7 +45,18 @@ public function save($curpas, $newpass)
...
+        }
+        else {
+            return false;
```
alecpl: Here, would be nice to return error code and message.

plugins/password/drivers/cpanel.php
```diff
@@ -70,8 +81,16 @@ function setPassword($address, $password)
...
 
         $data['password'] = $password;
+        
+        // Get the cPanel user
+        $query = $this->xmlapi->listaccts( 'domain', $data['domain'] );
+        $query = json_decode( $query, true );
+        if ( $query['status'] != 1 ) {
+            return false;
+        }
+        $cpanel_user = $query['acct'][0]['user'];
```
alecpl: We have username from config, why we need to overwrite it here? I don't know cPanel API, but can you call listaccts() when using user/password auth. Do we need it at all when not using hash auth?

julianmatz: Yes, it's required whether the password or the hash is used. The username retrieved from config is the API user (root user or reseller user). The cPanel user needs to be retrieved using listaccts() because otherwise, the following error will be returned:

`You do not have an email account named “email.alias@example.com”.`

(Even if authenticated as root).

This can easily be tested on a cPanel server, using the following example URL:

`https://hostname.example.com:2087/cpsess##########/json-api/cpanel?cpanel_jsonapi_user=user&cpanel_jsonapi_apiversion=2&cpanel_jsonapi_module=Email&cpanel_jsonapi_func=passwdpop&domain="example.com"&email="user"&password="12345luggage"`

plugins/password/drivers/cpanel.php
julianmatz: I've modified the code to call config->get() only once. Commit is here:
https://github.com/roundcube/roundcubemail/pull/5252/commits/7ed0d3308b2b5968cc739f3f1035c1b866661deb

The hash key here is a unique remote access key that's generated and stored in the WHM user's account. For example, for the root user, it would be stored in /root/.accesshash. I don't think it can be generated outside of the WHM interface.

https://documentation.cpanel.net/display/ALD/Remote+Access+Key

plugins/password/drivers/cpanel.php
julianmatz: I'm not too familiar with the error handling. Would it be done using the following? `$rcmail->output->command('display_message', 'message, 'error');`

PR #5265, opened 2016-05-23
use input fields of type 'password' for enigma plugin key generation

ghost:
proper use of rcube_output::get_edit_field within enigma plugin -> key generation form


pnebert:
when you look at the cube_output::get_edit_field function, you will see that the type parameter isnt applying for password fields, for password field you have to add the type into the attrib array parameter. a little bit confusing :-)

alecpl:
Thanks I fixed get_edit_field() in 58c036116b8.

PR #5268, opened 2016-05-23
Improve randomness

### LINKED ISSUE #5266: Don't use insecure RNGs for password salts
https://github.com/roundcube/roundcubemail/blob/ab6fdfa8bc3a441590778b026b52225473424eec/plugins/password/password.php#L657-L673

Recommendation:

1. Use [password_compat](https://github.com/ircmaxell/password_compat).
2. Failing that, use [random_compat](https://github.com/paragonie/random_compat) to polyfill `random_int()` from PHP 7 and use that function instead.

paragonie-scott:
Closes #5266

### DIFF
--- a/composer.json-dist
+++ b/composer.json-dist
@@ -22,6 +22,7 @@
     ],
     "require": {
         "php": ">=5.3.7",
+        "paragonie/random_compat": "^1|^2",
         "roundcube/plugin-installer": "~0.1.6",
         "pear-pear.php.net/net_socket": "~1.0.12",
         "pear-pear.php.net/auth_sasl": "~1.0.6",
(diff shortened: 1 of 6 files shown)


alecpl:
I don't like it. There's no exceptions handling where random_* functions are used. We should just use rcube_utils::random_bytes(). From the other side, we'll not hit lack-of-entropy issue if we use real random functions everywhere?

paragonie-scott:
In that case: You might also want to consider creating a rcube_utils::random_int() that handles exceptions from `\random_int()` and replace all `mt_rand()`/`rand()` invocations with `rcube_utils::random_int()`.

PR #5278, opened 2016-05-25
Create de_DE.inc

ThAchi1:
Added German translation for enigma-plugin.


alecpl:
Thanks. I added all missing localizations of Enigma in another commit. Use Transifex for updates.

PR #5281, opened 2016-05-26
Add new hook to modify head fields order

Takika:
Add hook to allow plugins to modify order of the contact form head fields

With this hook a plugin can change the order of the contact form head fields from
`'names'    => array('prefix','firstname','middlename','surname','suffix'),`
to eg.:
`'names'    => array('prefix','surname','firstname','middlename','suffix'),`
or any other language dependent special order.

### DIFF
--- a/program/steps/addressbook/func.inc
+++ b/program/steps/addressbook/func.inc
@@ -573,6 +573,10 @@ function rcmail_contact_form($form, $record, $attrib = null)
                 'jobtitle'     => array('jobtitle'),
             );
 
+            // Allow plugins to modify order of the contact form head fields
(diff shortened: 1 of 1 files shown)


alecpl:
You can already modify them using contact_form hook.

Takika:
I think that you can't change the field order in the contact_form hook. The order of the names fields are hardcoded in https://github.com/roundcube/roundcubemail/blob/master/program/steps/addressbook/func.inc#L568 .

alecpl:
Right.

Takika:
So, if I have right can you merge this?

alecpl:
I'd prefer not adding another hook for just this. So, moving $field_blocks definition before contact_form hook and adding field_blocks there would be better.

Takika:
Ok, I will be open a new PR for this.

Issue #5264, opened 2016-05-23
XSS vuln in body, position:fixed style escapes message element


riqpe:
Possible to escape message element by setting style `position:fixed`. The end results range from annoyance from blocked navigation elements, to possible click jacking.

Simple POC; send email with content `<p style="position: fixed; margin: 0px; top: 0px; left: 0px;">XSS</p>`

Works on Classic and Lenny themes, and on the print page, as seen on the attached screenshots. Tested on Roundcube versions 1.1.4 and 1.2.0, on Firefox and Chrome.

![xss](https://cloud.githubusercontent.com/assets/1518025/15470912/7e2c7a5c-20f3-11e6-9d03-51a06307ba7c.png)

Above you may see the text `XSS` in the top left corner.

alecpl:
I don't see what this has to do with XSS. Indeed "malformed" HTML can change the whole page, so yes click-jacking maybe. This can be fixed only by using an iframe, which I don't think we'd like to implement soon. I'm not sure if removing styles with "position" would be an acceptable solution.

Yes, I know there's styles sandboxing technology, but it's not supported by all browsers and I'm not sure it would help here.

riqpe:
Overriding the position to absolute would fix the issue, as the positioning would be relative to the message box.

alecpl:
Fixed in 6737e29 and backported to 1.2, 1.1, 1.0.

PR #5284, opened 2016-05-29
Wash position:fixed style in HTML mail for better security (#5264)

alecpl:
Possible solution for #5264 Though via PR as I don't really like the idea of replacing position fixed with absolute.


alecpl:
@thomascube please review.

program/lib/Roundcube/rcube_washtml.php
```diff
@@ -727,10 +732,9 @@ public static function fix_broken_lists(&$html)
      */
     protected function explode_style($style)
     {
-        $style = trim($style);
+        $pos = 0;
```
thomascube: What is this change supposed to fix or improve in this matter?

alecpl: There's no need to trim before we remove comments, better to trim after removing comments.

thomascube: OK. This seemed like an irrelevant change to the actual topic of this PR

thomascube:
I'm good with this change. position:fixed wouldn't fit well into the embedded message preview thus replacing it with absolute can make sense regardless the XSS situation.

PR #5288, opened 2016-05-30
Fix error in german translation.

heptalium:


alecpl:
Please, use Transifex service. Otherwise this can be overwritted by localization update in the future. We usually update localizations from Transifex service before every release.

heptalium:
How does this work? If I call the URL from the file I just get a 403 error message.

alecpl:
Try this https://www.transifex.com/roundcube/roundcube-webmail/translate/#de_DE/plugin-enigma

CRtEurope:
I get a 404 error
unbenannt

heptalium:
I'm too (after creating an account)

alecpl:
Maybe you need to login first. Hopefully this works: https://www.transifex.com/roundcube/roundcube-webmail/plugin-enigma

heptalium:
I can call the URI, but then I can't do anything.

CRtEurope:
Me'nt either

PR #5294, opened 2016-05-31
help plugin config.inc.php.dist

rmcaninch:
Updated 1.0 to 1.1 documentation. Added a note that the URL may need changed to https if Roundcube is served secure.

### DIFF
--- a/plugins/help/config.inc.php.dist
+++ b/plugins/help/config.inc.php.dist
@@ -2,7 +2,8 @@
 
 // Help content iframe source
 // %l will be replaced by the language code resolved using the 'help_language_map' option
-$config['help_source'] = 'http://docs.roundcube.net/doc/help/1.0/%l/';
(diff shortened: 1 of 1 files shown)


Tragen:
Why not remove http:// completely so the browser uses the current protocol?

rmcaninch:
I can't just remove it from the config variable. Or the protocol is missing in the output and there is a broken link. Would have to change the plugin php code to check if it is secure or not and adjust the URL accordingly I suppose. Using https all the time should work for most people. Only case I can think of is someone behind a corporate proxy that is blocking port 443.

Tragen:
Sorry. I was wrong. You have to remove only http so the link starts with //
It's a protocol relative url from RFC 3986
http://stackoverflow.com/questions/4071117/uri-starting-with-two-slashes-how-do-they-behave
http://www.paulirish.com/2010/the-protocol-relative-url/

PR #5295, opened 2016-05-31
Added an Option to Hide Organization from the New User Dialog

rmcaninch:
First try at this. I'm not sure if I got all the pieces right, but it is working for me.

### DIFF
--- a/plugins/new_user_dialog/config.inc.php.dist
+++ b/plugins/new_user_dialog/config.inc.php.dist
@@ -1,0 +1,10 @@
+<?php
+
+// New User Dialog Plugin Options
+// -----------------------
+// The Display Name and Signature are fields everyone should fill out
+//
+// Set this to true if you want to hide the organization field from 
+// the dialog. Default false to match original behavior.
+//
+$config['new_user_dialog_disable_organization'] = false;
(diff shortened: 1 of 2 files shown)


alecpl:
You still have the organization field in identity form in settings. So, what's the point? Or you use identity_form hook to hide it there? Then, maybe would be better to provide a plugin hook in new_user_dialog?

rmcaninch:
I just wanted to hide the field in the new user dialog so as not to distract my users. Having it remain in the settings page is fine. Is there a better way to approach this then?

alecpl:
This just does not sound right to me. Organization is distracting but signature is not?

rmcaninch:
Well in my case I'm deploying Roundcube for a company where everyone would have the same Organization and having it in the new user dialog doesn't seem to do any good for me so I wanted to hide it. The simpler the better :) And thought someone else might find it useful to turn off a field in the dialog. Could add an option to turn off the signature too. Not a big deal.

alecpl:
Well, for such cases I'd prefer a new core config option (e.g. identities_organization='Organization name') which would hide organization field in identity form and new_user_dialog and store configured value in the identity. Would you like to provide a PR for this?

rmcaninch:
I will try to look at that.

Issue #5291, opened 2016-05-30
Contact entries with multiple e-mail addresses not recognized as "known senders"


sffax:
Since upgrading to Roundcube 1.2.0, contact entries with multiple e-mail addresses don't seem to be recognized as "known senders". Although I have the "Display remote inline images" configuration parameter set to "from known senders", I'm always getting the warning message bar (To protect your privacy, remote images are blocked in this message.) when viewing their e-mail content. Clicking the "Always show images from..." button will add them as a separate contact entry in the address book although that particular e-mail address is already listed in a separate (previous) contact object.

alecpl:
Confirmed. This is because the separator on update is comma, but when we search we use space.

PR #5298, opened 2016-06-01
Fixed fetching photo for non primary email

prodrigestivill:
Now fails to fetch a profile photo when its not the primary e-mail (first contact e-mail).
Changed to a different type of search.


alecpl:
Works for me w/o this patch. And the query is ```AND ((((`email` LIKE 'test@alec.pl' OR `email` LIKE 'test@alec.pl,%' OR `email` LIKE '%,test@alec.pl,%' OR `email` LIKE '%,test@alec.pl'))))```

prodrigestivill:
Looks like it is already fixed in master branch.
This bug was seen in 1.2.0 final. But I see it has been already fixed yesterday in master for #5291.

Thanks!!!
Then I close this pull.

PR #5303, opened 2016-06-02
Honor dont_override setting for plugin config

pimlie:
Because I like to have the configuration of my plugins all in once place, this fix would help me do that. But maybe this is a security concern, as a plugin can now override all user preferences with it's own default values. Let me know what you think.

### DIFF
--- a/program/lib/Roundcube/rcube_config.php
+++ b/program/lib/Roundcube/rcube_config.php
@@ -425,6 +425,12 @@ class rcube_config
     public function merge($prefs)
     {
         $prefs = $this->fix_legacy_props($prefs);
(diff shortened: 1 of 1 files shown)


pimlie:
Maybe we should also either unset $prefs['dont_override'] before merging, use array_merge_recursive or implement a manual merge where we can choose which variables to overwrite and which to combine.

alecpl:
dont_override is a global option and should be placed in main config file. I think allowing it in multiple places can generate a confusing behaviour. Some users could expect it to overwrite the original and not be an extension of it.

thomascube:
I'm with @alecpl here. This is a global option and I don't see a need to introduce some merging magic for this.

PR #5323, opened 2016-06-14
Add $field_blocks to contact_form hook

Takika:
Add $field_blocks to contact_form hook, so plugins can change the order of the fields: https://github.com/roundcube/roundcubemail/pull/5281


thomascube:
`$field_blocks` is only relevant for the 'head' section. Thus better name the property something like `head_fields` in the plugin hook arguments.

Takika:
Renamed the $field_blocks variable to the better $head_fields :)

PR #5330, opened 2016-06-17
Do not mask fatal error when unable to load PEAR class

mejofi:
A failure to load PEAR consistutes a fatal PHP error, and @-loading it prevents that error from being logged, making debugging problems with loading the PEAR class really difficult. It should therefore be loaded without the @ sign, allowing the fatal error to end up in the logs, if it occurs.

### DIFF
--- a/program/lib/Roundcube/bootstrap.php
+++ b/program/lib/Roundcube/bootstrap.php
@@ -99,7 +99,7 @@ if (!preg_match($regexp, $path)) {
 spl_autoload_register('rcube_autoload');
 
 // set PEAR error handling (will also load the PEAR main class)
-@PEAR::setErrorHandling(PEAR_ERROR_CALLBACK, 'rcube_pear_error');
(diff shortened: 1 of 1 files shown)


alecpl:
It was there for a reason. See discussion in https://github.com/roundcube/roundcubemail/commit/4ab77dd13e106553a7c92d24aaa6799935517a8b. For 1.2 we could consider this change, but I'm not sure about 1.1.

urc:
I saw the discussion, and it seems odd to me, because you're overriding the error logging to exclude `E_DEPRECATED` anyway, on line 29. So why exactly would it throw anything but the fatal error, without the `@` sign?

alecpl:
Good point. I don't remember, but I see that in Roundcube 1.0 E_DEPRECATED wasn't excluded, so it probably comes from there. I'm going to accept this PR for 1.1, 1.2 and master.

PR #5364, opened 2016-07-21
LDAP password policy

ZiBiS:
The password driver to  appropriate change password when  the PPolicy overlay is installed in openLDAP server.
In plugins/password/drivers/ldap.php i've made small change to work properly with  Net_LDAP2 in 2.2.0 version.

### DIFF
--- a/plugins/password/drivers/ldap_ppolicy.php
+++ b/plugins/password/drivers/ldap_ppolicy.php
@@ -21,12 +21,12 @@ class rcube_ldap_ppolicy_password
     	$this->debug = $rcmail->config->get('ldap_debug');
     	 
     	$cmd    = $rcmail->config->get('password_ldap_ppolicy_cmd');
-        $uri    = $rcmail->config->get('password_ldap_uri');
-    	$baseDN = $rcmail->config->get('password_ldap_basedn');
-        $filter = $rcmail->config->get('password_ldap_search_filter');
-        $bindDN = $rcmail->config->get('password_ldap_searchDN');
-        $bindPW = $rcmail->config->get('password_ldap_searchPW');
-        $cafile = $rcmail->config->get('password_ldap_cafile');
+        $uri    = $rcmail->config->get('password_ldap_ppolicy_uri');
+    	$baseDN = $rcmail->config->get('password_ldap_ppolicy_basedn');
+        $filter = $rcmail->config->get('password_ldap_ppolicy_search_filter');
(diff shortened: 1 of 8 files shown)


plugins/password/drivers/ldap_ppolicy.php
```diff
@@ -0,0 +1,93 @@
...
+        	$result = stream_get_contents($pipes[1]);
+        	fclose($pipes[1]);
+        	
+        	$this->_debug('Result:'.$result);
+        	
+        	switch ($result) {
+        		case "OK":
+        			return PASSWORD_SUCCESS;
+        		case "Password is in history of old passwords":
+        			return  PASSWORD_IN_HISTORY;
```
alecpl: PASSWORD_IN_HISTORY constant does not exist. Did you forget to commit password.php changes?

ZiBiS: Yes, you're right. I've made new commit with it.

plugins/password/drivers/ldap_ppolicy.php
```diff
@@ -0,0 +1,93 @@
...
+        $bindPW = $rcmail->config->get('password_ldap_searchPW');
+        $cafile = $rcmail->config->get('password_ldap_cafile');
+
+        $descriptorspec = array(
+        		0 => array("pipe", "r"),  // stdin is a pipe that the child will read from
+        		1 => array("pipe", "w"),  // stdout is a pipe that the child will write to
+        		2 => array("file", "/tmp/ppolicy_ldap.err", "a") // stderr is a file to write to
+        );
+        
+        $cwd = 'plugins/password/drivers/';
```
alecpl: Shouldn't this be a path to helpers directory? Also does the command need to be configurable at all?

ZiBiS: I changed it to logs directory. This can be useful in case of any errors.

alecpl: First of all you should use log_dir option to get the logs dir location. Second, I was talking about 'plugins/password/drivers/'.

ZiBiS: I'm sorry, I missed that path is not correct and log_dir should be get in propper way now.

alecpl:
Last thing, all configuration options of this new driver must have prefix "password_ldap_ppolicy_". Please also solve the merge conflict.

Issue #5368, opened 2016-07-26
Update pt-BR translations and files


filhocf:
Hi

I'm using RC 1.2 and I'm also translator to pt-BR in Transifex.

Looking the code, I saw that in $RCROOT/program/localization/[en_US|pt_BR] directories have different number of files and in each file, have different number of strings.

In pt-BR dir haven't the file csv2vcard.inc. The pt-BR messages have 1 string less that in en-US, and in labels.inc file for pt-BR have 600 lines and 668 lines to en-US.

Can you update this files and/or update the transifex files to pt-BR, if it is the case?

Regards,
Claudio Ferreira

alecpl:
We always update localization files before a release. This we also did a few days ago when preparing 1.2.1 release.

As for the csv2vcard.inc file, it is not available on Transifex per purpose. This is not a simple translation file, it contains labels from other mail clients (Outlook, Thunderbird, etc.), so real labels have to be there. If you have this file translated, create a PR.

PR #5374, opened 2016-07-28
Adding localized in pt_BR csv2vcard.inc file

filhocf:
As suggested in #5368. Other files are ok. What changes is the header of each file.


alecpl:
Ok, but why only 10 labels is translated?

PR #5392, opened 2016-08-10
Convert charset wieh file based backend.

tsnr:


tsnr:
This patch fix a character corruption with file based backend when those charset is not ASCII.

PR #5393, opened 2016-08-10
Add retry if the imap server simply closes the connection on us

DanCld:
While operating RoundCube on our local servers running Courier-Imap, error messages like "Empty startup greeting (localhost:143)" were noted in the log while the user experience got affected (e.g. "message 1 of 0" texts appeared in the interface instead of "message 1of 15" or whatever). Sometimes these messages were shown to the user.

It was determined that in our scenario a limitation got enforced, no more than "N" connections from the same IP address were allowed by the IMAP server, additional attempts being simply rejected by connection reset. It was also determined that all these connections were caused by RoundCube instances opened by multiple users and also that the active connections were very short lived (while the php page loaded).

An obvious solution came to us, adding some retry mechanism to the IMAP connection in this particular scenario alone. I hope this may be useful for other people, too.


program/lib/Roundcube/rcube_imap.php
```diff
@@ -153,7 +161,10 @@ public function connect($host, $user, $pass, $port=143, $use_ssl=null)
             }
 
             $this->conn->connect($data['host'], $data['user'], $pass, $data);
-        } while(!$this->conn->connected() && $data['retry']);
+
+        // retry the connection if the current attempt is unsuccessful, the error code is RETRY and
+        // the attempt limit is not met
+        } while(!$this->conn->connected() && $this->conn->is_retry_error() && $attempt < 5);
```
AllanKlaus: This `$attempt < 5` could be a env var because the developer can set it in the config file without change core, by default it could be '5'

DanCld: Yes and No. Let's take a step back from this particular value. What I have tried to achieve here is to try to save the user experience in a particular, not sure how frequent scenario.

The purpose of the "connect" functionality is to get an IMAP connection opened and authenticated within a reasonable time frame (worst case scenario, before the browser timeout, normal case - before the user gets bored). We can define this time frame (e.g. 200ms, or 500ms, or 5 seconds or whatever) or - the current situation, without this pull request - ignore the situation altogether and just try a connection (succeed / fail). If we consider retrying a good strategy in whatever scenario we have, the best way to do it is, indeed, to try to fit the whole process within a predefined time frame.

Unfortunately this raises a different issue. We must retry with randomized or growing delays in order to both maximize the chances that we get a connection and (not likely but plausible) to avoid overloading the IMAP server. A randomized delay, coupled with a tight time frame, results in an unpredictable retry count. We could either live with that, or do what I did, hard code the retry count to 3-4-5-10 and just have it like this.

As about setting it to a different value: there is a "sweet spot", as there obviously are 2 extreme values (1 - the situation right now / some N - timeout or user gets bored and closes the window). My point is - what to choose? How would anyone know a certain value is better than the other? It's not like some mathematical function here, it's more like:

- 99% of the time the connection will be established upon the 1st attempt
- 99,8%   .... 2nd ...
- 99,92% .... 3rd ...
- 99,96% .. 4th...
(I made the figures up but it's a diminishing returns game after all).

These are my 2 cents.

alecpl:
I don't like it. It looks like a rare case. The fact that you catch only "startup greeting" error makes the solution even less universal. I propose to modify 'storage_connect' hook by adding ```'conn' => $this->conn``` argument. This way you would be able to write a plugin that checks $args['conn']->data['GREETING'] and sets $args['retry'] accordingly. Would that work for you?

DanCld:
Will think about a better solution, closing this one in the mean time.

Issue #5394, opened 2016-08-11
Missing locker icon in compose bar


sixtyfive:
I've installed RoundCube 1.2.1 (`git clone`; `git checkout release-1.2` instead of a zip file/etc. if it matters) and am using it with my old 1.1 `config.inc.php`. In Chromium, I've installed and configured Mailvelope and enabled it's "API" switch for the RCM installation's domain. I can receive and decrypt emails by way of Mailvelope fine. In RCM's compose window, however, no locker icon is shown; there doesn't seem to be any UI present for encrypting emails.

alecpl:
Do you use the default Larry or classic skin? Did you really activated the API? Activating the Mailvelope plugin on the page is not enough. You have to go to Options > List of Email Providers > <your roundcube domain> Edit and enable API there.

sixtyfive:
I used Larry but then switched to Classic. Neither show a locker icon. I'm aware that the API needs to be enabled manually. The automatic entry in Mailvelope was for `*.subdomain.domain.tld`. I enabled the API there. Just to be sure, I also created a second entry `*.domain.tld` and enable the API there, which didn't change anything either.

alecpl:
I'm afraid it might be mailvelope issue. Try without * in the name.

alecpl:
What Mailvelope version?

alecpl:
Works for me with Chromium 51 and Mailvelope 1.5.1.

sixtyfive:
According to Chromium's Extensions page, it's Mailvelope 1.5.1. I tried to create another E-mail Provider in Mailvelope using no * in its Domain pattern at all, but that causes Mailvelope to complain ("Domain match pattern should begin with *"). What I find curious is that email _de_cryption works just fine.

alecpl:
Decryption may work with disabled API, I suppose. Did you check browser console for errors? What if you use +Add button in Mailvelope menu instead providing domain name?

sixtyfive:
I deleted all E-mail Providers. I then reloaded RCM and clicked Mailvelope's "Add" button, went into the Mailvelope settings and enabled API for the newly created entry, which has the pattern "*.subdomain.domain.tld". I then reloaded RCM again while looking at Chromium's console. There was one error due to the uBlock extension, which I disabled, afterwards reloading RCM once more. No errors remain in the console, but there is still no locker icon present.

sixtyfive:
[...]

PR #5396, opened 2016-08-11
Adding warning about keeping new default skins versions

sixtyfive:
(See https://github.com/roundcube/roundcubemail/issues/5394#issuecomment-239166245 for the reasoning)


alecpl:
I don't think we need such a note. The point is to overwrite skins with new version. It looks like you did the opposite thing.

sixtyfive:
For the simple reason that I'm inexperienced with rsync. Isn't that what documentation is there for, to foresee possible problems and give advice on how to avoid them?

alecpl:
Yes, but why should treat the skins folder different? The same applies to all folders.

Issue #5397, opened 2016-08-11
Improve cache busting strategy


fabioperrella:
Hi,

we are using roundcube as the base of our webmail app.

We noticed that the cache busting strategy (using querystring `?s=timestamp`) is not working so well

A better strategy would be changing the asset file name, for example:

program/js/jquery-1469707857.min.js

In our code, we changed the method `file_mod` in class `rcmail_output_html` to do so and also changed nginx configurations to "redirect" (or rewrite) these requests to the correct file in filesystem using a `location` rule.

If you agree, I can send a merge request with this implementation

alecpl:
We've had such a request (for css files) in the past (#3927) and it was decided we don't want it in core. It should be possible to write such a plugin.

fabioperrella:
So what about add a hook inside method `file_mod`? Without this, I think is not possible to do so via plugin, is it?

alecpl:
You can use render_page or send_page hooks.

alecpl:
Create PR with the new hook and we'll consider this.

fabioperrella:
ok, great!

PR #5406, opened 2016-08-19
Added hook to be able to modify assets cache busting strategy

fabioperrella:
Added hook to be able to modify assets cache busting strategy using a plugin

refering to #5397

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -869,7 +869,7 @@ EOF;
             'file_timestamp' => $file_timestamp
         ));
 
-        if (isset($hook_response['file_with_cache_buster']) && $hook_response['file_with_cache_buster']){
+        if (isset($hook_response['file_with_cache_buster']) && !empty($hook_response['file_with_cache_buster'])){
             return $hook_response['file_with_cache_buster'];
         }
 
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -90,7 +90,7 @@ class rcmail_output_html extends rcmail_output
 
         $lic = <<<EOF
 /*
-        @licstart  The following is the entire license notice for the 
+        @licstart  The following is the entire license notice for the
         JavaScript code in this page.
 
         Copyright (C) 2005-2014 The Roundcube Dev Team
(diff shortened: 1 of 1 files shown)


program/include/rcmail_output_html.php
```diff
@@ -852,17 +852,30 @@ protected function file_mod($file)
...
         return $file;
     }
 
+    private function file_with_cache_buster($file_name, $file_timestamp){
+        $hook_response = $this->app->plugins->exec_hook('file_with_cache_buster', array(
+            'file_name' => $file_name,
+            'file_timestamp' => $file_timestamp
+        ));
+
+        if (isset($hook_response['file_with_cache_buster']) && $hook_response['file_with_cache_buster']){
```
AllanKlaus: You could use !empty($hook_response['file_with_cache_buster']) because without it php can launch notice

thomascube:
Looks good to me. @alecpl any objections?

program/include/rcmail_output_html.php
```diff
@@ -852,17 +852,30 @@ protected function file_mod($file)
...
+            $file = $this->file_with_cache_buster($file, $fs);
         }
 
         return $file;
     }
 
+    private function file_with_cache_buster($file_name, $file_timestamp){
+        $hook_response = $this->app->plugins->exec_hook('file_with_cache_buster', array(
+            'file_name' => $file_name,
+            'file_timestamp' => $file_timestamp
```
alecpl: I'd change file_name and file_timestamp to just name and timestamp.

program/include/rcmail_output_html.php
```diff
@@ -852,17 +852,30 @@ protected function file_mod($file)
...
+        ));
+
+        if (isset($hook_response['file_with_cache_buster']) && !empty($hook_response['file_with_cache_buster'])){
+            return $hook_response['file_with_cache_buster'];
```
alecpl: s/file_with_cache_buster/result/

alecpl: and isset() is not needed, !empty() is enough.

program/include/rcmail_output_html.php
```diff
@@ -852,17 +852,30 @@ protected function file_mod($file)
...
         if ($fs = @filemtime($this->assets_dir . $file)) {
-            $file .= '?s=' . $fs;
+            $file = $this->file_with_cache_buster($file, $fs);
         }
 
         return $file;
     }
 
+    private function file_with_cache_buster($file_name, $file_timestamp){
+        $hook_response = $this->app->plugins->exec_hook('file_with_cache_buster', array(
```
alecpl: I don't like the hook name. How about template-file-cache-buster?

Issue #4976, opened 2007-12-04
dereferrer for links in mails


rcubetrac:
_Reported by igor on 4 Dec 2007 19:13 UTC as Trac ticket #1484686_

a dereferrer would be nice, because its not nice that other website owners could look witch email-id it is, or wich domain...

I use at the moment a extern one.

its very easy...

index.php

```
<?php
function tpl ($template)
  {
    if (file_exists ('tpl/' . $template . '.html'))
    {
      include 'tpl/' . $template . '.html';
    }
  }

  function logfile ()
  {
    global $link_url;
    global $config;
    if ($config[== 1)
    {
     $fp = fopen ('log/logfile.txt', 'a');
     fwrite($fp, "$link_url");
     fwrite($fp, "\n");
     fclose ($fp);
    }
  }

  $config['prcl']('log']) = 'http';
  $config[= 0;
  $config['log']('time']) = 1;
  $link_url = urldecode($_SERVER[ if ($link_url != '')
  {
  /*  if (substr($link_url, 0, 7)!='http:_') || substr($link_url, 0, 8)!='https:_' || substr($link_url, 0, 6)!='ftp://'
    {
      $link_url = $config['prcl']('QUERY_STRING']);
) . '://' . $link_url;
    }
    */
    $link_head = $link_url;
    $link_time = $config['time'];
    logfile ();
    tpl ('page_redirect');
    exit ();
  }

  tpl ('page');
  exit ();

  ?>
```

Template:

```
<?
global $link_head;
global $link_time;
?>
<?
global $link_head;
global $link_time;
?>
<!DOCTYPE html PUBLIC "-_W3C_DTD XHTML 1.0 Transitional_EN" "http:_www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml"><head>
<head>
<title>Weiterleitung zu <?=$link_head?></title>

<meta http-equiv="refresh" content="<?=$link_time?>; URL=<?=$link_head?>">

</head>
<body>


<p align="center" valign="center">
    Bitte warte <?=$link_time?> Sekunden <br>
    Umleitung auf: <a href="<?=$link_head?>"><?=$link_head?></a>
</p>
</body>
</html>
```

it would be nice if it could be includet.
new window like "print view" with meta refresh
and a link like _action=deref&target=http://test.com/

_Migrated-From: http://trac.roundcube.net/ticket/1484686_

[...]

PR #5422, opened 2016-09-01
prevent referrer leaking

ghost:
For enhanced privacy, referrer leaking should be avoided.

### DIFF
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -84,3 +84,11 @@ $config['plugins'] = array(
 
 // skin name: folder from skins/
 $config['skin'] = 'larry';
+
+// referrer setting, possible values:
+// no-referrer
+// no-referrer-when-downgrade
+// origin
+// origin-when-cross-origin
+// unsafe-url
+$config['referrer'] = 'no-referrer';
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -88,6 +88,7 @@ $config['skin'] = 'larry';
 // referrer setting, possible values:
 // no-referrer
 // no-referrer-when-downgrade
+// same-origin
 // origin
 // origin-when-cross-origin
 // unsafe-url
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -1211,3 +1211,12 @@ $config['message_show_email'] = false;
 // 0 - Reply-All always
 // 1 - Reply-List if mailing list is detected
 $config['reply_all_mode'] = 0;
+
+// referrer setting, possible values:
+// no-referrer
+// no-referrer-when-downgrade
+// same-origin
+// origin
+// origin-when-cross-origin
+// unsafe-url
+$config['referrer'] = 'no-referrer';
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1578,6 +1578,7 @@ EOF;
         $referrer_values = array(
             "no-referrer",
             "no-referrer-when-downgrade",
+            "same-origin",
             "origin",
             "origin-when-cross-origin",
             "unsafe-url");
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1573,8 +1573,11 @@ EOF;
             $page_header.= $this->charset . '" />'."\n";
         }
         
-        // prevent referrer leaking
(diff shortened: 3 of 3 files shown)


alecpl:
Any examples of a referrer leak in Roundcube?

evildead45:
I read about a website privacy check tool at https://webbkoll.dataskydd.net/en in a blog (https://www.kuketz-blog.de/webbkoll-wie-datenschutzfreundlich-ist-deine-webseite/#more-49964) and tried it on my instance of roundcube. It reported "referrers leaked" to me.

This topic is more like a general privacy aspect. Please find more thoughts about this at https://go-to-hellman.blogspot.de/2015/06/protect-reader-privacy-with-referrer.html

I would like to have the websites I host on my server as privacy friendly as possible. Avoiding referrer leaking is another step into this direction.

CRtEurope:
The referer is not sent for HTTPS connections, per RFC 2616, so there's yet another good reason for just using HTTPS

evildead45:
My setup is HTTPS only and it still reported "referrers leaked".

I think you refer to the following passage of the RFC:

> 15.1.3 Encoding Sensitive Information in URI's
   [...]
   Clients SHOULD NOT include a Referer header field in a (non-secure)
   HTTP request if the referring page was transferred with a secure
   protocol.
   [...]

This would mean it is allowed to send the referrer to a HTTPS request.

alecpl:
My point was that we already have some referrer protection code. So, I'd like to see a use-case or maybe web browser version when it does not work. I see we have #4976 still open, but I'm not sure the solution proposed here is right.

evildead45:
As mentioned, the test suite at https://webbkoll.dataskydd.net/en complained about referrer leaking and suggests adding `<meta name="referrer" content="no-referrer">` to the `<head>` section of the page. They also link to http://caniuse.com/#feat=referrer-policy, which lists the browsers that support this setting.

CRtEurope:
Very strange according to RFC 2616...
Nevertheless, due to security whe should not wait until something happens... and should implement it

CRtEurope:
I have just tested your code @evildead45 - It is working great. Thank you very much this.

evildead45:
Thanks a lot! I just added a check to only allow valid values for the referrer.

alecpl:
I'm curious if that test suite tests real leaks or just checks for the meta tag existence ;)

Anyway, I'm not in opposition to this idea, but maybe we could improve it:
1. Some users may actually want referrer to work on the login screen. So, maybe better to use this only for authenticated pages.
2. We already use noreferrer for links and no one complained, so maybe making this new feature configurable is useless. I propose to always use "no-referrer" value for best privacy.
3. Some browsers (Edge, Safari) does not support "no-referrer" (and other values), but support "never", it would be good to set the value according to detected browser.
4. Finally, .htaccess file already contains an example how to do the same with Content-Security-Policy header.

evildead45:
@alecpl 
1: Can you give me an example please, when you might like a referrer on the login page? What could be the benefit? I personally would like to have the "no-referrer" everywhere, so maybe we could introduce two settings if need be: one for authorised and another for unauthorised pages. Maybe with the setting for unauthorised defaulting to the setting for authorised?
2: Maybe we could have it configurable with default to no-referrer? So people that do want to change the behaviour could do so and those who don't benefit from best privacy.
3: I think, browsers should stick to the referrer-policies listed at https://www.w3.org/TR/referrer-policy/#referrer-policies. For Edge there is a ticket stating the same: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7119603/
4: Yes, CSP could be used as well. Also the "Referrer-Policy" header could be an option.

alecpl:
@evildead45 if you use default login page keeping referrer on logon page would not change anything for you (as there will be no external connections), but if someone modified login page adding some (commercial) content there he might be interested in how it works from marketing perspective. Of course I can imagine an opposite need. So, a three-state option would be enough (all, auth, none).

Referrer-Policy is a new thing and not many browsers support it. CSP is more common.

evildead45:
@alecpl I agree to both. When I find time I will look into it.

thomascube:
Looks like an extended version of #5674 but using a `meta` tag instead of a HTTP header. Both seems to be suggested by https://www.w3.org/TR/referrer-policy/#referrer-policy-delivery-meta.

I'd vote for using headers in favor of a `meta` tag.

alecpl:
Implemented in 186f21c4c.

PR #5450, opened 2016-09-29
Fixed spelling

smares:


alecpl:
I fixed this on Transifex too.

PR #5464, opened 2016-10-08
Update ru_RU.inc

xm74:
Misprint corrected


alecpl:
Thanks. In future please fix localization issues on Transifex.

PR #5468, opened 2016-10-11
A plugin to check user password expiration in LDAP

ZiBiS:
In case of user password expiration or be in warning period, user will be redirected to change password with suitable message.


alecpl:
I don't think we want to support another very-specific plugin by including it into our repository, but we'll consider your changes to the password plugin.

ZiBiS:
Ok, i will move the plugin outside the repo and it will be good if the password plugin can handle parameters from it.

plugins/password/password.php
```diff
@@ -95,7 +95,13 @@ function password_init()
         if (rcube_utils::get_input_value('_first', rcube_utils::INPUT_GET)) {
             $rcmail->output->command('display_message', $this->gettext('firstloginchange'), 'notice');
         }
-
+        if (rcube_utils::get_input_value('_passwdexpwarning', rcube_utils::INPUT_GET)) {
+        	$expdate = rcube_utils::get_input_value('_passwdexpdate', rcube_utils::INPUT_GET, true);
+        	$rcmail->output->command('display_message', $this->gettext('passwdexpirewarning'). ': '. $expdate, 'warning');
```
alecpl: I propose here to:
1. use a localization message with variables, so date is not concatenated to the localized text.
2. print date in configured date_format.

ZiBiS: What do you think about passing expiration date/time in the session data instead of passing it via url ($args array)?

ZiBiS: Indeed, you're right, the only one parameter is needed. Can i save in session datetime object, or only formatted datetime text?

plugins/password/password.php
```diff
@@ -95,11 +95,13 @@ function password_init()
...
-        if (rcube_utils::get_input_value('_passwdexpwarning', rcube_utils::INPUT_GET)) {
-        	$expdate = rcube_utils::get_input_value('_passwdexpdate', rcube_utils::INPUT_GET, true);
-        	$rcmail->output->command('display_message', $this->gettext('passwdexpirewarning'). ': '. $expdate, 'warning');
-        } elseif (rcube_utils::get_input_value('_passwdexpired', rcube_utils::INPUT_GET)) {
-        	$rcmail->output->command('display_message', $this->gettext('passwdexpired'), 'error');
+
+        if (rcube_utils::get_input_value('_passwdexpired', rcube_utils::INPUT_GET)) {
+            $rcmail->output->command('display_message', $this->gettext('passwdexpired'), 'error');
+        } 
+        else if (isset ($_SESSION[passwdexpdatetime])) {
```
alecpl: Still, why not use one variable/argument? Mixing GET and session for the same purpose does not look nice.
ps. strings need to be surrounded by single quotes

alecpl: Use one argument either via GET or session, but one.

alecpl:
I slightly modified it and merged. Thanks.

Issue #5414, opened 2016-08-25
managesieve advanced editor


4nd3r:
one of the best features of https://github.com/JohnDoh/Roundcube-Plugin-SieveRules-Managesieve is ability to use "advanced editor" to view and edit sieve rules in "raw". since that plugin is no longer maintained, please consider adding this feature to managesieve plugin (with ability to enforce advanced editor). without that it's impossible to edit more complex sieve rules or activate special server-side filters. thank you!

Andrei-Pozolotin:
+1

Andrei-Pozolotin:
see #5416 #5417 #5418

Andrei-Pozolotin:
@4nd3r 
1) I think this should really be called `basic editor` :-)
2) do you think we could pull a PR here for raw edit mode?

Andrei-Pozolotin:
@JohnDoh can you please offer any advice on this?

4nd3r:
@Andrei-Pozolotin 
1) raw or basic or advanced - the name is not important here at the moment
2) this issue is blocker for me and "my users", so I probably hack something together before end of this year. not sure about PR because my PHP is really rusty since I mostly use it as tool aka get things done :-)

Andrei-Pozolotin:
@4nd3r if you hack something before I do, please share here, and I'll do the same.

Andrei-Pozolotin:
@alecpl please review #5473

alecpl:
PR merged and improved. Done.

Andrei-Pozolotin:
thank you

PR #5473, opened 2016-10-12
implemented raw editor for sieve filter sets

hansmaulwurf23:
This could solve #5414. I don't hide the filterslist since the user will be able to cancel the editing that way. It's also possible to add some syntax highlighting (maybe via https://ace.c9.io).


Andrei-Pozolotin:
@hansmaulwurf23 Martin: since there is no response from @alecpl Aleksander,  do you think we should make this into a plugin instead?

plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
```diff
@@ -481,10 +482,55 @@ function actions()
...
+        // check request depth limits
+        else if ($max_depth && count($_POST['_header']) > $max_depth) {
+            rcube::raise_error(array(
+                    'code' => 500, 'type' => 'php',
+                    'file' => __FILE__, 'line' => __LINE__,
+                    'message' => "Request size limit exceeded (one of suhosin.request.max_array_depth/suhosin.post.max_array_depth)"
+            ), true, false);
+            $this->rc->output->show_message('managesieve.filtersaveerror', 'error');
+        }
+
```
alecpl: You don't need all those checks above. There's no big array in POST here.

plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
```diff
@@ -1246,6 +1295,27 @@ function filter_frame($attrib)
     {
         return $this->rc->output->frame($attrib, true);
     }
+    
+    function filterset_editraw($attrib) 
+    {
+        $scriptName = isset($_GET['_set']) ? $_GET['_set'] : $_POST['_set'];
+    	$script      = $this->sieve->get_script($scriptName);
```
alecpl: $script can be False here. Maybe we should handle that more nice.

plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
```diff
@@ -1246,6 +1295,27 @@ function filter_frame($attrib)
...
+    	
+    	$hiddenfields = new html_hiddenfield(array('name' => '_task', 'value' => $this->rc->task));
+    	$hiddenfields->add(array('name' => '_action', 'value' => 'plugin.managesieve-saveraw'));
+    	$hiddenfields->add(array('name' => '_set', 'value' => $scriptName));
+    	$hiddenfields->add(array('name' => '_seteditraw', 'value' => 1));
+    	$hiddenfields->add(array('name' => '_framed', 'value' => ($_POST['_framed'] || $_GET['_framed'] ? 1 : 0)));
+    	
+    	$out .= $hiddenfields->show();
+    	
+    	$out .= '<textarea id="rawfiltersettxt" name="rawsetcontent" rows="15">' . $script . '</textarea>';
```
alecpl: $script need to be "escaped" here, e.g. htmlentities($script, ENT_NOQUOTES, RCUBE_CHARSET), or better use html_textarea class.

plugins/managesieve/managesieve.php
```diff
@@ -225,6 +226,23 @@ function managesieve_save()
...
+    /**
+     * Raw form save action handler
+     */
+    function managesieve_saveraw()
+    {
+        // load localization
+        $this->add_texts('localization/', array('filters','managefilters'));
+    
+        // include main js script
+        if ($this->api->output->type == 'html') {
```
alecpl: This check is not needed, there will be no ajax here.

plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
```diff
@@ -1299,22 +1274,33 @@ function filter_frame($attrib)
...
+                    'rows' => '15'
+            ));
+            $out .= $txtarea->show($script);
+            //$out .= '<textarea id="rawfiltersettxt" name="rawsetcontent" rows="15">' . $script . '</textarea>';
+            
+            $this->rc->output->add_gui_object('sievesetrawform', 'filtersetrawform');
+            return $out;
+        } else {
+            $this->rc->output->show_message('managesieve.filterunknownerror', 'error');
+            return '';
```
hansmaulwurf23: Not sure if this is enough, since the iframe is rendered and the save button is shown. Showing the error and reloading the page is probably better, but I don't know how to achieve that. Maybe you can help out ;-)

PR #5480, opened 2016-10-17
Update virtualmin.php

bialy7:
Correcting username if password_virtualmin_format is set to format 4.


plugins/password/drivers/virtualmin.php
```diff
@@ -51,6 +51,7 @@ function save($currpass, $newpass)
             break;
         case 4: // username-domain
             $domain = substr(strrchr($username, "-"), 1);
+            $username = current(explode("-", $username));
```
alecpl: Hmm.. and what if you have domains with '-' in the name? I'm asking because I have no idea about how virtualmin handles this case.

alecpl:
See #5759.

PR #5490, opened 2016-10-27
Make managesieve default headers configurable

tpayen:
Hi,
That would be great if we could add more headers for our users with an option in manageieve plugin.
Thanks


narcoticfresh:
Will this ever get merged or is something wrong about it?

alecpl:
Slightly improved version in c26c85629ce.

Issue #5518, opened 2016-11-14
Managesieve: RAW editor improvements


alecpl:
It would be nice to have code highlighting and line numbering. There are two javascript editors worth looking at, it's Ace and CodeMirror. Ace have no Sieve support, so it looks like we'll use CodeMirror.

When we have line numbering we could also display precise errors from managesieve server which usually contain line numbers.

trippleflux:
I vote +1 for this being merged 👍

alecpl:
Done.

PR #5520, opened 2016-11-15
Implemented #5518 (Managesieve: RAW editor improvements)

hansmaulwurf23:
I added the latest version of Codemirror in a dedicated folder within the managesieve directory so it can be easily upgraded. ```initeditor.js``` and ```overrides.css``` are the only files not coming from the Codemirror package (was not sure about the best location). 
I gave the error handling you mentioned (#5518) a try as well: the returned message is now split into single messages and then the line numbers and error messages are extracted and stored in the environment. The editor is showing this errors by highlighting the lines and placing red dots in the gutter (holding the actual error message as tooltip).


alecpl:
Thanks you. Merged with some additional improvements and fixes.

PR #5530, opened 2016-11-25
Do not bother trying GSSAPI auth to imap when there is not a ccache available

lorissantamaria:
The imap server may offer GSSAPI auth, but roundcube may not be properly configured to use it, so we should ignore GSSAPI auth method when we do not have a ccache.


alecpl:
Thanks, fixed in 940a27400e1 with a little bit less of code.

PR #5539, opened 2016-11-29
Added support for %h and %d in poppassd host

paspo:
With this modification, you can use %d and %h in $config['password_pop_host'] for referring the user's domain or imap host.


alecpl:
You should rather use rcube_utils::parse_host(). Also, add a note to sample config file, please.

paspo:
@alecpl You're right.
That's because I looked at another password backend that had hostname variable replacement. I took the wrong one (directadmin.php) :)

Thanks

PR #5549, opened 2016-12-07
Add cPanel Webmail driver

mkllnk:
This driver does not use an admin account. It uses the user's account to
authenticate at the Webmail API. It uses the recommended UAPI and does
not rely on third-party classes.

@alecpl I wasn't sure about the readme part and the version number. What do you think?

### DIFF
--- a/plugins/password/README
+++ b/plugins/password/README
@@ -28,7 +28,9 @@
  2.3.  Poppassd/Courierpassd (poppassd)
  2.4.  LDAP (ldap)
  2.5.  DirectAdmin Control Panel (directadmin)
- 2.6.  cPanel (cpanel)
+ 2.6.  cPanel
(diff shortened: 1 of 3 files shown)


alecpl:
I'd add a note to the README that the driver requires php-curl. Other than that it looks good. Thanks.

troystevenhunt:
do what you think is best agent

JouiDavide:
This... I've spent the last 2 days editing the cPanel driver to use the new UAPI  (mainly because I really didn't want to enter the host username/password in config), and now I think well, let's post it on GitHub. And I find this ^^

Obviously I should've checked before. I see that it's been merged, is it in the new 1.3 update? Could I still suggest some modifications? Although mine is almost exactly the same.

PR #5551, opened 2016-12-07
Possibility of setting port and ssl by imap server

dioubernardo:
This option is very important when webmail needs to connect to servers with different configurations:

for example:
$config['default_host'] = 'mail.%s';
$config['imap_hosts_config'] = array(
   'mail.b.com.br' => array(
      'port' => 993,
      'ssl' => 'ssl'
  )
);

User user@a.com.br connects with mail.a.com.br on port 143 without ssl
User user@b.com.br connects with mail.b.com.br on port 993 with ssl

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -153,6 +153,15 @@ $config['imap_auth_type'] = null;
 // );
 $config['imap_conn_options'] = null;
 
+// IMAP Possibility of setting port and ssl per host
+// $config['imap_hosts_config'] = array(
+//    'mail.domain.tld' => array(
+//       'port' => 993,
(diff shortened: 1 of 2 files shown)


alecpl:
default_host already accepts hostnames with a port. The only problem with it set to an array is that a select field will appear on logon page. But I'd rather somehow fix this or extend imap_conn_options option.
Note that any such a change should also involve changes to managesieve.

dioubernardo:
Ok no problems, as I was using "mail.%s" on the default_host, I found it easier to create the other option.
Just to mention I have 8 different mail servers, all using the same roundcube. Only 1 uses different setting.

I do not know what managesieve is.

PR #5554, opened 2016-12-08
Password driver cpanel_webmail: Correct decoding of JSON response

mkllnk:
I'm sorry for the inconvenience, but the previous pull request choked on successful password changes. I had not tested it well enough. I hope my added unit test fits into you test suite.

The old version assumed `json_decode` would return an associative array.
While it can do so if an extra option is specified, the default
behaviour is to return an object. Therefore, a successful password
change resulted in an error while parsing the response.

The new code is accessing the response as object instead.

### DIFF
--- a/plugins/password/drivers/cpanel_webmail.php
+++ b/plugins/password/drivers/cpanel_webmail.php
@@ -28,6 +28,15 @@
 
 class rcube_cpanel_webmail_password
 {
+    /**
+     * Changes the user's password. It is called by password.php.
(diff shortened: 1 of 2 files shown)


alecpl:
json_decode() has 2nd argument. I think it's better to enable this argument and additionally check if the function result is an array. It may return NULL on error.

mkllnk:
You are right that I should check for NULL. I will add that.

Just curious: Why do you prefer an array over an object?

alecpl:
I does not matter really here.

PR #5565, opened 2016-12-15
[WIP / POC] Add user export script / user import and export functionality

duk3luk3:
Hello,

I am currently working on merging two existing roundcube installations into one.

One of them is running with an sqlite db backend and has been upgraded to roundcube 1.2.2. The other one is running with a mysql backend.

There is no easy way to convert either of those backends to another db, let alone the primary key transmogrifications necessary to merge them.

Since roundcube has no facilities for this yet, I am writing scripts to import/export the following user data:

* User records
* User identities
* User contacts (contactgroups + contacts + contact/group associations)

In addition to my usecase of merging roundcube instances, this will also allow switching database backends by exporting the users, setting up a new database, and then reimporting.

This PR currently implements a POC user export script for a potential future
user import/export feature.

I would like to request feedback as to whether
* This is a functionality the roundcube project is interested in merging
* The approach taken in this script looks promising
* There is any user data missing (caches and searches are intentionally ignored at this time; if an argument is made as to why they should be included I will certainly look into doing that.)

If no significant objections are raised, I intend to continue as follows:

* Fix up the codestyle
* Replace the `print_r` at the end by dumping to a suitable data format such as JSON
* Write a corresponding import/merge script

Thanks to the existing db upgrade infrastructure, the import/export scripts I'm writing will not have to "care" about data versions, they can simply be kept current with roundcube; both "source" and "target" roundcube instances can simply both be upgraded to the same (current) version and the import/export scripts should then work.

Thank you in advance for taking the time to read and review this PR!

### DIFF
--- a/bin/userexport.sh
+++ b/bin/userexport.sh
@@ -27,14 +27,20 @@ require_once INSTALL_PATH.'program/include/clisetup.php';
 
 function print_usage()
 {
-	print "Usage:  userexport -u username -h host -l limit\n";
+	print "Usage:  userexport -f file -u username -h host -l limit\n";
+	print "--file   Output file\n";
 	print "--user   IMAP user name\n";
 	print "--host   User IMAP host\n";
 	print "--limit  Limit\n";
 }
 
+function vputs($str)
+{
+	$out = $GLOBALS['args']['file'] ? STDOUT : STDERR;
+	fwrite($out, $str);
+}
 // get arguments
-$opts = array('u' => 'user', 'h' => 'host', 'l' => 'limit');
+$opts = array('u' => 'user', 'h' => 'host', 'l' => 'limit', 'f' => 'file');
 $args = rcube_utils::get_opt($opts);
 
 if ($_SERVER['argv'][1] == 'help')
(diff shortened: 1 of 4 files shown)


alecpl:
We have #5112 for this and we'd be interested.

As for the implementation:
- I'm not sure JSON is a good choice here. I'm not sure what, but something that does not require to load the whole file into memory on restore operation. You can work with JSON for now.
- we can ignore caches, but not data like preferences, contacts/groups, identities, dictionaries and searches.
- some plugins may add some more important tables, so we'd need some plugin api to handle this.

duk3luk3:
> I'm not sure JSON is a good choice here. I'm not sure what, but something that does not require to load the whole file into memory on restore operation. You can work with JSON for now.

Hmm... "line-based json" is the first thing that comes to mind (One JSON object per line / one record per line). Of course at that point it's not a standardized format anymore. I'm not sure what formats there are that php has a streaming parser for, I'm not a full-time php dev.

> we can ignore caches, but not data like preferences, contacts/groups, identities, dictionaries and searches.

OK, I will add searches and dictionaries (preferences are part of the user data). That's everything then, right?

> some plugins may add some more important tables, so we'd need some plugin api to handle this.

So adding a serialize / deserialize function to the plugin api?

alecpl:
I understand that such scripts will be used mostly on small installations, but we should design them for everyone. Because of that I think that creating big "user records" is not good. We should "dump" every table separately. And add some logic to import script to match that data with a user. So, dump will be much faster (only a few queries), and import should not be much slower that with your solution.

duk3luk3:
I've now added what I think should be good enough as a first version of the export script.

Dumping each table separately is directly in conflict with the requirement that the importing should be possible without loading all the data into memory at once, because if we have to reassemble the user data from the table relationships when importing we will need the whole user table in memory all the time and at least have each of the related tables in memory when adding data from those tables. It also means we have to use and map the primary keys. It would make everything a lot more complicated.

alecpl:
That is not true. We can process the file line by line and keep in memory only relevant data (keys map). We can INSERT immediately and goto another line. Of course, order of tables would be important.

duk3luk3:
With 61f8287 there's now actually something happening. It can now import users and contacts.

I had to extend some roundcube functions to expose things.

miaulalala:
@duk3luk3 could you rebase this and resolve the conflicts? Let's work together on this so we can get this in :wink:

PR #5584, opened 2017-01-07
avoid closure compiler suspicious code warning

johndoh:
../program/js/app.js:364: WARNING - Suspicious code. The result of the 'getprop' operator is not being used.
if (n++) try { this.contentWindow.document; ref.enable_command('print', true); }
^^^^^^^^^^^^^^^^^^^^^^^^^^^


alecpl:
Thanks for catching this. I fixed this with an if() statement.

PR #5623, opened 2017-01-27
Archive: Add Thunderbird compatible Month option

hjs404:
Hi,

This adds another option to the Archive plugin, which allows the user to divide the archive by months, but using the same subfolder format as Thunderbird, as shown below:

Thunderbird Archive Options

This means that the Archive folders of Roundcube and Thunderbird remain compatible and the user can use whichever of the clients to archive mail successfully and without having two folders for each month.

This is a tweak that I've found useful when implementing and transitioning to Roundcube in an environment that previously primarily used Thunderbird, so I thought other users may benefit from using it. I've done the English localizations, but I don't know any other languages, sorry!

Thanks! :)

### DIFF
--- a/plugins/archive/archive.php
+++ b/plugins/archive/archive.php
@@ -188,6 +188,12 @@ class archive extends rcube_plugin
                             . $delimiter . $rcmail->format_date($message->timestamp, 'm');
                         break;
 
+                    case 'thunderbirdmonth':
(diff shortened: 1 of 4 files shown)


RolandSM:
Is it possible that this change is not available in current roundcube 1.3.10???

alecpl:
Yes. It's only in 1.4.

PR #5639, opened 2017-02-12
Enable to change Content-Type charset and MIME header encoding

dfukagaw28:
This enables two changes:

- Add a configuration variable "default_head_encoding" which enables to change the encoding for MIME header (e.g., Subject) from Q-encoding (default) to B-encoding (preferred for multibyte languages).

- Change the "charset" value in the "Content-Type" MIME header if:

  - the value of $message_charset is "ISO-2022-*" (which means, for example, that POST[_charset] is set to be that value), and

  - the message body contains ESC (0x1B).

These changes do NOT disturb users at all as long as they do not set the configuration variable "default_head_encoding" and they do not use ESC (0x1B) in the message body.

### DIFF
--- a/program/steps/mail/sendmail.inc
+++ b/program/steps/mail/sendmail.inc
@@ -490,6 +490,10 @@ if (preg_match('/[^\x00-\x7F]/', $MAIL_MIME->getTXTBody())) {
     $text_charset      = $message_charset;
     $transfer_encoding = $RCMAIL->config->get('force_7bit') ? 'quoted-printable' : '8bit';
 }
+else if (strncmp(strtoupper($message_charset), 'ISO-2022', 8) == 0 && preg_match('/\x1B/', $MAIL_MIME->getTXTBody())) {
+    $text_charset      = $message_charset;
+    $transfer_encoding = '7bit';
+}
 else {
     $text_charset      = 'US-ASCII';
     $transfer_encoding = '7bit';
--- a/program/steps/mail/sendmail.inc
+++ b/program/steps/mail/sendmail.inc
@@ -516,10 +516,13 @@ if ($pgp_mime) {
     $MAIL_MIME->setParam('preamble', 'This is an OpenPGP/MIME encrypted message (RFC 2440 and 3156)');
 }
 
+// default MIME head encoding
+$default_head_encoding = $RCMAIL->config->get('default_head_encoding', 'quoted-printable');
+
 // encoding settings for mail composing
 $MAIL_MIME->setParam('text_encoding', $transfer_encoding);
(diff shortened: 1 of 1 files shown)


alecpl:
I don't like the idea of new configuration option for this purpose. Wouldn't be better to set the value automatically depending on the encoding used?

BTW, what is the reason to not use UTF-8?

alecpl:
We need $text_charset to use format=flowed. By automatic I meant to set head_encoding according to $message_charset not a config option. Similar to what you did with text_charset.

dfukagaw28:
> We need $text_charset to use format=flowed.

Okay.
If it is the only purpose, sendmail.inc should not assume that charset=US-ASCII just because it does not contain ``[^\x00-\x7F]``.

> By automatic I meant to set head_encoding according to $message_charset not a config option. Similar to what you did with text_charset.

I think that it is difficult to find a concrete solution to automatically determine ``head_encoding``.
Someone may use ``quoted-printable`` for UTF-8 message, and other may use ``base64``.
What I can say is that it is bad idea to use ``quoted-printable`` for multi byte message, while it is acceptable.

Although I don't understand why you don't like to use new configuration option, it is possible to use a query parameter instead to set ``head_encoding``.

```
$head_encoding = isset($_POST['_transfer_encoding']) ? $_POST['_transfer_encoding'] : 'quoted-printable';
```

What I need is a way to change Content-Type charset and Content-Transfer-Encoding, which I believe is worth to share with those users who compose multibyte messages with RCM.
I don't mind either configuration options or query parameters.

alecpl:
I think you should be able to use `message_ready` hook to modify `head_charset` param.

dfukagaw28:
Thank you for your advice and sorry for my late reply.

I am closing this PR because I recently met a problem which led me to read the code again to find that:

* To change `$message_charset` (e.g., by setting `POST[_charset]`) before calling `rcmail_email_input_format()` may be harmful, as it causes unexpected results in parsing structured headers like To, Cc or Bcc.

I think using `message_ready` hook is reasonable, as you suggested.
Actually I misunderstood the behavior of `$MAIL_MIME->headers($headers)` and noticed that it is possible to change `head_charset` param by _the second_ call.

By the way, is it possible to (conditionally) abort in or just after the `message_ready` handler?
Do we need another `abort` check after the hook?

alecpl:
We could handle abort, but what exactly would you expect there? You can always put this into your code to stop processing and return an error:
```php
    $OUTPUT->show_message('sendingfailed', 'error');
    $OUTPUT->send('iframe');
```

dfukagaw28:
Thank you for your quick information.
I put `global $OUTPUT;` in my plugin then it worked 😝

PR #5650, opened 2017-02-17
Fix to suppress use of problematic characters

ohno-designet:
I found a problem with IMAP folder character check.

When roundcube creates IMAP folder, it suppresses the use of the following characters.
delimiter, '%'(percent), '*'(asterisk)

Actually, only delimiter characters are forbidden.


alecpl:
Yes, but other are problematic. I'm not sure we really want that.

ohno-designet:
OK.
I close this request.

PR #5654, opened 2017-02-17
extend plugin password to avoid sudo (use ssh instead calling chpasswd)

gnadelwartz:
As I already told on the dev mailing list:

I used the password plugin with chasswd and was not pleased to allow the webserver to call "sudo chpasswd".
After some investigation and testing I ended up with a new helper script to  change password via ssh using the provided and excelent expect-passwd method.

Additionally I rewrote the chpasswd driver to provide the old password in a compatible way and extended it to pass error messages back to roundcube.

### DIFF
--- a/plugins/password/README
+++ b/plugins/password/README
@@ -262,6 +262,9 @@
 
  Attached wrapper script (helpers/chpass-wrapper.py) restricts password changes
  to uids >= 1000 and can deny requests based on a blacklist.
+ 
+ For a more secure setup see the notes in section sudo or use the new wrapper
+ chpass-wrapper-expect.py, for usage see there. 
 
 
  2.12.  LDAP - no PEAR (ldap_simple)
@@ -317,7 +320,9 @@
 
  Driver to change user password via the 'expect' command.
  See config.inc.php.dist file for configuration description.
-
+ 
+ For easyer and more secure of the expect script a new wrapper is
+ is provided: chpass-wrapper-expect.py, see there and sudo for setup
 
  2.18.  Samba (smb)
  -----------------------------------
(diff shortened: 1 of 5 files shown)


alecpl:
I don't like the policy checks. I'd propose to remove them. Policy checks are implemented in the plugin core. Only relevant check that should be used nowadays is password length, imho.

gnadelwartz:
Hi,

OK, I will remove them ...

Kay

Am 10.04.2017 um 10:42 schrieb Aleksander Machniak:

alecpl:
Sorry, if I wasn't clear. The password plugin already has password length option, so please remove these checks completely from wrappers.

gnadelwartz:
ah, you are rigth, the original script has ony a check for userid <1000.
I should have look first instead of asking my own memory :-)

alecpl:
Hmm... Just after I merged the PR I noticed that we already have a password driver that uses expect. So, I'm tend to revert this merge. Why would I not?

gnadelwartz:
Give me some time to have a look at the expect implementation you mentoined.
I was assuming there is no implementation, because it is not supported 
by the chpasswd driver, sorry.

My rewritten chpasswd driver support both methods, expect and chpasswd.
additonally my aim was also to avoid direct sudo to expect and chpasswd, 
to incease security.

alecpl:
I reverted the merge, also because we're planning a release candidate soon. So, if you review the expect driver and you see a place for improvement open a new PR, please.

gnadelwartz:
After sleeping over the issue, I wanted to propose the same :-)

Mainly beacause two drivers using the same method is confusing from end 
user perspective and release is so close.

My proposal to proceed is to split my changes in two parts:

1. new chpasswd driver and instructions to use sudo more secure
    driver can stay with expect compatible way to provide old password,
    but I don't promote it and no inclusion of the expect wrapper

2. optionally merge drivers ... but here I have to investigate

PR #5657, opened 2017-02-20
Fixes sieve parsing bug when comments contain empty lines

hansmaulwurf23:
Sample sieve filter set to reproduce the error:

```sieve
require ["fileinto","envelope","regex","body","relational","subaddress","variables","imap4flags","date","include"];

# simple comment

# another comment
if envelope :detail "to" "addressextension" {
    setflag "\\Flagged";
}

# Spam
if header :is "X-Spam-Flag" "YES" {
 	fileinto "Junk";
  	stop;
}
```

Wihtout removing the empty line between `# simple comment` and `# another comment` the parser produces rubbish.


alecpl:
Thanks for the patch. I fixed this in a cleaner way.

Issue #5640, opened 2017-02-12
Can not set Content-Type charset to 7-bit encoding (e.g., ISO-2022-JP)


dfukagaw28:
I can not send a mail with "Content-Type: text/plain; charset=ISO-2022-JP".

Some non-ASCII character encodings use only 7-bit bytes [0x00-0x7F].
Currently, a message encoded by such a 7-bit encoding is unexpectedly sent with "charset=US-ASCII", because of the following judge.

https://github.com/roundcube/roundcubemail/blob/3e427b4a429754906f73a31a4bf29be1c0bf8b40/program/steps/mail/sendmail.inc#L489

Pull request #5639 may solve this.

alecpl:
What if we just replace the regexp with `/[^\x09\x0A\x0D\x20-\x7E]/`?

dfukagaw28:
In that case, the Content-Transfer-Encoding of an ISO-2022-JP message will be either "8-bit" or "quoted-printable", which may cause problems in some situations.

I believe that Content-Transfer-Encoding for an ISO-2022-JP message should be "7-bit", as most people do so.

To encode an ISO-2022-JP message by using "quoted-printable" is nonsence, because ISO-2022-JP does not use the 8th bit in any octet.
"8-bit" seems acceptable as for me, but I heard that some systems re-encode every message sent with "Content-Transfer-Encoding: 8-bit" by using "quoted-printable" (to transfer them in 7-bit).

alecpl:
Fixed.

dfukagaw28:
Thanks!

But, the changes may have a problem as it has been pointed out in a comment to the commit 6f87a32, 
https://github.com/roundcube/roundcubemail/commit/6f87a320528f21f5804620dfa74e8ce7c2dafb63#commitcomment-22816595

PR #5665, opened 2017-02-25
It's better to remove '//IGNORE' from $iconv_options and let mbstring to convert non-strict characters

ernix:
We sometimes get broken character encodings such as:
Subject: =?iso-2022-jp?B?GyRCLWo7M3l1OSk2SBsoQgo=?=

This actually is not a strict ISO-2022-JP string, but a CP50220 string that
is a variant of ISO-2022-JP with extended characters proposed by Microsoft.

https://www.iana.org/assignments/charset-reg/CP50220

iconv can not handle these encodings well.

Example:

    $ echo '㈱山﨑工業' \
        | nkf --ic=UTF-8 --oc=ISO-2022-JP \
        | iconv -f ISO-2022-JP -t UTF-8//IGNORE

    $B-j;3yu9)6H

Remove '//IGNORE' from $iconv_options to produce:

    $ echo '㈱山﨑工業' \
        | nkf --ic=UTF-8 --oc=ISO-2022-JP \
        | php -r '
            $stdin = fgets(STDIN);
            mb_substitute_character('none');
            echo mb_convert_encoding($stdin, "UTF-8", "ISO-2022-JP");'

    山工業

Add alias from ISO-2022-JP to ISO-2022-JP-MS to produce:

    $ echo '㈱山﨑工業' \
        | nkf --ic=UTF-8 --oc=ISO-2022-JP \
        | php -r '
            $stdin = fgets(STDIN);
            mb_substitute_character('none');
            echo mb_convert_encoding($stdin, "UTF-8", "ISO-2022-JP-MS");'

    ㈱山﨑工業

### DIFF
--- a/program/lib/Roundcube/rcube_charset.php
+++ b/program/lib/Roundcube/rcube_charset.php
@@ -212,6 +212,7 @@ class rcube_charset
             $aliases = array(
                 'WINDOWS-1257' => 'ISO-8859-13',
                 'US-ASCII'     => 'ASCII',
+                'ISO-2022-JP'  => 'ISO-2022-JP-MS',
             );
 
             $mb_from = $aliases[$from] ?: $from;
--- a/program/lib/Roundcube/rcube_charset.php
+++ b/program/lib/Roundcube/rcube_charset.php
@@ -170,7 +170,7 @@ class rcube_charset
      */
     public static function convert($str, $from, $to = null)
     {
-        static $iconv_options = null;
+        static $iconv_options = '';
         static $mbstring_sc   = null;
 
         $to   = empty($to) ? RCUBE_CHARSET : strtoupper($to);
@@ -186,24 +186,9 @@ class rcube_charset
             return $str;
         }
 
-        if ($iconv_options === null) {
-            if (function_exists('iconv')) {
-                // ignore characters not available in output charset
-                $iconv_options = '//IGNORE';
-                if (iconv('', $iconv_options, '') === false) {
-                    // iconv implementation does not support options
-                    $iconv_options = '';
-                }
-            }
-            else {
-                $iconv_options = false;
-            }
-        }
-
         // convert charset using iconv module
-        if ($iconv_options !== false && $from != 'UTF7-IMAP' && $to != 'UTF7-IMAP') {
-            // throw an exception if iconv reports an illegal character in input
-            // it means that input string has been truncated
+        if ($from != 'UTF7-IMAP' && $to != 'UTF7-IMAP') {
+            // throw an exception if iconv fails
             set_error_handler(array('rcube_charset', 'error_handler'), E_NOTICE);
             try {
                 $out = iconv($from, $to . $iconv_options, $str);
@@ -232,7 +217,7 @@ class rcube_charset
             $mb_from = $aliases[$from] ?: $from;
             $mb_to   = $aliases[$to] ?: $to;
 
-            // Do the same as //IGNORE with iconv
+            // Ignore characters not available in output charset
             mb_substitute_character('none');
 
             // throw an exception if mbstring reports an illegal character in input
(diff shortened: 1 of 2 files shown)


alecpl:
Since we require `mbstring` I suppose we could completely get rid of `iconv` use in `rcube_charset::convert()`.

ernix:
I'm not sure if we should remove `iconv` from `rcube_charset::convert()` or not.
Because 99% of incoming mails are pretty sane and `iconv` is much faster than mbstring in those cases.  Following is my poor-mans benchmark test between `iconv` vs `mbstring`:

    <?php

    $INPUT = 'こんにちわーるど';

    function bench($callback) {
        $before = microtime(true);
        for ($i = 0; $i < 100000; $i++) {
            $callback();
        }
        $after = microtime(true);
        return $after - $before;
    }

    $funcs = array(
        'iconv' => function () use ($INPUT) {
            return iconv('UTF-8', 'ISO-2022-JP', $INPUT);
        },
        'mbstring' => function () use ($INPUT) {
            return mb_convert_encoding($INPUT, 'ISO-2022-JP', 'UTF-8');
        },
    );

    foreach ($funcs as $name => $func) {
        printf("%-10s: %f\n", $name, bench($func));
    }

Output:

    iconv     : 0.148123
    mbstring  : 0.456378


Plus, types of encoding insanity highly depends on user's languages/culture.
There is no single unified conversion theory, so I think we should add some API hooks for composing/extracting MIMEs so that plugins can tweak preferences for users.

This might be what https://github.com/roundcube/roundcubemail/issues/5640 needs.
It also helps to keep `rcube_charset::convert()` clean and we don't have to consider bypassing `UTF7-IMAP` anymore.

CRtEurope:
In my case mbstring is pretty much faster using php 7.0 or 7.1. Every system is differently configured.

```
iconv : 2.156017
mbstring : 0.535206
```

ernix:
@CRtEurope Agree.  Do you think `iconv` must be used after `mbstring`?  or should be removed from `rcube_charset::convert()`?

I don't know how `mbstring` extension is commonly used.  I'm not a PHP guy.
But, to be a MUA on PHP, it sounds reasonable to require `mbstring` extension.

CRtEurope:
I just can say that iconv should not be removed for various compatibility reasons.

You can check it out to use mbstring first and then iconv.

For mbsting use just take a look at here: http://fr2.php.net/manual/en/function.mb-internal-encoding.php#53265

ernix:
I added convert test for this P-r.

I also fixed BIG-5 raw string in Framework/Charset.php to avoid
multiple encodings appearing in the file.

alecpl:
If we want to keep `iconv` we should also keep `//IGNORE` flag use. So, the only option (besides `iconv` removal) would be to skip `iconv` use for problematic charsets (as we do for UTF7-IMAP). Note: `mbstring` is required, so we know that if we skip iconv, mbstring will be used.

Please, also note that in my tests on PHP7 iconv is still faster for some charsets, e.g. ISO-8859-2. It's slower for ISO-2022-JP, but the difference is not big.

ernix:
I recreated a pull request in order to meet with your requirement
https://github.com/roundcube/roundcubemail/pull/5668

> If we want to keep `iconv` we should also keep `//IGNORE` flag use.

Yes, I wouldn't break any compatibility.

> So, the only option (besides `iconv` removal) would be to skip
> `iconv` use for problematic charsets (as we do for UTF7-IMAP). Note:
> `mbstring` is required, so we know that if we skip iconv, mbstring
> will be used.

Removing `iconv` is a huge topic, since `mbstring` is an extension.

-- 
Shin Kojima

alecpl:
> Removing `iconv` is a huge topic, since `mbstring` is an extension.
Not really. Roundcube requires `mbstring`, if you use Roundcube without required extensions installed you're asking for trouble. But, anyway, let's keep `iconv` code for some more time. Maybe some day we'll remove it.

PR #5668, opened 2017-02-26
Skip iconv for problematic ISO-2022-JP strings 

ernix:
ref: https://github.com/roundcube/roundcubemail/pull/5665

### DIFF
--- a/program/lib/Roundcube/rcube_charset.php
+++ b/program/lib/Roundcube/rcube_charset.php
@@ -200,30 +200,10 @@ class rcube_charset
             }
         }
 
-        $aliases = array(
-            'WINDOWS-1257' => 'ISO-8859-13',
-            'US-ASCII'     => 'ASCII',
-        );
-
         // convert charset using iconv module
-        do {
-            // Skip using iconv if it is not available
-            if ($iconv_options === false) { break; }
-
-            // `iconv` doesn't know modified UTF-7 (a.k.a., UTF7-IMAP)
-            if ($from == 'UTF7-IMAP') { break; }
-            if ($to   == 'UTF7-IMAP') { break; }
-
-            // CP5022x
-            if (function_exists('mb_check_encoding')) {
-                if ($from === 'ISO-2022-JP'
-                    && !mb_check_encoding($str, 'ISO-2022-JP')
-                ) {
-                    $aliases['ISO-2022-JP'] = 'ISO-2022-JP-MS';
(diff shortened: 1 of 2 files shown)


tests/Framework/Charset.php
```diff
@@ -64,13 +64,27 @@ function test_parse_charset($input, $output)
...
-        return array(
+        $tests = array(
             array('ö', 'ö', 'UTF-8', 'UTF-8'),
             array('ö', '', 'UTF-8', 'US-ASCII'),
             array('aż', 'a', 'UTF-8', 'US-ASCII'),
             array('&BCAEMARBBEEESwQ7BDoEOA-', 'Рассылки', 'UTF7-IMAP', 'UTF-8'),
             array('Рассылки', '&BCAEMARBBEEESwQ7BDoEOA-', 'UTF-8', 'UTF7-IMAP'),
         );
+
+        if (extension_loaded('mbstring')
```
alecpl: As I said in another place, this is required extension. No need to check it here.

program/lib/Roundcube/rcube_charset.php
```diff
@@ -200,8 +200,30 @@ public static function convert($str, $from, $to = null)
...
         }
 
+        $aliases = array(
+            'WINDOWS-1257' => 'ISO-8859-13',
+            'US-ASCII'     => 'ASCII',
+        );
+
         // convert charset using iconv module
-        if ($iconv_options !== false && $from != 'UTF7-IMAP' && $to != 'UTF7-IMAP') {
+        do {
```
alecpl: I think you made this code overcomplicated. I would change only the if statement. I don't like do/while and mb_check_encoding().

ernix:
> I don't like do/while and mb_check_encoding().

OK, I removed do/while and mb_check_encoding.

-- 
Shin Kojima

ernix:
> No need to check it here.

OK, I kept mb_substitute_character as current code from `rcube_charset::convert`

    223         if ($mbstring_sc === null) {
    224             $mbstring_sc = extension_loaded('mbstring') ? mb_substitute_character() : false;
    225         }

-- 
Shin Kojima

ernix:
@alecpl Should I remove `mb_substitute_character` check in the test too?

alecpl:
Yes, it does not make sense.

ernix:
@alecpl done.  Is it acceptable for you?

PR #5674, opened 2017-03-02
Unset HTTP Referer when following external links

da2x:
Keep user’s webmail provider private by unsetting the `Referer`
header when clicking through to or requesting any external resources.

https://www.w3.org/TR/referrer-policy/#referrer-policy-header
https://www.w3.org/TR/referrer-policy/#referrer-policy-same-origin

### DIFF
--- a/program/lib/Roundcube/rcube_output.php
+++ b/program/lib/Roundcube/rcube_output.php
@@ -190,6 +190,9 @@ abstract class rcube_output
 
         // Request browser to disable DNS prefetching (CVE-2010-0464)
         header("X-DNS-Prefetch-Control: off");
+     
+        // Request browser disable Referer (sic) header
+        header("Referrer-Policy: same-origin");
 
         // send CSRF and clickjacking protection headers
         if ($xframe = $this->app->config->get('x_frame_options', 'sameorigin')) {


thomascube:
Duplicate of #5422?

da2x:
@thomascube, almost. I wasn’t aware of the other issue. However, making this browser instruction  configurable seems like unnecessary complication. I can’t imagine anyone ever wanting to change this header from the default of `same-origin` (as suggested as default in this pull request).

As I see it, the two patches differ in the following ways: the `meta` tag approach in #5422 will work if the document is saved to disk and opened in a browser later. The HTTP header approach (this pull request) will just work for everyone and stay out of everyone’s hair.

program/lib/Roundcube/rcube_output.php
```diff
@@ -190,6 +190,9 @@ public function common_headers()
 
         // Request browser to disable DNS prefetching (CVE-2010-0464)
         header("X-DNS-Prefetch-Control: off");
+     
+        // Request browser disable Referer (sic) header
+        header("Referrer-Policy: same-origin");
```
Rotzbua: string should use single quote

Rotzbua:
Why is this pr not merged?

alecpl:
Implemented in 186f21c4c.

PR #5693, opened 2017-03-16
Avoid string truncation

Dreeg:
Avoid truncation of the string like screen: http://oi67.tinypic.com/2ius5d5.jpg
The sense is still understandable.


alecpl:
This need to be done via the Transifex service.

PR #5702, opened 2017-03-21
Add support for DelSp=Yes messages

dfukagaw28:
Currently RCM can not handle `delsp=yes` messages correctly.

When receiving a message with `format=flowed;delsp=yes`, RCM unfold the message body incorrectly.
For example, `Round \r\nCube` is unfolded into `Round Cube` although `RoundCube` is expected.

Some MUAs, including Thunderbird (>45.0), generate MIME messages with "format=flowed;delsp=yes" by default if the charset is set to be ISO-2022-JP.

MIME option `delsp=yes` is specified in RFC 3676
https://www.ietf.org/rfc/rfc3676.txt

>  If the line is flowed and DelSp is "yes", the trailing space immediately prior to the line's CRLF is logically deleted.  If the DelSp parameter is "no" (or not specified, or set to an unrecognized value), the trailing space is not deleted.

> The DelSp parameter was added specifically to permit Format=Flowed to be used with languages/coded character sets in which the ASCII space character is rarely used, or not used at all.

### DIFF
--- a/program/lib/Roundcube/rcube_mime.php
+++ b/program/lib/Roundcube/rcube_mime.php
@@ -437,12 +437,13 @@ class rcube_mime
     /**
      * Interpret a format=flowed message body according to RFC 2646
      *
-     * @param string $text Raw body formatted as flowed text
-     * @param string $mark Mark each flowed line with specified character
+     * @param string  $text  Raw body formatted as flowed text
+     * @param string  $mark  Mark each flowed line with specified character
+     * @param boolean $delsp Remove the trailing space of each flowed line
      *
      * @return string Interpreted text with unwrapped lines and stuffed space removed
      */
-    public static function unfold_flowed($text, $mark = null)
+    public static function unfold_flowed($text, $mark = null, $delsp = false)
     {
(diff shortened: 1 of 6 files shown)


alecpl:
This looks to me as incomplete. Did you forget about other changes or not implemented them yet? I see you added an argument to unfolding function, but I don't see changes in the code that uses it.

dfukagaw28:
Yes, you are right.
I will put another commit which uses the unfolding function.
Thank you for your comment.

dfukagaw28:
I hope it works this time.

Note that this PR is just for unfolding (receiving) "DelSp=Yes" messages, not for generating (sending) them.

alecpl:
Looks good, but there are more places where delsp value need to be passed as an argument:
https://github.com/roundcube/roundcubemail/blob/master/program/steps/mail/compose.inc#L913
https://github.com/roundcube/roundcubemail/blob/master/program/steps/mail/compose.inc#L930

dfukagaw28:
Thanks again.
I realized that the fix is necessary to compose a appropriate reply to a message with DelSp=Yes.

PR #5727, opened 2017-04-18
password plugin: better error message on constraint violation

torbjo:
If the password policy on the ldap server is violated when changing
password, report error as "constraint violation" instead of the
misleading "connection error".

Note: LDAP_CONSTRAINT_VIOLATION can be returned for multiple
reasons – not just password policy violation – but in this case
it's a pretty safe bet it's caused by a ppolicy violation.

Only tested with OpenLDAP (v2.4).


plugins/password/password.php
```diff
@@ -25,6 +25,7 @@
 define('PASSWORD_ERROR', 2);
 define('PASSWORD_CONNECT_ERROR', 3);
 define('PASSWORD_IN_HISTORY', 4);
+define('PASSWORD_LDAP_CONSTRAINT_VIOLATION', 5);
```
alecpl: I think we could skip _LDAP in the constant name. Other drivers might also want to use it.

alecpl (changes requested):

plugins/password/password.php
torbjo: Sure. Have pushed fixed.

I was unsure about the naming of this one, so I defaulted to keeping it similar to the constant from ldap.h

alecpl (changes requested):
You have to rename the constant in all places

torbjo:
Whops. My bad. Fixed.

lbguilherme:
Typo: Password constraint violation. Password probably ~to~ **too** weak.

torbjo:
Thanks.
Can I just push a fix to this branch/PR or should I create a new one?

alecpl:
A new one please. The typo is already fixed.

PR #5742, opened 2017-04-29
SRI + XHTML

corrideat:
I've written a patch that does the following:

1) Impment support for subresource requrest integrity (off by default)
2) Fixes some of the regexes used for matching tag attributes (the current expression would match `src="a'`; now, it enforces that the same quotation style be used)
3) Fixed the missing "/" at the end of tags that aren't closed, useful for XHTML compliance
4) Added the xhtml-html5 doctype to support xhtml5

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -770,7 +770,7 @@ EOF;
         }
 
         list($file) = explode('?', $path, 2);
-        if (!empty($file) && is_readable($file)) {
+        if (!empty($file) && is_readable($this->assets_dir . $file)) {
 	    $content = @file_get_contents($this->assets_dir . $file);
 	    if ($content !== false) {
 	        return 'sha384-'.base64_encode(hash('sha384', $content, true));
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -85,7 +85,7 @@ class rcmail_output_html extends rcmail_output
         $this->set_env('skin', $skin);
 
         $this->set_assets_path($this->config->get('assets_path'), $this->config->get('assets_dir'));
-        $this->assets_sri = $this->config->get('assets_sri');
+        $this->assets_sri = $this->config->get('assets_sri') && in_array('sha384', hash_algos(), true);
 
         if (!empty($_REQUEST['_extwin']))
             $this->set_env('extwin', 1);
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -46,6 +46,7 @@ class rcmail_output_html extends rcmail_output
     protected $body   = '';
     protected $base_path = '';
     protected $assets_path;
+    protected $assets_sri = false;
     protected $assets_dir = RCUBE_INSTALL_PATH;
     protected $devel_mode = false;
 
@@ -84,6 +85,7 @@ class rcmail_output_html extends rcmail_output
         $this->set_env('skin', $skin);
 
         $this->set_assets_path($this->config->get('assets_path'), $this->config->get('assets_dir'));
+        $this->assets_sri = $this->config->get('assets_sri');
 
         if (!empty($_REQUEST['_extwin']))
             $this->set_env('extwin', 1);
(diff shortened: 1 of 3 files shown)


program/include/rcmail_output_html.php
```diff
@@ -756,6 +758,29 @@ public function raise_error($code, $message)
...
+        // iframe content can't be in a different domain
+        // @TODO: check if assests are on a different domain
+
+        if (!$this->assets_sri || in_array($path[0], array('?', '/', '.')) || strpos($path, '://')) {
+            return false;
+        }
+
+        list($file) = explode('?', $path, 2);
+        if (!empty($file) && is_readable($file)) {
+	    $content = @file_get_contents($this->assets_dir . $file);
```
alecpl: Here you use assets_dir, but one line above you don't.

program/include/rcmail_output_html.php
```diff
@@ -756,6 +758,29 @@ public function raise_error($code, $message)
...
+	    if ($content !== false) {
+	        return 'sha384-'.base64_encode(hash('sha384', $content, true));
```
alecpl: Would be good to check if 'hash' function exists, e.g. when you set $this->assets_sri.

program/include/rcmail_output_html.php
```diff
@@ -826,7 +851,34 @@ protected function file_callback($matches)
...
+        return $matches[1] . '=' . $matches[2] . $file . $matches[2] . $matches[4];
+    }
+
+    /**
+     * Add SRI tags to assets
+     */
+    protected function include_assets_sri($output)
+    {
+        return preg_replace_callback(
+            '!(src|href|background)=(["\']?)([a-z0-9/_.?=-]+)\2([\s>])?!i',
```
alecpl: W3C specification talks only about link and script tags. Do we really want to handle all others, e.g. img?

alecpl:
Would be also add a note to defaults.inc.php about the new option and performance implications.

program/include/rcmail_output_html.php
corrideat: Fair point. I missed that.

program/include/rcmail_output_html.php
corrideat: Fair point.

program/include/rcmail_output_html.php
corrideat: I went this route because it was easier to implement for _everything_ and because future browsers may support SRI in other tags. I concur that for performance reasons this might not be the best approach.

corrideat:
Can't seem to be able to reply to your last comment, but I added a commit that documents the option in defaults.inc.php.

program/include/rcmail_output_html.php
corrideat: A slightly off-topic comment: the current way of implementing these features using templates (i.e., generate a document, then use regexes) seems error-prone and inefficient. I think a DOM-based (or similar) document generation approach could be more reliable and flexible (even though it's hard to say what the effect on performance may be.) Have you considered changing this?

program/include/rcmail_output_html.php
```diff
@@ -756,6 +758,29 @@ public function raise_error($code, $message)
...
+
+        if (!$this->assets_sri || in_array($path[0], array('?', '/', '.')) || strpos($path, '://')) {
+            return false;
+        }
+
+        list($file) = explode('?', $path, 2);
+        if (!empty($file) && is_readable($this->assets_dir . $file)) {
+	    $content = @file_get_contents($this->assets_dir . $file);
+	    if ($content !== false) {
+	        return 'sha384-'.base64_encode(hash('sha384', $content, true));
```
alecpl: I propose to use hash_file() instead of hash() and file_get_contents().

program/include/rcmail_output_html.php
```diff
@@ -84,6 +85,7 @@ public function __construct($task = null, $framed = false)
         $this->set_env('skin', $skin);
 
         $this->set_assets_path($this->config->get('assets_path'), $this->config->get('assets_dir'));
+        $this->assets_sri = $this->config->get('assets_sri') && in_array('sha384', hash_algos(), true);
```
alecpl: I don't think we need to check hash_algos(), but I would check if function_exists('hash_file').

dereks:

```
git fetch upstream
git rebase upstream/master
```

Neustradamus:
@corrideat: Have you progressed on your PR?

corrideat:
> @corrideat: Have you progressed on your PR?

@Neustradamus This is a pretty old PR that I didn't update. At the moment, I'm no longer running RC but I'll see whether I can update it or close it

PR #5767, opened 2017-05-19
newmail_notifier plugin: don't notify for junked messages.

rigrig:
I don't think users need to be notified of new mail in their junk folder.

When the mail server filters spam directly into the junk folder, but leaves it marked unread, so you can check for false positives every now and then.


rigrig:
Thanks!

PR #5773, opened 2017-05-28
Some corrections

yesszus:


alecpl:
Please, use https://www.transifex.com/roundcube/roundcube-webmail/ for localization updates.

PR #5777, opened 2017-05-30
Update zipdownload.php

lucasmarin:
Adding a counter before .(extension) to files with same name on zip.
Before only the last file on message with equals names was added on zip


plugins/zipdownload/zipdownload.php
```diff
@@ -141,18 +143,8 @@ public function download_attachments()
...
-
-            if ($filename === null || $filename === '') {
-                $ext      = (array) rcube_mime::get_mime_extensions($part->mimetype);
-                $ext      = array_shift($ext);
-                $filename = $rcmail->gettext('messagepart') . ' ' . $pid;
-                if ($ext) {
-                    $filename .= '.' . $ext;
-                }
-            }
+            $disp_name = $this->_create_displayname($part->filename);
```
alecpl: _create_displayname() uses different arguments. Did you test this code?

plugins/zipdownload/zipdownload.php
```diff
@@ -190,6 +182,49 @@ public function download_messages()
...
+    
+    /**
+     * Create and get display name of attachment part to add on zip file
+     *
+     * @param $part stdClass Part of attachment on message
+     * @param $pid string Part Id of attachment
+     *
+     * @return string Display name of attachment part
+     */
+    private function _create_displayname($part, $pid)
```
alecpl: $pid is just $part->mime_id, so we don't need an additional argument.

alecpl (changes requested):

plugins/zipdownload/zipdownload.php
lucasmarin: Ouch. Sorry for mistake, I tested on different version of roundcube (1.1.9), and it confused me to make a commit to a master version... 

I will make a new pull request with correct code and tested with master version

plugins/zipdownload/zipdownload.php
lucasmarin: The $part variable is updated on

`        foreach ($message->attachments as $part) {
            $pid      = $part->mime_id;
            $part     = $message->mime_parts[$pid];
`
And on the actual code the variable is mentioned directly 

`-            $filename = $part->filename;
 -
 -            if ($filename === null || $filename === '') {
 -                $ext      = (array) rcube_mime::get_mime_extensions($part->mimetype);
 -                $ext      = array_shift($ext);
 -                $filename = $rcmail->gettext('messagepart') . ' ' . $pid;
 -                if ($ext) {
 -                    $filename .= '.' . $ext;
 -                }
 -            }
`
I didn't know that it was intentional to get the right value, or, with some circunstances 
on $part variable updated the value of mime_id wans't set, so I put it with argument on method.

I will update the pull request with this change.

Thanks alecpl.

Regards.

lucasmarin:
Correcting script to merge with master.

Thx @alecpl.

Issue #5781, opened 2017-06-02
list object, touch events and IE11/Edge


JohnDoh:
When using IE11 or Edge if you touch and hold on a message in the message list then the drag widget appears for a moment or two but then the message you touched on loads in the preview pane and the drag event is canceled.

It seems that in IE11 or Edge any time you touch a message in the message list it is loaded into the preview pane regardless of how long you touched it for or anything. Even if you are just scrolling up and down the list. In Firefox it does not load the message into the preview pane so quickly you need to touch it and let go, if you are just scolling it ignores the touch.

Similar behaviour can be seen with the Contextmenu plugin. In Firefox if you touch and hold on a message then the menu is displayed and the preview pane remains blank but in IE11/Edge the menu is displayed and the message loaded into the preview pane.

JohnDoh:
I think I've narrowed it down to the mousedown event handler on line 156 in list.js. mousedown is triggered at the start of the touch, there is no difference if its a short touch or a long touch or a drag.

pointerdown and pointerup events might help - the event contains a property that says if its a mouse or touch event but there is still the problem of determining the purpose of the touch (long, short, drag etc)

alecpl:
Merged.

PR #5783, opened 2017-06-03
Add support for Pointer events

johndoh:
Related to #5781 

Add support in list.js for pointer events. Also add detection for Edge browser.

In IE11/Edge the mousedown event is fired too easily for touch events, This uses the pointerdown and pointerup events to check the length of time the touch was held and act accordingly


alecpl (changes requested):
Resolve conflicts, please.

PR #5786, opened 2017-06-03
Adding hostNET Password Driver

jtkoerting:
New driver file to support hostNET Managed-Root Servers.

### DIFF
--- a/plugins/password/drivers/dovecot_passwdfile.php
+++ b/plugins/password/drivers/dovecot_passwdfile.php
@@ -1,16 +1,31 @@
 <?php
 
 /**
- * hostNET Password Driver
+ * Dovecot passwdfile Password Driver
  *
- * Driver that adds functionality to change the mailuser password on hostNET Servers.
+ * Driver that adds functionality to change the passwords in dovecot 2 passwd-file files.
  * The code is derrived from the Plugin examples by The Roundcube Dev Team
  *
- * It only works on hostnet.de Managed-Root Server Systems.
+ * Intentionally written for hostnet.de Managed-Root Server Systems, it determines
+ * the hostBSD System to preset any needed settings. hostNET Managed-Root Customers only need to set
+ *
+ * $config['password_driver'] = 'dovecot_passwdfile'
+ *
+ * in roundcubes plugins/password/config.inc.php to have all set.
+ * (But don't forget to enable the 'password' plugin at all...)
+ *
+ * On vanilla dovecot 2 environments, use the correct values for these config settings:
+ *
+ * $config['password_dovecot_passwdfile_path']: The path of your dovecot passwd-file '/path/to/filename' as set in dovecot/conf.d/auth-passwdfile.conf.ext
+ * $config['password_dovecotpw']: Full path and 'pw' command of doveadm binary - like '/usr/local/bin/doveadm pw'
+ * $config['password_dovecotpw_method']: Dovecot hashing algo (http://wiki2.dovecot.org/Authentication/PasswordSchemes)
+ * $config['password_dovecotpw_with_method']: True if you want the hashing algo as prefix in your passwd-file
  *
  * @version 1.0; Jimmy Koerting
  *
- * Copyright (C) 2017, hostNET Medien GmbH
+ * Copyright (C) 2017, hostNET Medien GmbH, www.hostnet.de
+ *
+ *
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


alecpl:
What software is hostNET using? Is this available for other hosting providers or hostNET only? I wouldn't want to merge provider-specific code.

jtkoerting:
I guess every server using the dovecot 2 mailuser file format can use this. This is why I added the password_hostnet_mailuserfile option, but I can't guarantee that it's working for everyone.

We have roundcube preinstalled for our customers since years and they are using in on >500.000 Mailboxes here, so the patch would be great for many people as they do there roundcube updates on their own on our managed-root systems.

As you have other provider-specific code, I thought that is no general problem. We are happy to do some donation if this is of help, as this would help us and our customers.

alecpl:
We have indeed one (domainFactory), I don't know how that happened. We'll consider this. You can make a donation anyway ;).

jtkoerting:
I did. And many thanks for considering that. I'm sure it will help many roundcube users!

thomascube:
Donation thankfully received :-)

If this is generic for dovecot it should probably be named something like "dovecot_mailuser" or at least be described in the README file.

alecpl:
I think the plugin name should be dovecot_passwdfile (to match with dovecot's naming, https://wiki.dovecot.org/BasicConfiguration). Also, instead of using doveadm directly you could use `password::hash_password($password)` method. So, it could be configured with password scheme. Then configuration would look like this:
```
$config['password_driver'] = 'dovecot_passwdfile';
$config['password_algorithm'] = 'sha512-crypt';
$config['password_algorithm_prefix'] = '{SHA512-CRYPT}'; // optional, I suppose
// or if you prefer dovecot way:
$config['password_algorithm'] = 'dovecot';
$config['password_dovecotpw'] = '/usr/local/sbin/doveadm pw';
$config['password_dovecotpw_method'] = 'SHA512-CRYPT';
$config['password_dovecotpw_with_method'] = false; // or true
```

jtkoerting:
I rewrote and renamed the plugin, respecting your concerns.

_One remark might be allowed for me: If you have even ONE provider specific module - thus supporting non opensource - you should allow this to everyone else. Otherwise, you should remove all provider specific code. An excuse like 'This module was always here...' is not fair. :)_

By now, I have the code that way, that it could be used in any vanilla dovecot 2 environment using the dovecot passwd-file format, files and methods given in dovecot/conf.d/auth-passwdfile.conf.ext

On the other hand - and please keep at least this for us - I put in a little piece of code which is checking, if this roundcube instance is running on our own BSD derivate (hostBSD) to set some other defaults.

That way it can be used nearly out of the box for our customers and tweaked for everyone else using the dovecot 2 passwd-file driver for user and passwort authentication.

One remark: the `password::hash_password($password)` (crypt() function) is not working/is not compatible with (at least) the SHA512-CRYPT used from doveadmin pw - so, it shouldn't be used here.

Only these (existing) plugins/password/config.inc.php options are needed:

**$config['password_driver']** = '_dovecot_passwdfile_'
**$config['password_dovecot_passwdfile_path']**: _The path of your dovecot passwd-file '/path/to/filename' as set in dovecot/conf.d/auth-passwdfile.conf.ext_
**$config['password_dovecotpw']**: _Full path and 'pw' command of doveadm binary - like '/usr/local/bin/doveadm pw'_
**$config['password_dovecotpw_method']**: _Dovecot hashing algo (http://wiki2.dovecot.org/Authentication/PasswordSchemes)_
**$config['password_dovecotpw_with_method']**: T_rue if you want the hashing algo as prefix in your passwd-file_


Please let me know if this is fine for you this way.

jtkoerting:
ouch - my mistake. This one is new:

**$config['password_dovecot_passwdfile_path']**: _The path of your dovecot passwd-file '/path/to/filename' as set in dovecot/conf.d/auth-passwdfile.conf.ext_

thomascube:
Fine, thanks a lot!

Now if you could squash your 3 commits into one, it'd be perfect for merging into master.

thomascube:
One thing to note though: with `file_get_contents()` > (alter content) > `file_put_contents()` there's the chance that another process has changed the file in the meantime and those changes are lost due to a complete overwrite. Consider to lock the file while updating or use something like `sed -i` to replace the one entry inline.

alecpl:
Also, are you sure about escapeshellcmd() use?

jtkoerting:
sorry guys - completely filled up with other jobs at my company atm. I will come back asap.

alecpl (changes requested):
Waiting for the fixes to the last couple of comments.

alecpl:
dovecot_passwdfile already exists.

PR #5820, opened 2017-06-27
drop ereg dependency, fix compat with PHP 7+

remicollet:


remicollet:
Notice: phpcompatinfo is your friend ;)
https://github.com/llaville/php-compat-info

PR #5821, opened 2017-06-27
Copyright 2016 -> 2017

shoujii:
Changed the Copyright in many comments and in the visible about page to 2017...

### DIFF
--- a/plugins/password/drivers/cpanel.php
+++ b/plugins/password/drivers/cpanel.php
@@ -16,7 +16,7 @@
  * @version 3.1
  * @author Christian Chech <christian@chech.fr>
  *
- * Copyright (C) 2005-2016, The Roundcube Dev Team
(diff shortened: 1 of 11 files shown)


alecpl:
Copyright date/year is not a simple subject, but we have a custom of updating the year only with significant changes in the file. So, I'd not apply these. Only change in about.html make sense. @thomascube what do you think?

BoBBer446:
Maybe it makes sense to Update the Year because we are in 2017 and this Software has a copyright in 2017. Why dont show it?

marclaporte:
Please don't do commits for this. It pollutes the logs.
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

alecpl:
We removed year from copyright headers.

PR #5838, opened 2017-07-03
fix wrong "text" parsing

skiner68:
accept lines containing dots only - I thing check position index should be $position, not $pos


alecpl:
Looks like you might be right, but do you have a sample sieve script that fails to parse correctly? I'd like to add a test for this.

alecpl:
Nevermind, I fixed this in  8b61d6a2d1.

skiner68:
Great, you were fast. Just one thing - should not $position be incremented once more in case line ends with "\r\n"? Honestly I have not checked this practically, but length of "\r\n" is 2 symbols, not 1 symbol like in case of simple "\n", so the $position value may be wrong in this case. Please see my original pull request.

alecpl:
You're right. I couldn't find a sample that breaks without this, but I fixed it anyway.

PR #5839, opened 2017-07-03
Disable qrcode if Endroid QrCode library is missing

Philippe23:
Treat a missing QrCode library the same as roundcube treats a missing
php-gd library.  Hide the button to disallow access.

Requested by distributions that don't use composer and don't yet have
packages for the QrCode library.


alecpl:
Thanks. Merged.

PR #5842, opened 2017-07-05
Driver for ispconfig (change password)

xScorp1oNx:


alecpl:
Is there any reason you can't use sql driver? I'd rather add info how to configure it for ISPConfig in documentation that create a new driver.

Issue #5846, opened 2017-07-10
php settings in .htaccess ignored by php-fpm


tessus:
The php settings in `.htaccess` are ignored by php-fpm.

```
php_flag    display_errors  Off
php_flag    log_errors      On
# php_value    error_log    logs/errors

php_value   upload_max_filesize   5M
php_value   post_max_size         6M
php_value   memory_limit          64M

php_flag    zlib.output_compression     Off
php_flag    suhosin.session.encrypt     Off

#php_value   session.cookie_path     /
#php_value   session.hash_function   sha256
php_flag    session.auto_start       Off
php_value   session.gc_maxlifetime   21600
php_value   session.gc_divisor       500
php_value   session.gc_probability   1
```

If you want them to be honored by php-fpm as well, you have to create a file called `.user.ini` and add the following content:

```
display_errors=Off
log_errors=On
;error_log=logs/errors

upload_max_filesize=5M
post_max_size=6M
memory_limit=64M

zlib.output_compression=Off
suhosin.session.encrypt=Off

;session.cookie_path=/
;session.hash_function=sha256
session.auto_start=Off
session.gc_maxlifetime=21600
session.gc_divisor=500
session.gc_probability=1
```

alecpl:
Merged.

PR #5847, opened 2017-07-10
.user.ini for php-fpm

tessus:
PHP settings in .htaccess are ignored by php-fpm. They have to
be put in a file called .user.ini

see #5846

### DIFF
--- a/.user.ini
+++ b/.user.ini
@@ -1,0 +1,21 @@
+; This file is read and processed by the CGI/FastCGI SAPI
+; PHP settings in .htaccess are only available to the PHP Apache module
+; also see: http://php.net/manual/en/configuration.file.per-user.php
+
+display_errors=Off
+log_errors=On
+;error_log=logs/errors
+
+upload_max_filesize=5M
+post_max_size=6M
+memory_limit=64M
+
+zlib.output_compression=Off
+suhosin.session.encrypt=Off
+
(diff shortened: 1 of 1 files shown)


alecpl:
Could you add a comment on top of the file that describes its purpose and maybe with reference to http://php.net/manual/en/configuration.file.per-user.php?

tessus:
Sure, will do. I just haven't done it because there was no comment in `.htaccess` either. I mean a comment related to what the purpose of the file is.
Which brings me to my next point: there's no reason why you shouldn't add the IfModule test for php7 as well.
I'll grab late lunch and make a few changes afterwards (in about 60 minutes). You can always decide to cherry pick the commits. Or I'll create another PR for the .htaccess file.

tessus:
Done. Added a description as requested.

I've also created a separate PR #5850 for the `.htaccess` file. There's no reason to have the user change the file manually, if it can be easily done by adding another `IfModule` test. No more manual interaction necessary.

alecpl:
.htaccess is old and known technology, while .user.ini is less common and the file name is configurable. That's why I think the educative comment is needed. Thanks.

tessus:
Yes, it is true that .htaccess has been around for a long time, but you should be aware that the filename of .htaccess is also configurable.
Anyway, it now has a description. :-)

PR #5850, opened 2017-07-10
add section for PHP7 to .htaccess

tessus:
It makes no sense to make the user change the file, if it can be
added to the file in its own block <IfModule mod_php7.c>

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,5 +1,26 @@
-# WARNING: For PHP 7 the module name in the line below need to be modified!
+# Change this section for PHP 5.x
 <IfModule mod_php5.c>
+php_flag    display_errors  Off
+php_flag    log_errors      On
+# php_value    error_log    logs/errors
+
+php_value   upload_max_filesize   5M
+php_value   post_max_size         6M
+php_value   memory_limit          64M
+
+php_flag    zlib.output_compression     Off
+php_flag    suhosin.session.encrypt     Off
+
+#php_value   session.cookie_path     /
+#php_value   session.hash_function   sha256
+php_flag    session.auto_start       Off
+php_value   session.gc_maxlifetime   21600
+php_value   session.gc_divisor       500
+php_value   session.gc_probability   1
+</IfModule>
+
+# Change this section for PHP 7.x
+<IfModule mod_php7.c>
 php_flag    display_errors  Off
 php_flag    log_errors      On
 # php_value    error_log    logs/errors
@@ -70,7 +91,7 @@ Header set X-Robots-Tag "noindex, nofollow"
 #Header always set Public-Key-Pins "max-age=3600; report-uri=\"\"; pin-sha256=\"\"; pin-sha256=\"\"" env=HTTPS
 
 # X-Xss-Protection
-# This header is used to configure the built in reflective XSS protection found in Internet Explorer, Chrome and Safari (Webkit). 
+# This header is used to configure the built in reflective XSS protection found in Internet Explorer, Chrome and Safari (Webkit).
 #Header set X-XSS-Protection "1; mode=block"
 
 # X-Frame-Options


alecpl:
If the user is about to change the file, what's the problem in changing module name too? I don't like duplicated code.

tessus:
Well, usually I don't like duplicating code either, but the current .htaccess file only works for php5 out of the box. This is not a good design.
Don't you want it to work for all major PHP versions? If the user changes something, they can still delete the stanza not relevant to their environment or just ignore it. But at least it will work for both PHP versions right from the start.

alecpl:
I'm just saying that both solutions can lead to confusion. User can update the config option in wrong section. Imho, your solution is not better, but maybe others will think different. We'll see. Maybe after some time we'll just switch to php7-only solution as a more common.

rmcaninch:
If I'm understanding correctly, testing for php5 or php7 would let it work out of the box with either version available. If someone is going into the file to change configuration then some notes on changing the correct block may help.

The install.sh or upgrade.sh could possibly look for what version of php is installed and rewrite the .htaccess? Probably more work than it is worth.

tessus:
@rmcaninch yes, you understood correctly. You brought up a good point and I'll add comments to the change.

I've noticed that the php settings are not absolutely essential for roundcube to work (since most of them might be set for a production system anyway), but if .htaccess sets them for PHP5, they should also be set for PHP7.

@alecpl I really do understand that duplication is not very nice, but unfortunately the `IfModule` directive does not allow a logical OR nor pattern matching.

rmcaninch:
Possible approach that would work for Apache >= 2.4. It would be less duplication.

<IfModule mod_php5.c>
    Define PHP_IS_ENABLED
</IfModule>

<IfModule mod_php7.c>
    Define PHP_IS_ENABLED
</IfModule>

<IfDefine PHP_IS_ENABLED>
# Current Config
</IfDefine>

tessus:
@rmcaninch I thought about that as well, but as with PHP, a lot of people still use Apache 2.2.
But I'd be ok with that too, if this were more appealing to the developers.

alecpl:
Really, if we agree that comment is needed then one is enough. I see no point in adding redundant sections. I like @rmcaninch proposal, but I'm not sure we should do this. Probably we could move to Apache 2.4 in master. Will that syntax cause errors or will be just ignored by Apache < 2.4?

tessus:
No, they would not be ignored by Apache 2.2. They would raise an error.

However, there's another directive you can use:
```
<IfVersion >= 2.4>
    <IfModule mod_php5.c>
        Define PHP_IS_ENABLED
    </IfModule>

    <IfModule mod_php7.c>
        Define PHP_IS_ENABLED
    </IfModule>**

    <IfDefine PHP_IS_ENABLED>
        # Current Config
    </IfDefine>
</IfVersion>
```

Please note that this means that there won't be any php settings set for Apache 2.2.
So however you turn it, one setup is always left out - unless you duplicate it.

There's also another reason for 2 php sections: some php settings might not be available for newer versions. e.g. when a PHP setting has been removed (like safe_mode was in 5.4.0), the attempt to set it will raise a fatal error.
Or you could want to set or use different PHP settings, depending on the version.
If that's not the case, then you can ignore my reasoning for 2 php sections.

tomsommer:
Please also consider `<IfModule mod_lsapi.c>`

tessus:
@alecpl any update on this?

alecpl:
Marked for consideration for the next major release.

alecpl:
There's one more option. Get rid of the whole php configuration section from .htaccess file. We can force/check some options in the code and leave all the rest to the user. We can describe some of them in documentation. @thomascube what do you think?

thomascube:
I generally agree to extract the php settings from the .htaccess file. Too many combinations of Apache and PHP versions plus there are other webservers and environments that are not covered. Will take some effort to properly transfer this into documentation and to make people aware.

Regarding checks in the code: yes but not in index.php which handles every single HTTP request. Should be part of the installer and update.sh scripts. However, the caveat here is, that php.ini settings can vary between cli and webserver environment and update.sh is executed from the shell. So update problems and support questions are granted. Therefore proper documentation and hints in README, FAQ, etc. is required.

tessus:
In this case `installto.sh` should not create copies of `.htaccess` and `.user.ini`, but rather make the new files  available with a `.new` extension.
This would decrease the downtime during an upgrade.

What do you think? If you agree, I can create a PR.

alecpl:
Makes sense.

tessus:
Ok, I'll work on it tomorrow and send a PR.

Btw, while I'm at it, is it really necessary to copy the `installer` directory in case of an upgrade?
e.g. it's the first thing I delete after the upgrade, and AFAIK the installer dir is only needed for a fresh install.

alecpl:
Good question. It is mentioned in manual upgrade instructions, but when using  the script I think we can ommit that folder. So, please include that change in your PR.

tessus:
I'm almost done. One thing left:

What shall I do, if the target dir has an `installer` directory?

- delete it without notifying the user
- ask, if it should be deleted
- do nothing

tessus:
Please see https://github.com/roundcube/roundcubemail/pull/6261

If you want me to implement either:
- delete it without notifying the user
- ask, if it should be deleted

please let me know.

alecpl:
Now when you asked... I think we should check if the directory exist and in such case update it. As we did before, but conditionally. Of course we could notify the user that "The installer directory exist, you should remove it after installation/upgrade".

tessus:
Hmm, why would you update a directory that shouldn't be there in the first place?

But sure, I can do that.

I'll rsync the dir, if it exists on the target. At the end of the script I'll notify the user that it should be removed. I'll update the PR in a sec.

tessus:
here you go... PR updated.

PR #5860, opened 2017-07-15
Adding warning about keeping new default skins versions

sixtyfive:
(See https://github.com/roundcube/roundcubemail/issues/5394#issuecomment-239166245 for the reasoning - it seems I forgot to ever make this PR. No hard feelings at all if it's rejected right away.)


alecpl:
Well, the same would apply to all folders, not only skin. I think the warning is not needed.

PR #5867, opened 2017-07-18
Added .log extension to log files, for better handling by logrotate

luckysimek:


alecpl:
It's already implemented. See a few lines above your change.

PR #5874, opened 2017-07-23
[WIP] Rewrite Dockerfile

J0WI:


J0WI:
@alecpl would you like to give me any feedback, if it's worth to spent more time on this?

alecpl:
Personally I'm not interested in supporting Docker. I don't know Docker. Maybe @thomascube, @prashantjois @doublerr, @swarren83, @alunduil

thomascube:
I don't understand why the new Dockerfile downloads the package from Github. I suggest a Docker build using the integrated Dockerfile should use the sources and then basically perform the actions described in [Build From Source](/roundcube/roundcubemail/wiki/Build-from-source).

J0WI:
This is to avoid installing unnecessary packages, like Composer or Git. The result should be the same.
Compression of CSS/JS can be done by a reverse proxy.

thomascube:
Implemented in master therefore closing this PR. Thanks a lot for the inspiration!

PR #5879, opened 2017-07-26
use list_widget.get_selection() rather than list_widget.selection

johndoh:
related to [contextmenu/#87](https://github.com/JohnDoh/Roundcube-Plugin-Context-Menu/issues/87). now there is a getselection hook in the list widget the get_selection() method should be used rather than accessing list.selection directly.

### DIFF
--- a/plugins/acl/acl.js
+++ b/plugins/acl/acl.js
@@ -157,8 +157,8 @@ rcube_webmail.prototype.acl_list_init = function()
 // ACL table row selection handler
 rcube_webmail.prototype.acl_list_select = function(list)
 {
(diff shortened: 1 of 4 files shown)


alecpl:
Thanks for working on this. Are you sure we should not use get_selection(false) in these places? That would definitely be safer.

JohnDoh:
Good point. I'll look into that. For sure all the ones that are not on the message list should have false. Perahps just to keep things as similar as possible to how they are currently then they should all have false.

JohnDoh:
added false for get_selection() on message list object

PR #5882, opened 2017-07-27
Split steps to install jsdeps

hefee:
For building Debian packages, we can't download the jsdeps during the building phase. That's why we need to split the get and extract steps. Also for Debian packages, there is a default place, where 3rdparty code should be put.

### DIFF
--- a/bin/install-jsdeps.sh
+++ b/bin/install-jsdeps.sh
@@ -44,9 +44,13 @@ if (empty($UNZIP)) {
 if (empty($FILEINFO)) {
   die("ERROR: Required program 'file' not found\n");
 }
+
 $CACHEDIR = sys_get_temp_dir();
 
-if (is_writeable(INSTALL_PATH . 'temp/js_cache') || @mkdir(INSTALL_PATH . 'temp/js_cache', 0774, true)) {
+if (getenv("CACHEDIR") != false && is_writeable(getenv("CACHEDIR"))) {
+  $CACHEDIR = getenv("CACHEDIR");
+}
+else if (is_writeable(INSTALL_PATH . 'temp/js_cache') || @mkdir(INSTALL_PATH . 'temp/js_cache', 0774, true)) {
   $CACHEDIR = INSTALL_PATH . 'temp/js_cache';
 }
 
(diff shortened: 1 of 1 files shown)


bin/install-jsdeps.sh
```diff
@@ -44,16 +44,16 @@ if (empty($UNZIP)) {
...
-  die("ERROR: Required program 'wget' or 'curl' not found\n");
-}
 
 $CACHEDIR = sys_get_temp_dir();
 
 if (is_writeable(INSTALL_PATH . 'temp/js_cache') || @mkdir(INSTALL_PATH . 'temp/js_cache', 0774, true)) {
   $CACHEDIR = INSTALL_PATH . 'temp/js_cache';
 }
 
+if (getenv("CACHEDIR") != false) {
```
alecpl: You have to move this up.  So, mkdir() does not create a js_cache directory when CACHEDIR env is set. I think it would be good also to check if the CACHEDIR from env is writeable.

bin/install-jsdeps.sh
```diff
@@ -295,8 +304,8 @@ function delete_destfile($package)
 
 //////////////// Execution
 
-$args = rcube_utils::get_opt(array('f' => 'force:bool', 'd' => 'delete:bool'))
-        + array('force' => false, 'delete' => false);
+$args = rcube_utils::get_opt(array('f' => 'force:bool', 'd' => 'delete:bool', 'g' => 'get:bool', 'e' => 'extract:bool'))
+        + array('force' => false, 'delete' => false, 'get' => true, 'extract' => true);
```
alecpl: This will not allow you to run the script for either get or extract. Will it?

bin/install-jsdeps.sh
hefee: should be solved now

bin/install-jsdeps.sh
hefee: I didn't wanted to change the way the script was used before that's why I made get and extract true by default. In the Debian case I have to make it like this, it is not nice but works and i don't mess up other usecases.

to extract:
install-jsdeps.sh --get=0 --extract=1

to get:
install-jsdeps.sh --get=1 --extract=0

to clean:
install-jsdeps.sh --get=0 --extract=0 --delete=1

alecpl:
I modified slightly the change so you can now just use -g or -e or -d.

PR #5890, opened 2017-08-04
Change answered icoon (wide screen layout Roundcube 1.3)

ghost:
Hello everybody,

We used the wide screen layout option. But one thing is against us! If an email has been answered, you will not see that. (Email is then unread)

I think this is better:

roundcube

I think that should be easy to make that!

Thank you! We hope you will changed that.


alecpl:
You PR is broken. You should use master branch only. Anyway, this is the same in Desktop view. If the message is replied and unread at the same time unread icon is displayed. Indeed you can have a separate Read column then you see both icons. However, the case when a message is unread and replied at the same time is quite rare, no?

dave670:
Thank you for the reply! We sometimes answer messages in the sense that we are coming back. Then we put him back on unread. At the desktop view you will see both icons. 

With the wide screen layout option you didn't see both icons. (only unread icon)

PR #5895, opened 2017-08-10
accumlated patches from Debian

hefee:
Since a while we have some patches in the pipeline to patch rcm. This are a first bunch of patches. It makes sense to review them on a single commit base.

I think at least the last two "make update script work together with debian package manager." and "Set INSTALL_PATH to /var/lib/roundcube/ in bin/cleandb.sh" need to be reworked to enter rcm.

from the postinst script we call update.sh:
```
DEBIAN_PKG=TRUE RCMAIL_CONFIG_DIR=/etc/roundcube /usr/bin/php /usr/share/roundcube/bin/update.sh --version=${OLD_UPSTREAM_VERSION} --accept=true
```

### DIFF
--- a/bin/cleandb.sh
+++ b/bin/cleandb.sh
@@ -19,7 +19,7 @@
  +-----------------------------------------------------------------------+
 */
 
-define('INSTALL_PATH', realpath(__DIR__ . '/..') . '/' );
+define('INSTALL_PATH', '/var/lib/roundcube/' );
 
 require INSTALL_PATH.'program/include/clisetup.php';
 
--- a/bin/update.sh
+++ b/bin/update.sh
@@ -18,6 +18,13 @@
  | Author: Thomas Bruederli <roundcube@gmail.com>                        |
  +-----------------------------------------------------------------------+
 */
+if (getenv('RCMAIL_CONFIG_DIR')){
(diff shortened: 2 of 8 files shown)


alecpl:
We definitely will not apply the last commit. I'm not yet sure what about the previous to last. @thomascube what do you think?

thomascube:
I'd also reject "Avoid fetching jQuery from Google, use the embedded one". This is an example page which most likely will live outside the Roundcube docroot or environment. Having relative paths to Roundcube resources is IMO counter intuitive for an example like this.

Please split this PR up into individual commits we can decide about.

hefee:
I don't get it why it is easier to talk in individual PRs you can give me at this state a fist thumb up for some commits than I'll split them out of this PR. And also if you want one commit to be discuss in more detail, i'll split this. Do you have any issues with the first three commits, to enter rcm?

jpggithub:
IMHO it is not very secure to directly call Google stuff or other external providers.
At least, the users must be aware of this calls or have an option to desactivate them.

dereks:

```
git fetch upstream
git rebase upstream/master
```

alecpl:
Wontfix. Most of it I don't like.

PR #5899, opened 2017-08-12
managesieve: fix bug when opening exists rule

johndoh:
if you create a rule in managesieve and set the test the to exists.

For example Subject Exists. Save it. navigate away and then open that rule up again the form will load with Subject Contains.

This is a little tweak just to make it load the form with Subject Exists.

### DIFF
--- a/plugins/managesieve/managesieve.js
+++ b/plugins/managesieve/managesieve.js
@@ -677,7 +677,7 @@ function rule_header_select(id)
     dateheader.style.display = h == 'date' ? '' : 'none';
 
   $('[value="exists"],[value="notexists"]', rule).prop('disabled', h == 'string');
-  if (!rule.val() || rule.val().match(/^(exists|notexists)$/))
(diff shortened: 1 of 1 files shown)


plugins/managesieve/managesieve.js
```diff
@@ -677,7 +677,7 @@ function rule_header_select(id)
     dateheader.style.display = h == 'date' ? '' : 'none';
 
   $('[value="exists"],[value="notexists"]', rule).prop('disabled', h == 'string');
-  if (!rule.val() || rule.val().match(/^(exists|notexists)$/))
+  if (!rule.val() && !rule.val().match(/^(exists|notexists)$/))
```
alecpl: This code does not make sense now. I think the point was to reset the select to 'contains' if the selected value was disabled. In such cases the `rule.val()` will return `null` and there will be a javascript error. So, I think just
```
if (!rule.val())
    rule.val('contains');
```
should be enough. Thanks for spotting this issue.

alecpl:
Fixed.

PR #5902, opened 2017-08-14
password plugin ldap driver: use full %login as %name/%n when no @ in…

buborh:
… username.


buborh:
Our mail server accept both username@domain and username in login. We store username in ldap.
I use %login in DN settings,  password change failes when I login with only username ( without @domain).
I use %username in DN settings, password change failes when I login with username@domain.

PR #5903, opened 2017-08-16
extend disabled_actions config

johndoh:
Occasionally there are buttons in the skin that do not have a command attribute. Things that do skin actions like opening popups. My idea is to fall back to the button name when no command is present so that these buttons can still be disabled via config.

In addition I added a check that ignores the $task so if you want to disable a button that is in multiple tasks or you want to remove something that a plugin might add in mutiple tasks or if you are just not so good at Roundcube and don't understand about the tasks.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1358,12 +1358,14 @@ EOF;
         }
 
         $command = $attrib['command'];
+        $action = $command ?: $attrib['name'];
 
         if ($attrib['task']) {
-            $element = $command = $attrib['task'] . '.' . $command;
+            $command = $attrib['task'] . '.' . $command;
+            $element = $attrib['task'] . '.' . $action;
         }
         else {
-            $element = ($this->env['task'] ? $this->env['task'] . '.' : '') . $command;
+            $element = ($this->env['task'] ? $this->env['task'] . '.' : '') . $action;
         }
 
         if ($disabled_actions === null) {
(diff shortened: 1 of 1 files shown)


program/include/rcmail_output_html.php
```diff
@@ -1357,7 +1357,7 @@ public function button($attrib)
             $attrib['type'] = ($attrib['image'] || $attrib['imagepas'] || $attrib['imageact']) ? 'image' : 'button';
         }
 
-        $command = $attrib['command'];
+        $command = $attrib['command'] ?: $attrib['name'];
```
alecpl: I don't think you can modify $command variable this way. See line 1456.

alecpl (changes requested):

program/include/rcmail_output_html.php
JohnDoh: agreed. I missed that. What do you think about this?
```patch
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1451,7 +1451,7 @@ EOF;
             if ($attrib['classact'])
                 $attrib['class'] = $attrib['classact'];
         }
-        else if ($command && !$attrib['onclick']) {
+        else if ($attrib['command'] && !$attrib['onclick']) {
             $attrib['onclick'] = sprintf(
                 "return %s.command('%s','%s',this,event)",
                 self::JS_OBJECT_NAME,
```
I think it do not think it makes sense to create an onclick attrib when there was no command specified (and the contents of $command could only be `task.` or empty)

alecpl: It could work, but I'd just set `$element` appropriately instead of `$command` if `$attrib['command']` is empty, because this variable is used to check disabled actions.

PR #5904, opened 2017-08-16
replace display_version with display_product_version

johndoh:
Currently it is possible to hide the version number on the login screen by config but not the product name. If, for example, the logo already contains the product name then displaying it again is redundant. I did not want to introduce another config option so my idea is to replace display_version with display_product_version and then allow for 3 settings: nothing, product name only (default, just as now), and product and version.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -439,8 +439,9 @@ $config['login_rate_limit'] = 3;
 // Includes should be interpreted as PHP files
 $config['skin_include_php'] = false;
 
-// display software version on login screen
-$config['display_version'] = false;
+// display product name and software version on login screen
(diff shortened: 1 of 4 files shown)


config/defaults.inc.php
```diff
@@ -439,8 +439,9 @@
 // Includes should be interpreted as PHP files
 $config['skin_include_php'] = false;
 
-// display software version on login screen
-$config['display_version'] = false;
+// display product name and software version on login screen
+// 0 - hide product name and version number, 1 - show product name only, 2 - show product name and version number
+$config['display_product_version'] = 1;
```
alecpl: I propose to rename the option to 'display_product_info'. Other than that it looks good.

PR #5922, opened 2017-08-25
improve detection for Egde browser and add pointer event support (release-1.3) (credit JohnDoe)

dapphp:
add support for pointer events on list widget

use either pointer or touch events, not both

ensure bw.pointer is a boolean

Credit to @JohnDoh - Re-creating patch for release-1.3 branch.

### DIFF
--- a/program/js/common.js
+++ b/program/js/common.js
@@ -58,14 +58,15 @@ function roundcube_browser()
   this.dom = document.getElementById ? true : false;
   this.dom2 = document.addEventListener && document.removeEventListener;
 
-  this.webkit = this.agent_lc.indexOf('applewebkit') > 0;
+  this.edge = this.agent_lc.indexOf(' edge/') > 0;
+  this.webkit = !this.edge && this.agent_lc.indexOf('applewebkit') > 0;
   this.ie = (document.all && !window.opera) || (this.win && this.agent_lc.indexOf('trident/') > 0);
 
   if (window.opera) {
     this.opera = true; // Opera < 15
     this.vendver = opera.version();
   }
-  else if (!this.ie) {
+  else if (!this.ie && !this.edge) {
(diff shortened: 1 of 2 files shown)


dapphp:
We recently deployed 1.3.0 into production and a number of our users were plagued by this issue.  It almost completely prevents IE/Edge users on touch from opening emails in any layout.

If we could get this merged into 1.3 for future 1.3.x releases that would be great, otherwise 1.3 is unusable on Win10/IE/Edge.

So far all who were reporting problems are saying it's resolved.

dapphp:
I only have limited ability to test this, but someone with Windows 10, Edge 15.15063 on a Lenovo Flex 4 tablet/laptop reported that the above change didn't quite fix the issue.  They are using the "List" layout.  Since we received a number of complaints about double-click going away, I backported in Thomas' change to revive it.  In List view on the tablet, the only way to open the message is to double click.  However, in IE 11 and Edge 15, double tapping on the screen resulted in a zoom action instead of a click event.  Chrome on the same tablet did not suffer from this and instead would open the message.

In order to resolve the issue for that person, I added the following CSS to a global stylesheet.

    body {
        -ms-touch-action: manipulation !important;
        touch-action: manipulation !important;
    }

It may be possible to make the selector more specific to the message list div but I didn't try.

I personally confirmed that on a Microsoft Surfacebook tablet/laptop running Win 10 and Edge 15 that double tapping on the message list in tablet mode & list layout didn't result in a zoom action but have not tested that on a fresh install of Roundcube under the conditions of 1) list layout 2) no double click backported in and 3) without the above touch-action CSS.

Other tablet users reported that only this patch worked for them, but the Flex user still got the zoom action on double click and not the click action.

I'll see if I can do further testing in the coming week but there may still be outstanding issues on touch tablets with the message list in List or Desktop view.

PR #5934, opened 2017-09-03
Fix typo in de_CH

dol:


alecpl:
Please use our translation service at https://www.transifex.com/roundcube/roundcube-webmail.
Fixed this right now there.

dol:
@alecpl Thank you for the hint of the translation service. I'll use it in the future. Thank you.

PR #5935, opened 2017-09-04
Add Travis CI

Slamdunk:
Hi, this PR is intended to add Travis continuous integration builds to this github repository.
If this PR is going to be merged, the maintainer should activate travis triggers for pushes and pull-requests.

As you can see in the build done against my fork:
https://travis-ci.org/Slamdunk/roundcubemail/builds/271667707
there are failures that should be addressed, expecially on the newcomer PHP 7.2

This PR is **not** intended to solve the issues, but only to introduce the CI process. New PR will be opened after this one is merged and the triggers are activated; fixing bugs before having a running ci process should be avoided.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: php
+
+sudo: false
+
+matrix:
+  fast_finish: true
+  include:
+    - php: 5.4
+    - php: 5.5
+    - php: 5.6
+    - php: 7.0
+    - php: 7.1
+      env: CODE_COVERAGE=1
+    - php: 7.2
+
+cache:
+  directories:
+    - $HOME/.composer
+
+install:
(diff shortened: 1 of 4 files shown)


alecpl:
@thomascube That sounds like a good idea. Could you take care of it?

thomascube:
Brilliant! I'll have a look at this

thomascube:
A first build has run. See https://travis-ci.org/roundcube/roundcubemail/builds/281293683

Slamdunk:
@thomascube At the time being I'm a bit busy and can't help with proper PR to fix 5.x errors, but I can suggest to try mcrypt fallback like zend-math did: https://github.com/zendframework/zend-math/blob/release-2.7.0/src/Rand.php#L35-L62

thomascube:
The issue was much simpler and is already fixed in 3723f3f178. Thanks for your support!

PR #5936, opened 2017-09-04
Fix repeated word in language file de_CH

dol:


alecpl:
Please use our translation service at https://www.transifex.com/roundcube/roundcube-webmail.
Fixed this right now there.

PR #5964, opened 2017-09-25
Enable quiet running

Takika:


alecpl:
Maybe if that was implemented in all our scripts we could consider it a good addition, but only for one script it does not make much sense.

PR #5966, opened 2017-09-25
Add special chars as exception at spellcheck

maureenwonderland:
We are facing a problem that sometimes it's a little bit difficult to be reproduced.
Sometimes an error occurs and sometimes not.
We are using spell checker (pspell) and when trying to send an email with some specific special chars like º or ª, it crashes and we are unable to send emails. We realize that this error occurs on pspell_suggest() function from pspell's library. In the is_exception() function there is no condition that can handle those special chars.
So, it was added those chars as exception.

### DIFF
--- a/program/lib/Roundcube/rcube_spellchecker.php
+++ b/program/lib/Roundcube/rcube_spellchecker.php
@@ -255,7 +255,7 @@ class rcube_spellchecker
     }
 
     /**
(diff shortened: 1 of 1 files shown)


alecpl:
You can't put multibyte chars in regex like this.

alecpl:
Wontfix.

Issue #5766, opened 2017-05-18
enhancement to smb password driver


nbensa:
I'm having the same problem as the person in this post:

http://www.roundcubeforum.net/index.php?topic=24019.0

I quote the post:

> Currently my installation used Samba 4.2 for user directory.
>
> The smb engine in password plugin sends username@domainname as the user name to change password and my Samba fails to accept it.
> I introduced a mod to the plugin by adding the second line (below) in the drivers/smb.php:
>
> $username = $_SESSION['username'];
> // added little hack, this only keeps the login name as a paremeter
> $username = explode( '@', $username)[0];
        
> It would be nice if there was a parameter allowing the reduction of session user name to only user name for this password change.
> The parameter 'session_domain' in global roundcube configuration does not do the trick

alecpl:
Implemented via more generic password_username_format option.

PR #6008, opened 2017-10-21
Password: add option to configure username sent to change password...

dsoares:
... in samba driver (#5766)


alecpl:
Implemented via more generic password_username_format option.

Issue #5752, opened 2017-05-05
Attackers might be able to set arbitrary SMTP servers


Arinerron:
https://github.com/roundcube/roundcubemail/blob/master/config/defaults.inc.php#L128

`$_SERVER['SERVER_NAME']` can be modified via the `Host` header. I'm not sure about the severity of this issue, but it's worth reporting.

See:
https://expressionengine.com/blog/http-host-and-server-name-security-issues
http://php.net/manual/en/reserved.variables.server.php
![screenshot](https://cloud.githubusercontent.com/assets/3926753/25749599/ed8d7c94-3163-11e7-9ca2-7bfa8374898c.png)

alecpl:
I don't know if we can make Roundcube more secure by default in this context. It looks that the only option to prevent issues related to Host header spoofing is to implement host checker, like trusted_host_patterns in Drupal: https://www.drupal.org/node/1992030.

phoikoi:
Just for purposes of research and planning on this issue, it might be worth noting that Django (the Python web framework) also has [a similar config setting](https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts).  It was introduced several years ago; here is [a link to the release notes for that version](https://docs.djangoproject.com/en/1.11/releases/1.4.4/) in case it might be helpful to look at their thinking on the issue in general.  Hope it helps you.

Jean-Daniel:
This issue can be easily mitigated by properly configuring your web server as the Host header is also the one used to route to the target vhost. If your webmail is hosted on anything but the default vhost, the Host header is guarantee to be the vhost server name.

alecpl:
Implemented.

PR #6009, opened 2017-10-21
Add option trusted_host_patterns

dsoares:
Related to #5752.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -475,6 +475,11 @@ $config['ip_check'] = false;
 // X_FORWARDED_* and X_REAL_IP headers are only accepted from these IPs
 $config['proxy_whitelist'] = array();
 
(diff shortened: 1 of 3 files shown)


alecpl:
I think we should use the new method also in `rcube_utils::parse_host()`. And because of that it would make sense to do `preg_replace('/:\d+$/', '', $server_name);` in your new method (optionally maybe).

As you see in `rcube_utils::parse_host()` we also use `$_SERVER['HTTP_HOST']`. Maybe another case for an option in your new method, so we could use it for both.

dsoares:
I did some modifications accordingly. I'm not satisfied with the result, in the sense that calling `rcube_utils::server_name($_SERVER['HTTP_HOST')` is weird. However, i did not want to change the so simple and direct function name `server_name`.

PR #6010, opened 2017-10-21
Added $config['max_upload_size']

chtisgit:
The new config value can be used to further restrict the
maximum size of attachments. So you can increase the
post_max_size and upload_max_filesize in php.ini without
affecting RoundCube.

I'm currently running Roundcube with this patch, because
unfortunately -- due to another PHP script -- I need the
_post_max_size_ and _upload_max_filesize_ settings at a
level that my mail server would refuse to handle (and I 
don't want users to try to upload insanely large attachments).

Is there a simpler way to achieve the same functionality?
I can't change these two settings via ini_set() ...

### DIFF
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -84,3 +84,8 @@ $config['plugins'] = array(
 
 // skin name: folder from skins/
 $config['skin'] = 'larry';
+
+// restrict maximum upload file size
+// if set to a negative value, the maximum file size of attachments is
+// determined by the settings in your php.ini only
+$config['max_upload_size'] = -1;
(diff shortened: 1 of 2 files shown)


alecpl:
How about .htaccess/.user.ini?

chtisgit:
Oh, I didn't know about .user.ini, **thank you!**
So this patch is probably irrelevant, feel free to close it.

For those stumbling upon this pull request: After enabling .user.ini files in php.ini and restarting php-fpm, I placed a .user.ini file in the root directory of roundcube, in which I adjusted the settings post_max_size and upload_max_filesize for roundcube.

PR #6011, opened 2017-10-23
Allow for free-form flag filter action

daniele-athome:
This small patch adds a text input field to the managesieve plugin, allowing a user-defined flag (actually keyword in this case) to be set on a message, along with the default IMAP existing flags.

Signed-off-by: Daniele Ricci <daniele.athome@gmail.com>

### DIFF
--- a/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
+++ b/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
@@ -2133,6 +2133,12 @@ class rcube_sieve_engine
                 ))
                 . rcube::Q($this->plugin->gettext('flag'.$fidx)) .'<br>';
         }
+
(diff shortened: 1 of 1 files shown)


alecpl:
I guess it would make more sense to make the field a data-type="list" textarea. We already do this for other multi-input fields, see the code for examples.

daniele-athome:
Makes sense. I'll work on it.

daniele-athome:
How about adding another 3 actions called "add/set/remove keyword on message" with a single text input? Seems more in line with other actions behaving in a similar way, e.g. "send message copy to". That way we separate flags from keywords (even if they are in fact IMAP flags).

alecpl:
I wouldn't do that because the script can have flags and keywords in the same action, and I don't want to have to split them. I also do not want a long selection list.

daniele-athome:
Done, with list input!

PR #6022, opened 2017-11-03
add charset to watermark.html

johndoh:
this is trival but the console in firefox gives a warning because there is no charset defined for watermark.html: 

```The character encoding of a framed document was not declared. The document may appear different if viewed without the document framing it.```

so I've added a utf-8 charset to the HTML.


alecpl:
I thought it might require Content-Type header use, but it looks working. Thanks.

peterkroon:
Note:
I've got version ``Roundcube Webmail 1.3.6`` installed via ``DirectAdmin`` without this commit.

PR #6027, opened 2017-11-07
Managesieve forward settings

### LINKED ISSUE #6021: managesieve: new settings dialog for simple forwarding of emails
I have created a settings dialog very similar to the vacation settings.

In this forwarding dialog you can set very simple forwarding, only `redirect to` or `send copy to`, and setting the status to `on` or `off`.

I needed to do this for a customer who often uses forwarding and says that the filters are too powerful for them to configure.

I will create a push request for this, perhaps it is useful for other users too. It can be enabled/disabled in the `managesieve.inc.php`, similar to vacation.

tpokorra:
fixes #6021


alecpl:
Could you add classic skin support, please.

tpokorra:
No problem, here is the classic skin support

PR #6036, opened 2017-11-09
Fix typos on optional security header comment

jamesmacwhite:
Couple of typo's on the `.htaccess` in relation to the optional security headers.


alecpl:
Your change to the second sentence does not make sense to me.

jamesmacwhite:
@alecpl Hi. Thanks for the comment.

On a second look, you're right, I was interpreting that statement differently. I've update the changes to the comment header.

I also believe that because there are several headers that are deemed security related, it's appropriate for the comment section to be `Optional security headers` as in plural.

Let me know if that's OK.

PR #6037, opened 2017-11-11
Add optional SSL/TLS redirect to .htaccess

jamesmacwhite:
This adds an optional RewriteRule for essentially running Roundcube under HTTPS. Given that the SSL setup of different configurations (or the lack of), this is commented out by default, but I thought might be useful to have?

The RewriteRule should be fairly flexible. In my case, I have Roundcube deployed with DirectAdmin and its `/roundcube` path is accessible via the server hostname, as well as individual client domains. Having this code deployed, allows the client domains to utilise there own SSL certificate to run Roundcube under SSL and automatically redirect plain HTTP requests to the `/roundcube` path.

Hope this might be useful!


alecpl:
Roundcube already has `force_https` option. I don't know, maybe it would make sense to remove it and leave this to http server configuration.

PR #6038, opened 2017-11-11
Replace outdated referrer policy header

jamesmacwhite:
The way the referrer policy example is set is outdated, there is a more standard "Referrer-Policy" header to do it.

I can still include the "Content-Security-Policy" header with some basic defaults commented out, so to not remove CSP being there, but wasn't sure.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -82,8 +82,12 @@ Header set X-Robots-Tag "noindex, nofollow"
 # It prevents Google Chrome and Internet Explorer from trying to mime-sniff the content-type of a response away from the one being declared by the server.
 #Header set X-Content-Type-Options: "nosniff"
 
-# CSP - Content Security Policy
+# Referrer policy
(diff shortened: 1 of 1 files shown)


Rotzbua:
Hi,

I added the template. The basic intention was to give admins a template to (optional) secure the mail application. Here a good testing site for recommended additional secure headers: https://securityheaders.io/

I agree, my template is a bit outdated. The csp is v1 so it should be updated to v2: https://www.w3.org/TR/CSP2/

Two improvements:

* By default there should be also a wildcard for images and fonts, because otherwise mails with external pictures could not be shown.

* By default the referrer should have another value than "no-referrer" because it will be an issue with rc referrer check:
https://github.com/roundcube/roundcubemail/blob/515d496808cfc4b188d4cc7116329f101fd042b8/config/defaults.inc.php#L479

Best.

alecpl:
See 186f21c4c.

PR #6040, opened 2017-11-13
use skined confirmation boxes where possible

johndoh:
use simply dialog boxes to make skinnble confirm boxes rather than JS ```confirm()``` where possible


alecpl:
There are some more confirm() function uses in plugins. Also, I think it would make sense to create a wrapper on simple_dialog just for confirm dialogs. So, in the code we would only do:
```
this.confirm('confirm-text-label', 'confirm-title-label', function() { ... });
```
It could make the second argument optional. I.e. if 2nd argument is a function set title to this.get_label('confirmationtitle').

It may also make sense to pass the `window.rcmail` (`this`) object as the callback's argument. In such a case plugins wouldn't need to create a temp variable nor use `window.rcmail`. Or just set `this` context of the function to `window.rcmail`, so we can use `this` (I'm talking about `callback.call(ref)` use).

JohnDoh:
The `simple_dialog` function is already a wrapper for `show_popup_dialog` and I was nervous of putting wrappers on top of wrappers.

I suggest the title should be fixed in the wrapper function (do you have use case for changing it?) and there needs to be an param for the label of the primary button, so may be:
```js
this.confirm('confirm-text-label', 'confirm-button-label', function() { ... });
```
About the core plugins – I’ll check those, I wanted to see how open you were to the idea before doing too much. Plus there is also the possibility for doing something similar for the `alert()` boxes.

alecpl:
Thank you very much. I slightly modified the change and merged. Now, I'm going to merge dev-elastic into master.

PR #6068, opened 2017-11-28
Add GeoIP and verbose datas when logging user sendmail

beaujean:
Add GeoIP, UserAgent and verbose datas when logging user sendmail in order to accurately track spammers when they script RoundCubeMail to send spams.


dsoares:
I don't agree with this PR. Roundcube core shouldn't have any calls to `geoip*` functions because:
- it should be a roundcube admin's choice (eventually, with the new [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation), an user's choice);
- it demands extra configuration for the geoip library to work and that's not related to core Roundcube.

GeoIP functionality should be always optional.

philipwhiuk:
This does definitely feel like opt-in plugin functionality.

alecpl:
And it should be possible with write_log hook.

PR #6069, opened 2017-11-28
Add directory level hashes when storing temp files

beaujean:
For large deployment infrastructure (more than thousand users) it becomes mandatory to improve storing temporary files performances by adding directory level hashes when storing temp files. Most of the filesystems can't handle more than 1.000 files per directory. Plus thumbs and attachments are stored in different directories to allow different retention times when purging files.


alecpl:
This maybe should be an optional feature. Also, I guess rcube::gc_temp() needs some alignment. Finally, I see you modify paths, but I don't see a code that creates these subdirectories.

alecpl:
With current git-master you have to only modify rcube_utils::temp_filename() and rcube::gc_temp().

alecpl:
Incomplete. Maybe it should become a plugin feature.

PR #6071, opened 2017-11-28
Support additional connect parameters in PostgreSQL database wrapper

georgeto:
Most notably this change enables you to specify whether or with what
priority a secure SSL TCP/IP connection will be negotiated with the
database server.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 CHANGELOG Roundcube Webmail
 ===========================
 
+- Support additional connect parameters in PostgreSQL database wrapper
 - Use UI dialogs instead of confirm() and alert() where possible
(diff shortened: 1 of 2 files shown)


alecpl:
You should use single-quotes and there should be a space after `foreach` keyword. Other than that it looks good.

alecpl:
Ah, I'm not sure we can/should change PHP version requirement. Your change brings PHP 5.6 as a minimum. Maybe it's too early, I think we should stay with 5.4 for Roundcube 1.4. After that we probably could switch to PHP7-only.

alecpl:
You could keep the foreach() loop to make the code short. Please, also resolve conflict by adding the changelog entry on top of changelog.

PR #6089, opened 2017-12-12
check for minified CSS files

johndoh:
if a plugin has minified css files created by the bin/cssshrink.sh script, and the regular files are not present then the CSS is not included. the check added here is based on code in rcmail_output_html::file_mod()


johndoh:
It is also related to this change https://github.com/roundcube/roundcubemail/commit/69f50b122d62ebdcdbe708c78daaf9d8483fddb4 which adds existance checks (note, there have been other changes since this commit, but this is the original) but does not allow for minifiled versions files. previous to this change the `.min` was added to the file name in rcmail_output_html and the CSS included.

PR #6105, opened 2017-12-21
'imap_log_session' option doesn't work

tsnr:
Hi.

I found that the IMAP session user is not logged correctly and will be fixed with this patch.

Thank you,

### DIFF
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -481,8 +481,8 @@ class rcube
      */
     public function storage_log_session($args)
     {
-        if (!empty($args['session']) && session_id()) {
-            $this->write_log('imap_session', $args['session']);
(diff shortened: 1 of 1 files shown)


alecpl:
This is not user, this is session. See https://github.com/roundcube/roundcubemail/blob/master/program/lib/Roundcube/rcube_imap.php#L174-L185

alecpl:
This feature has been added specifically for Cyrus IMAP. I don't know if other servers support imap session identifiers. So, I guess that's why it's confusing. On supported server imap_session.log entry looks like this:
```
[22-Dec-2017 09:04:17,686173 +0100]: <8flve2lj> <kolab.example.org-26650-1513933457-1-10759407694059984249>
```
which gives us a way to match PHP session with IMAP session.

PR #6115, opened 2017-12-30
Update idn convert

kesselb:


alecpl:
Does it fix anything? If not, I don't like the callback idea.

danielkesselberg:
I dropped the callback. It does not fix anything at all.

alecpl:
I didn't check, but will that work for addresses as 'рф.ru@рф.ru', 'xn--p1ai.ru@xn--p1ai.ru'?

danielkesselberg:
Good catch. I added this as test case and restored the old version.

PR #6120, opened 2018-01-04
Migrate rcube_session_memcache to memcached

kesselb:
Memcache is not available for newer PHP and looks unmaintained. This PR migrates the rcube_session_memcache to Memcache**d** and adds a basic test.

rcube->get_memcache is unchangen because its used by other code. I realized that Memcache**d** provides a custom session handler: http://php.net/manual/de/memcached.sessions.php 

rcube_session_memcache now extends from rcube_session_php and configures session.save_handler and session.save_path for memcache. Using the custom session handler has some advantages:

- transparent for roundcube
- less code to maintain


alecpl:
I never used memcache as a session save handler. Is that "locking" as the default php session handler? I.e. does it lock the session for the time of the whole http request? If so, I don't like it and I think we should not remove the old way of using memcache.

danielkesselberg:
By default session_locking is enabled. After https://github.com/php-memcached-dev/php-memcached/issues/269 there are some issues with session_locking and php7. I added another commit and disable session locking.

alecpl:
Really, I'm not sure we can just disable memcached.sess_locking. This would require some testing. However, since we have support for PHP session, I'd really prefer to keep the custom memcache session handler. You can always use PHP session with save_handler=memcache.

On Ubuntu I have both memcache and memcached extensions installable with PHP 7.2, so I don't see where's the problem, but I guess we can consider a switch to memcached.

danielkesselberg:
Thanks for your thoughts ;-)

> On Ubuntu I have both memcache and memcached extensions installable with PHP 7.2, so I don't see where's the problem, but I guess we can consider a switch to memcached.

Oh 😞 I didn't know that. The package is built from https://github.com/websupport-sk/pecl-memcache (an unofficial repository with patches for php7)

> You can always use PHP session with save_handler=memcache.

I guess this is the best solution. 

> I'd really prefer to keep the custom memcache session handler

I think it would be better to remove the custom memcache (and redis) session handler. In my opinion the custom implementation does nothing better and adds unnecessary application logic.

PR #6136, opened 2018-01-14
Fix warning "strtolower() expects parameter 1 to be string, array given"

Vanav:
$rule['arg1'] may be an array of headers.

### DIFF
--- a/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
+++ b/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
@@ -1653,7 +1653,7 @@ class rcube_sieve_engine
                     $rule['arg1'] = $rule['arg1'][0];
                 }
 
-                $matches = ($header = strtolower($rule['arg1'])) && isset($this->headers[$header]);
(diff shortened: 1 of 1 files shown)


plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
```diff
@@ -1653,15 +1653,15 @@ function rule_div($fid, $id, $div = true, $compact = false)
...
+                $matches = !is_array($rule['arg1']) && ($header = strtolower($rule['arg1'])) && isset($this->headers[$header]);
                 $test    = $matches ? $header : '...';
             }
             else if ($rule['test'] == 'exists') {
                 if (is_array($rule['arg']) && count($rule['arg']) == 1) {
                     $rule['arg'] = $rule['arg'][0];
                 }
 
-                $matches = ($header = strtolower($rule['arg'])) && isset($this->headers[$header]);
+                $matches = !is_array($rule['arg1']) && ($header = strtolower($rule['arg'])) && isset($this->headers[$header]);
```
alecpl: Here we use `arg` not `arg1`.

Issue #6154, opened 2018-02-02
The save button from Edit Filter Set is not visible using Classic skin


ssnsaini:
Hi
When using the "Edit filter set" within the managesieve plugin the save button is not visible, it's actually hidden underneath the text box. See attached screenshots.
Regards

![normal view](https://user-images.githubusercontent.com/13794586/35743536-d0b55804-0835-11e8-919a-0ae79c1f55e0.png)

![resize text form to reveal save button](https://user-images.githubusercontent.com/13794586/35743540-d2e0102e-0835-11e8-9b85-887348e5d47a.png)

Jan-E:
https://github.com/roundcube/roundcubemail/pull/6164 should fix this.
@alecpl Please review

PR #6164, opened 2018-02-06
Managesieve: Fix seteditraw classic save button

Jan-E:
Fixes https://github.com/roundcube/roundcubemail/issues/6154


ssnsaini (approved):
the fix works - thanks.

PR #6175, opened 2018-02-13
Patched bug where rcube_db::quote() was causing an infinite loop.

laodc:
As rcube_db::quote() checks to see if the connection is up before quoting, this would cause the class to try connect again, as rcube_db::$dbh was not being set until AFTER conn_configure was completed, causing a loop.

So updated rcube_db::$dbh in the rcube::conn_create() function instead allowing access to the new object straight away.


alecpl:
Because of edeb5d7ab4. Good catch.

PR #6180, opened 2018-02-15
add some new events to list/treelist widgets

johndoh:
make it so that the event `getselection` is triggered when either `rcube_list_widget::get_selection()` or `rcube_list_widget::get_single_selection()` is called

### DIFF
--- a/program/js/list.js
+++ b/program/js/list.js
@@ -139,6 +139,8 @@ init: function()
     this.list.parentNode.onclick = function(e) { me.focus(); };
   }
 
+  rcmail.triggerEvent('initlist', { obj: this.list });
+
   return this;
 },
 
@@ -1311,8 +1313,10 @@ get_selection: function(deep)
  */
 get_single_selection: function()
 {
-  if (this.selection.length == 1)
-    return this.selection[0];
+  var selection = this.get_selection(false);
+
+  if (selection.length == 1)
+    return selection[0];
   else
     return null;
 },


johndoh:
new events:

`init_list` - triggered when a new list widget is created. required so that one plugin can interact with a list created by another plugin. If one plugin creates a list widget in the rcmail.onload event then another plugin might not be able to detect it eixists from its own onload event. a global init_list event lets plugins listen out for the list they want.

`initrow` - add an initrow even to treelist widget, similar to the initrow event in list widget

alecpl:
When the treelist widget is created on the existing structure (like we do with folders list) the initrow event will not be triggered. Maybe we need it in another place for consistency?

johndoh:
perhaps in both `insert` and `walk_list`? `insert` for new items added after page load and `walk_list` for the inital creation?

johndoh:
May be not. It seems the reset() is called each time a new folder is added - so the `initrow` is triggered for every row every time.

Change of plan. What if there was a `insertrow` event in `rcmail.add_folder_row()` and then something like `initfolderlist` in `rcmail.init_subscription_list()`?

Shall I move the treelist stuff to a separate PR so the list widget stuff can be merged?

johndoh:
i've removed the treelist stuff and tidied this PR up a bit

alecpl:
As for the treelist changes. Adding insertrow for folders list may broke some plugins that expect this event for messages list only. Imo, `initrow` in treelist widget would be the best solution. We just need to make sure it's triggered for all records.

Issue #5779, opened 2017-05-31
managesieve plugin does not work with gssapi / krb_authentication plugin


dl8bh:
Hi guys,

after I finally managed to configure gssapi / krb_authentication correctly, I failed to get the managesieve plugin running with this setup.
After a bit of research, I found out, that this functionality is not included in the managesieve plugin.

For gssapi / Kerberos enabled environments, this would be really nice to have.

Best regards,

Bernd

alecpl:
We have some code in rcube_imap_generic so it shouldn't be very hard to copy it to Net_Sieve. Patches welcome.

neilmunday:
Hi,

As suggested I have updated the Net_Sieve class and also the ManageSieve plugin to work with my modification.

See the two pull requests above for details.

alecpl:
Done. Requires Net_Sieve 1.4.3.

dl8bh:
Wow, cool, will definitely check this out!

PR #6184, opened 2018-02-17
Added GSSAPI support for issue #5779

neilmunday:
Ref: issue #5779 

Notes:

- requires updated Net_Sieve class see https://github.com/roundcube/Net_Sieve/pull/2 for updated class.
- requires krb_authentication plugin
- set $config['managesieve_auth_type'] = 'GSSAPI'; in config.inc.php to use

### DIFF
--- a/plugins/krb_authentication/krb_authentication.php
+++ b/plugins/krb_authentication/krb_authentication.php
@@ -24,6 +24,7 @@ class krb_authentication extends rcube_plugin
         $this->add_hook('authenticate', array($this, 'authenticate'));
         $this->add_hook('login_after', array($this, 'login'));
         $this->add_hook('storage_connect', array($this, 'storage_connect'));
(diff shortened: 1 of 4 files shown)


alecpl:
I think it requires some changes to krb_authentication plugin so it configures managesieve_connect hook parameters. Right? It would also force auth_type automatically.

Could you also update the code to set CN param added in https://github.com/pear/Net_Sieve/commit/16aa77535038eb09276e8b3dc8311752f41b2a49 ?

neilmunday:
I'll see what I can do.

neilmunday:
I've added the requested changes and update managesieve to use the latest version of the Sieve class from Net_Sieve.

Please also this pull request https://github.com/pear/Net_Sieve/pull/7 which fixes a typo in the Sieve.php class. Without this typo fixed, the managesieve plugin will not work.

alecpl:
Please, in future don't use tabs in the code.

PR #6187, opened 2018-02-19
Use JQ() instead of Q() in rcmail_js_message_list

dsoares:
`rcmail_js_message_list()` returns JS commands to add rows to the message list; it should use `rcube::JQ()` instead of `rcube::Q()` to replace unicode and line separators in the generated string.

This PR fixes this bug: when the message subject contains the LINE_SEPARATOR or PARAGRAPH_SEPARATOR character (`\u2028` or `\u2029`), Roundcube fails to display the message list with a Javascript error. 

Test case: 

    Subject: =?utf-8?Q?PRISM_-_Toolkit_for_evaluating_the_outcomes_and_im?=
    =?utf-8?Q?pacts_=E2=80=A9of_small/medium-sized_conservation_projects?=

### DIFF
--- a/program/steps/mail/func.inc
+++ b/program/steps/mail/func.inc
@@ -513,7 +513,7 @@ function rcmail_js_message_list($a_headers, $insert_top=false, $a_show_cols=null
             else if ($col == 'subject') {
                 $cont = trim(rcube_mime::decode_header($header->$col, $header->charset));
                 if (!$cont) $cont = $RCMAIL->gettext('nosubject');
-                $cont = rcube::Q($cont);
+                $cont = rcube::JQ($cont);
             }
             else if ($col == 'size')
                 $cont = $RCMAIL->show_bytes($header->$col);
@@ -526,10 +526,10 @@ function rcmail_js_message_list($a_headers, $insert_top=false, $a_show_cols=null
                     $last_folder_name = str_replace($delimiter, " \xC2\xBB ", $last_folder_name);
                 }
 
-                $cont = rcube::Q($last_folder_name);
+                $cont = rcube::JQ($last_folder_name);
             }
             else
-                $cont = rcube::Q($header->$col);
+                $cont = rcube::JQ($header->$col);
 
             $a_msg_cols[$col] = $cont;
         }


dsoares:
If you have a message with that subject in a folder, Roundcube fails to list the messages with a Javascript error:

```
servername/:7 Uncaught SyntaxError: Invalid or unexpected token
    at rcube_webmail.http_response (app.js?s=1518801751:8982)
    at Object.success (app.js?s=1518801751:8932)
    at u (jquery.min.js?s=1518802099:36)
    at Object.fireWith [as resolveWith] (jquery.min.js?s=1518802099:36)
    at k (jquery.min.js?s=1518802099:36)
    at XMLHttpRequest.<anonymous> (jquery.min.js?s=1518802099:36)
jquery.min.js?s=1518802099:36 XHR finished loading: POST "https://servername/?_task=mail&_action=refresh".
(anonymous)
send @ jquery.min.js?s=1518802099:36
ajax @ jquery.min.js?s=1518802099:36
http_request @ app.js?s=1518801751:8930
http_post @ app.js?s=1518801751:8943
refresh @ app.js?s=1518801751:9609
(anonymous) @ app.js?s=1518801751:9581

```

Roundcube did reply to the request with a JSON response and in the exec commands you'll see something like:

    this.add_message_row(7,{\"subject\":\"PRISM - Toolkit for evaluating the outcomes and impacts <U+2029>of small/medium-sized conservation projects\",

<U+2029> is not visible to the eye, so you only see something like 2 spaces:

    this.add_message_row(7,{\"subject\":\"PRISM - Toolkit for evaluating the outcomes and impacts  of small/medium-sized conservation projects\",

And the Console shows the Javascript error above.

Browsers: 
- Google Chrome on WIndows (did not see the version)
- Google Chrome 63 on MacOS
- Safari 11 on MacOS
- Firefox 58 on MacOS (Error is: _SyntaxError: unterminated string literal_)

Roundcube's code does already prevent this characters (see `rcube_utils::rep_specialchars_output()` when it's called with `$enctype='js'`, lines after the comment _Unicode line and paragraph separators_). 
But `Q()` calls the function with `$enctype='html'` (which does not remove those characters since they are OK in HTML) and `JQ()` calls with `$enctype='js'`.

This fix could be only applied to:

            else if ($col == 'subject') {
                $cont = trim(rcube_mime::decode_header($header->$col, $header->charset));
                if (!$cont) $cont = $RCMAIL->gettext('nosubject');
                $cont = rcube::Q($cont);
            }

But i think it should also be applied to the folder and column names (the next 2 calls to rcube::Q()).

alecpl:
I've found where the issue comes from. "Escaped U+2028 and U+2029 when JSON_UNESCAPED_UNICODE is supplied as json_encode options and added JSON_UNESCAPED_LINE_TERMINATORS to restore the previous behaviour." in release notes for PHP 7.1.0. Which means there's no issue in PHP >= 7.1. I guess we can just disable the option on older versions in rcube_output::json_serialize(). As I see it's been added in git-master version only.

alecpl:
Fixed with 1556eb0 .

dsoares:
Works for me. Thanx.

PR #6190, opened 2018-02-20
Set date_default_timezone_set of DateTime object

lucasmarin:
I propose this update to guarantee that the out value on date php function of timestamp created on beginning are same of datetime argument object on function when the caller of this function set to not convert to timezone of config file. 

This situation occur on calendar plugin when I'm on GMT -3 and I receive a invite of all day event. 

captura de tela de 2018-02-20 14-48-01

But how the datetime object is converted to unix timestamp format to after be presented using the date php function, I see that this situation may occur not only of this way above.

The ics files received on gmail invite:
[icss.zip](https://github.com/roundcube/roundcubemail/files/1741247/icss.zip)

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -1201,8 +1201,11 @@ class rcmail extends rcube
      */
     public function format_date($date, $format = null, $convert = true)
     {
+        $stz = date_default_timezone_get();
+
         if (is_object($date) && is_a($date, 'DateTime')) {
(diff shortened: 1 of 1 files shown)


alecpl:
This change does not fix the calendar plugin issue and I'm not sure it fixes anything. I can agree that we should probably not convert DateTime into timestamp as it's redundant (but we need to support strftime formats).

PR #6198, opened 2018-02-27
Added exec_hook to send error to an external error tracking tool

AllanKlaus:
I'm working with roundcube and I felt the need of a hook to send a log to my error tracking tool when rouncube handles an exception. So I add a hook to do it in using a plugin.

issue  #6199

### DIFF
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -1303,13 +1303,6 @@ class rcube
      */
     public static function raise_error($arg = array(), $log = false, $terminate = false)
     {
-        // send error to external error tracking tool
-        $arg = self::$instance->plugins->exec_hook('raise_error', $arg);
-
-        if ($arg['abort']) {
-          return;
-        }
-
         // handle PHP exceptions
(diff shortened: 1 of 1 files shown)


alecpl:
The hook should be named just `raise_error`. I guess we could also extend that a little bit so plugins can modify $args or abort the execution.

AllanKlaus:
@alecpl nice suggestions, I changed it.

AllanKlaus:
Can I do something else?

alecpl:
I think it would make sense to move that code later in this function, i.e. after the $arg input has been unified. So, plugin author does not have to do all that conversion that is in the start of the raise_error() method. Also, I guess we could add `$arg['cli'] = $cli; $arg['log'] = $log; $arg['terminate'] = $terminate;` before.Please, use 4-spaces indent.

AllanKlaus:
@alecpl nice.

I have done:
- Moved hook to later in function
- Added `$arg['cli']; $arg['log']; $arg['terminate'];`
- Removed `$arg['abort']` because, later in function `$arg` is overwrited.

PR #6218, opened 2018-03-18
Fix typo in en_GB translation file

karolaug:
There is no closing quotation mark after "Deleted Items" in en_GB translation, which is visible in menu entry.


alecpl:
Please, use https://www.transifex.com/roundcube/roundcube-webmail/ for localization updates. I applied the change for you right now.

Issue #6219, opened 2018-03-21
no hook 'config_get' with plugins when plugins came from default.inc.php


motom001:
If you config your plugins in default.inc.php and not in config.inc.php so the hook 'config_get' will never be fired for the option $config['plugins']
At a second point, I think the rcube_config::all should reference to rcube_config::get - a possible pull request will follow in the next minutes

alecpl:
default.inc.php should be never modified by users. We do not support that.

motom001:
Okay, I understand this point of view.
I'am a normal user and I don't know, where I can find a complete list of options. So I looked in default.inc.php and founded a beautiful commented option-file.
When I had found the special part for my wishes, I wouldn't copy this to a new file (config.inc.php). Instead I modify it in the opened file 'defaults.inc.php'
I also founded not onw part, with a note, that I don't have to modify this file.

So maybe this are more then one parts:
1. modify the default.inc.php and add a big note at the beginning of the file
2. check the filesize of default.inc.php if it modified (maybe create checksum at installation?)
3. use rcube_config::get from rcube_config::all instead copy and paste source code from one methode to the next - they do booth the same, I think, don't tey? -> see pullrequest #6220

alecpl:
Maybe we should move `defaults.inc.php` to `program/include/defaults.php` and copy the contents to wiki for documentation purposes. I'm not sure if just a warning in the file will be good enough. @thomascube what do you think?

thomascube:
@alecpl I agree that might be a good move.

motom001:
In my point of view there are three topics:
1. hide defaults.inc.php to normal user (like suggestion from @alecpl ) to avoid changes
2. different results for "get()" and "all()"
3. performance problems if a developer have to iterate the complete array, when hook with * were called

The second topic is the most important. Is that okay when two similar methods return different results? Who should understand this behavior in two or three years?

alecpl:
The all() method is kind of special, it returns initial/non-converted values (e.g. for timezone), which is sometimes useful. We can describe the method better, but I don't really see another way to improve this or remove this distinction.

motom001:
Maybe a additional parameter (bool) 'getRaw = false' for the get methode, which could called from the all method and from each developer who need also the raw content?
Do you need an example for this from me?

PR #6220, opened 2018-03-21
fix issue #6219

motom001:
I think the rcube_config::all should reference to rcube_config::get
This is also interessting for the single-hooks of config

### DIFF
--- a/program/lib/Roundcube/rcube_config.php
+++ b/program/lib/Roundcube/rcube_config.php
@@ -469,10 +469,10 @@ class rcube_config
      */
     public function all()
     {
-        $props = $this->prop;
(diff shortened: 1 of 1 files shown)


alecpl:
We didn't do this before for performance reasons. I'm afraid `config_get` hook might be slow, and here we'd call it many times. That's why there's another `config_get` hook call in all() method (with special 'name' argument value). If we decide to do that, we should remove that special hook use.

Oh, and there's another issue. There are two cases in get() where we convert stored config into something different (timezone, client_mimetypes). That converted value might be unexpected in the result of all() method. And that might be even bigger issue than performance.

PR #6226, opened 2018-03-24
ensure tooltip is only set on toolbar icon

johndoh:
I'm not sure if its better to fix this in the contextmenu plugin or here.

Elastic adds a tooltip to the reply all icon on the mail toolbar to indicate which type of reply all mode is active. This change ensures the tooltip is only set on the icon in the mail toolbar and not on any other elements in the DOM with the same classes (eg in the contextmenu). The reason is having only 1 item in the contextmenu with a tooltip looks odd imo.


johndoh:
On closer inspection fixing it in a plugin means listening for the enable-command event and hoping they are executed in the correct order. So I think a change in the core is more reliable.

johndoh:
fixed in plugin

PR #6263, opened 2018-04-25
Improve installto.sh (follow-up to PR #6261)

tessus:
I'm sorry, but I missed 2 things in the first PR. I thought I had considered and tested all eventualities, but I did not.

1) The files .htaccess and .user.ini were not copied to the target, if
they did not exist on the target
2) If the contents `file` and `file.new` were the same, `file.new` was not removed.

This commit takes care of these 2 cases.

### DIFF
--- a/bin/installto.sh
+++ b/bin/installto.sh
@@ -88,6 +88,9 @@ if (strtolower($input) == 'y') {
       if (file_get_contents("$target_dir/$file.new") != file_get_contents("$target_dir/$file")) {
         echo "\n!! New $file file saved as $file.new !!";
       }
+      else {
+        @unlink("$target_dir/$file.new");
+      }
     }
   }
 


bin/installto.sh
```diff
@@ -71,9 +71,17 @@ if (strtolower($input) == 'y') {
...
-      $copied[$file] = copy(INSTALL_PATH . $file, "$target_dir/$file.new");
-      if ($copied[$file]) {
-        echo "$file\n";
+      if (file_exists("$target_dir/$file")) {
+        $copied[$file] = copy(INSTALL_PATH . $file, "$target_dir/$file.new");
+        if ($copied[$file]) {
+          echo "$file\n";
+        }
+      }
+      else {
```
alecpl: Adding such a file may break the installation. So, I'd not copy the file if it not exists on target.

alecpl: or we add .new extension so it is not activated.

alecpl (changes requested):

bin/installto.sh
tessus: ok, now I am totally confused. that's the `.htaccess` file that comes with roundcube. it has RewriteRules and usually it's the other way round. a lot of web apps that use mod_rewrite stop working after `.htaccess` is removed.

alecpl:
If user removed the file it means he don't need it and configured Roundcube in another way (e.g. in Apache config). The file contains only some optimizations and security features. Nothing is required there.

tessus:
Right, but if a user configures the stuff in Apache they also usually set `AllowOveride None` thus .htaccess is ignored anyways. But sometimes even an Override makes sense and can be very useful.
It all depends.

Anyways, I like the idea of using the .new extension. In that case it is not activated and the user can still check, if something has changed (that he might want to add to their Apache config).

If you are ok with it, I change the code to copy the file as .new.

alecpl:
Yes. I'm ok with it.

tessus:
Ok, I've reverted the first part.

Now a .new file is created again, if it doesn't exist on the target.

Issue #6272, opened 2018-04-29
extend skin_logo to allow different images for differnt skins


johndoh:
Different skins can require different logos (colours/sizes/backgrounds) please consider extending the `skin_logo` config option to allow the use of different logos in different skins.

alecpl:
Yup. How about adding skin prefix to the skin_logo array keys? E.g.:
```
array(
    "elastic:*" => "/images/logo.png",
    "larry:messageprint" => "/images/logo_print.png"
)
```
Would you create a PR?

johndoh:
Sure. Do you think the order should be: 

- \<skin>:\<template>
- \<template>
- \<skin>:*
- \*

or

- \<skin>:\<template>
- \<skin>:*
- \<template>
- \*

alecpl:
Good question. I think the later will be better.

PR #6273, opened 2018-04-29
extend skin_logo allowing per skin logos

johndoh:
for #6272


program/include/rcmail_output_html.php
```diff
@@ -2251,4 +2243,51 @@ protected function about_content($attrib)
...
+    /**
+     * Get logo URL for current template based on skin_logo config option
+     *
+     * @param string  $template Name of the template to get the logo for
+     *                          default is current template
+     * @param boolean $strict   True if logo should only be returned for specific template
+     *
+     * @return string image URL
+     */
+    protected function get_template_logo($template = null, $strict = false)
```
alecpl: Do you see use cases for $template and $strict arguments? I'd remove that part.

johndoh: Perhaps $template is not needed. I was thinking about plugins overriding default behaviour but may be this does not really make sense.

The idea for $strict came while I was working on #6274 - I thought perhaps the skin_logo config option could also be used to define the logo on the watermark screen but in this case you would not want to fall back to the default logo.

alecpl:
I'm not a big fan of the logo image on printouts. Also, as you might not notice all printing features are hidden on phones. So, I'm not sure.

johndoh:
Right, no point in worring about the layout on small screens. I've removed that.

I thought having the logo on the print screen might be important for branding, also (less important) it makes Elastic consistent with Classic and Larry. So how about making it optional? There will not be one by default but if a print logo is configured in `skin_logo` it will be displayed?

alecpl:
We really should not complicate things with a different logo image for different screen sizes. We can use the same image and just scale it in css. I'd remove that part.

The watermark part has some performance implications, I'm not sure I like to have that in core.

johndoh:
On regular screens the space available for the logo is very limited. Not really enough for a full size logo. But on phones there is more spage, so why not give the option to fill that space?

rclogos

It could be an optional thing (via the skin_logo config) and not used by default?

The watermark stuff. Yes ok, I think that got a bit over complicated so I'll remove it. At the moment there are 6 different template objects used to create the watermakr iframe. If I work on unifying that into 1 object then a plugin would only need to work with 1 object and it could change the URL of the default watermark page if required with no need for watermark.html to be in the template system.

alecpl:
I also think that that part about favicon.html is too much.

Issue #5025, opened 2011-03-03
A config option for custom watermark


rcubetrac:
_Reported by brandond on 3 Mar 2011 02:16 UTC as Trac ticket #1487816_

A recent change to trunk added a new config option for custom logo. It would be nice if there was a similar option for the watermark image. Having both of these made configurable would remove almost all of my local skin customizations, and let me go back to the stock files.

_Migrated-From: http://trac.roundcube.net/ticket/1487816_

ad221282:
I can change the watermark to my logo, open the root file where roundcube is hosted, C:/roundcube/skins/larry and classic/images put your logo .gif, then in the larry/classic folder there is a watermar.html, open it with dreamweaver and replace the path of the old watermark for the new one, save it and ready!
PS. dont forget to clear your browsing data in order to check if it works or not!

thomascube:
There's a plugin for this: https://plugins.roundcube.net/packages/roundcube/customizr

tomsommer:
At least allow us to modify it using CSS/LESS? There is no CSS file imported in `watermark.html`

alecpl:
There's undocumented `blankpage_url` (default: `/watermark.html`). I guess we should document that and see if that's it.

Extending `skin_logo` to work on watermark page will be much more complicated.

gianlucagiacometti:
I used a tmp file if it may be useful:
https://github.com/gianlucagiacometti/roundcube-toolbox

alecpl:
Option documented with
```
// Location of the blank (watermark) frame page. By default it is the watermark.html
// file from the currently selected skin. Prepend name/path with a slash to use
// current skin folder. Remove the slash to point to a file in the Roundcube
// root directory. It can be also a full URL.
$config['blankpage_url'] = '/watermark.html';
```

PR #6274, opened 2018-04-29
initial work to bring watermark screen into template system

johndoh:
for #5025.

initiaal work to move watermark screen from watermark.html into the roundcube template system with the eventual goal of improving extensibility.


johndoh:
merged into #6273

PR #6275, opened 2018-04-30
create favicon template object

johndoh:
for #5025.

By turning the favicon into a template object will become easier for plugins to workwith. For example to add App Icons to roundcube. Also by using the config in meta.json it means the favicon path is only stored in one place for the older skins making changes easier.


johndoh:
merged into #6273

PR #6277, opened 2018-05-01
Added missing method scope

ionux:
and minor test code cleanup/formatting.

### DIFF
--- a/tests/Framework/Enriched.php
+++ b/tests/Framework/Enriched.php
@@ -7,11 +7,10 @@
  */
 class Framework_Enriched extends PHPUnit_Framework_TestCase
 {
-
     /**
      * Class constructor
      */
-    function test_class()
+    public function test_class()
     {
         $object = new rcube_enriched();
 
@@ -21,7 +20,7 @@ class Framework_Enriched extends PHPUnit_Framework_TestCase
     /**
      * Test to_html()
      */
-    function test_to_html()
+    public function test_to_html()
     {
         $enriched = '<bold><italic>the-text</italic></bold>';
         $expected = '<b><i>the-text</i></b>';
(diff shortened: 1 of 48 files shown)


tests/Framework/Utils.php
```diff
@@ -528,16 +532,15 @@ function test_words_match()
...
     {
         if (strtoupper(substr(PHP_OS, 0, 3)) == 'WIN') {
             $test = array(
                 '' => false,
                 "C:\\" => true,
                 'some/path' => false,
             );
-        }
-        else {
+        } else {
```
alecpl: Here at Roundcube we're not fans of this CS, we prefer else in a new line.

tests/bootstrap.php
```diff
@@ -19,23 +19,27 @@
...
 
-define('TESTS_DIR', __DIR__ . '/');
+define('TESTS_DIR', __DIR__ . PATH_SEPARATOR);
 
 if (@is_dir(TESTS_DIR . 'config')) {
     define('RCUBE_CONFIG_DIR', TESTS_DIR . 'config');
 }
 
-require_once(INSTALL_PATH . 'program/include/iniset.php');
+require_once(INSTALL_PATH . 'program' . PATH_SEPARATOR . 'include' . PATH_SEPARATOR . 'iniset.php');
```
alecpl: It should be DIRECTORY_SEPARATOR, but anyway I'd prefer to keep that as it was.

tests/Framework/Enriched.php
```diff
@@ -33,30 +32,30 @@ function test_to_html()
...
-            array('<flushleft>', '<span style="text-align: left">'),
-            array('</flushleft>', '</span>'),
-            array('<flushright>', '<span style="text-align: right">'),
-            array('</flushright>', '</span>'),
-            array('<flushboth>', '<span style="text-align: justified">'),
-            array('</flushboth>', '</span>'),
-            array('<indent>', '<span style="padding-left: 20px">'),
-            array('</indent>', '</span>'),
-            array('<indentright>', '<span style="padding-right: 20px">'),
+            array('<bold>',         '<b>'),
```
alecpl: Such spacing does not really make sense to me.

alecpl (changes requested):

PR #6280, opened 2018-05-04
Parse all quotas from GETQUOTAROOT

ledgr:
RoundCube was parsing only the first quota triplet for every Quota Root.
This PR fixes it and in result displays all the enabled quotas.

screen shot 2018-05-04 at 15 29 29

### DIFF
--- a/program/lib/Roundcube/rcube_imap_generic.php
+++ b/program/lib/Roundcube/rcube_imap_generic.php
@@ -3130,8 +3130,9 @@ class rcube_imap_generic
                 if (preg_match('/^\* QUOTA /', $line)) {
                     list(, , $quota_root) = $this->tokenizeResponse($line, 3);
 
-                    while ($line) {
-                        list($type, $used, $total) = $this->tokenizeResponse($line, 1);
+                    $quotas = $this->tokenizeResponse($line, 1);
+                    foreach (array_chunk($quotas, 3) as $quota) {
+                        list($type, $used, $total) = $quota;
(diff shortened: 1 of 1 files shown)


program/lib/Roundcube/rcube_imap_generic.php
```diff
@@ -3130,8 +3130,9 @@ public function getQuota($mailbox = null)
                 if (preg_match('/^\* QUOTA /', $line)) {
                     list(, , $quota_root) = $this->tokenizeResponse($line, 3);
 
-                    while ($line) {
-                        list($type, $used, $total) = $this->tokenizeResponse($line, 1);
+                    $quotas = $this->tokenizeResponse($line, 1);
+                    foreach (array_chunk($quotas, 3) as $quota) {
```
aspettl: I get the following PHP warning with the current release (1.3.7) in this line:
```
[04-Aug-2018 15:38:30 UTC] PHP Warning:  array_chunk() expects parameter 1 to be array, null given in /var/www/html/program/lib/Roundcube/rcube_imap_generic.php on line 3129
[04-Aug-2018 15:38:30 UTC] PHP Warning:  Invalid argument supplied for foreach() in /var/www/html/program/lib/Roundcube/rcube_imap_generic.php on line 3129
```
My IMAP server (Courier 4.17.2) is not configured for quotas, probably that is the reason. I don't know anything about the IMAP protocol itself, but probably we should do nothing in case of `$quotas === NULL`?

PR #6285, opened 2018-05-08
unify content frame (watermark) template objects

johndoh:
There are currently 6 different template objects used in the core + those in plugins to create the content frame with the watermark screen in. The idea behind this PR is to replace these similar objects with a single object all templates can use.


johndoh:
Good point thanks. Also, after thinking more I think its better to check for valid variable expressions rather than just `:` so I'm thinking a regex like `/^(session|config|env|request|cookie):([a-z0-9_]+)$/i`. config and env are clearly required but what do you think about session, request and cookie? I guess being able to store a URL in the session might be useful somewhere but request and cookie?

alecpl:
Support `config` and `env` only, please.

PR #6286, opened 2018-05-11
Show percentage of the highest used quota in the summary

ledgr:
Users get surprised when summary shows 0%, but clicking details reveal that there is another quota that is already 100% used.

#### before 
screen shot 2018-05-04 at 15 29 29

#### after
screen shot 2018-05-11 at 10 27 11


alecpl:
1. It looks to me like you compare apples with oranges (the $free checks). Probably comparing only percent value would be enough.
2. Some code expects storage quota values, e.g. if you have only one messages count quota and no other quota the tooltip will most likely display invalid information. I'm not sure we should care about such a corner case. Probably not.

PR #6320, opened 2018-06-08
Fix spelling of occurred

pepsiman:


alecpl:
I fixed this via Transifex service (https://www.transifex.com/roundcube/roundcube-webmail) now. Please use the service for further translation fixes.

pepsiman:
This isn't a translation fix - the error is in the original language.

alecpl:
The original language is en_US.

pepsiman:
Yes, occurred has 2 rs in en_US.

alecpl:
Then provide a PR for en_US. All the rest is via Transifex. Thanks.

PR #6361, opened 2018-07-25
modoboa driver for password plugin

stefaweb:
This driver for the modoboa mail system (https://github.com/modoboa/modoboa) can be used with the  Roundcube password plugin. It use the modoboa API and token to communicate.

### DIFF
--- a/plugins/password/drivers/modoboa.php
+++ b/plugins/password/drivers/modoboa.php
@@ -7,7 +7,7 @@
  * Return value is a json string saying result: true if success.
  *
  * @version 1.0.1
- * @author stephane @actionweb
+ * @author stephane @actionweb.fr
  *
  * Copyright (C) 2018, The Roundcube Dev Team
  *
@@ -26,7 +26,7 @@
  *
  * The driver need modoboa core 1.10.6 or later 
  *
- * You need to define in plugin/password/config.inc.php theses variables:
+ * You need to define theses variables in plugin/password/config.inc.php
  *
  * $config['password_driver'] = 'modoboa'; // use modoboa as driver
  * $config['token_api_modoboa'] = ''; // put token number from Modoboa server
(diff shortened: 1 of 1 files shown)


plugins/password/drivers/modoboa.php
```diff
@@ -0,0 +1,124 @@
...
+        $rcmail = rcmail::get_instance();    
+        $ModoboaToken = $rcmail->config->get('token_api_modoboa');
+
+        $RoudCubeUsername = $_SESSION['username'];
+        $IMAPhost = $_SESSION['imap_host'];
+
+        // Write variables in log
+        rcube::write_log('errors', "ModoboaToken: " . $ModoboaToken);
+        rcube::write_log('errors', "IMAP host: " . $IMAPhost);
+        rcube::write_log('errors', "RoudCubeUsername: " . $RoudCubeUsername);
```
alecpl: Please, remove/comment the code that writes debug info to the error log.

plugins/password/drivers/modoboa.php
```diff
@@ -0,0 +1,124 @@
...
+
+        // Call GET to fetch values from modoboa server
+        $curl = curl_init();
+
+        curl_setopt_array($curl, array(
+          CURLOPT_URL => "https://" . $IMAPhost . "/api/v1/accounts/?search=" . $RoudCubeUsername,
```
alecpl: I think you should urlencode() the username

plugins/password/drivers/modoboa.php
```diff
@@ -0,0 +1,124 @@
...
+        ));
+
+        $response = curl_exec($curl);
+        $err = curl_error($curl);
+
+        curl_close($curl);
+
+        if ($err) {
+          return PASSWORD_CONNECT_ERROR;
+          rcube::write_log('errors', "Modoboa cURL Error #: " . $err);
```
alecpl: Code after return is not executed ;) Also it's better to use rcube::raise_error().

plugins/password/drivers/modoboa.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        // Decode json string
+        $decoded = json_decode($response);
+
+        // Get user ID (pk)
+        $userid = $decoded[0]->pk;
```
alecpl: It would be good to check if $decoded is an array first.

plugins/password/drivers/modoboa.php
```diff
@@ -0,0 +1,124 @@
...
+        ));
+
+        $response = curl_exec($curl);
+        $err = curl_error($curl);
+
+        curl_close($curl);
+
+        if ($err) {
+          return PASSWORD_CONNECT_ERROR;
+          rcube::write_log('errors', "Modoboa cURL Error #: " . $err);
```
alecpl: Code after return...

alecpl (changes requested):

alecpl (approved):

PR #6369, opened 2018-08-01
Privacy: Remove roundcube version string in header

Rotzbua:
Avoid unnecessary information leakage.

Less metadata for NSA & Friends.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -688,7 +688,7 @@ $config['max_group_members'] = 0;
 $config['product_name'] = 'Roundcube Webmail';
 
 // Add this user-agent to message headers when sending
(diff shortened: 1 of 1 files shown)


alecpl:
I have no strong opinion, but maybe we should not set the header at all by default?

Rotzbua:
I think for backwards compatibility it would be better to keep a identifying header. The scenario could be that roundcube produces false-positives at the spam filter and so a extra rule depending on the header can avoid the false-positives. I do not work with mail servers and spam filters so I do not know how realistic this scenario is...

Rotzbua:
@alecpl Rebased the pr and solved the conflict. Have you made a decision?

Rotzbua:
Obsolete by 6b8881271f271422c259a33d842fcd1979baa4c8

PR #6384, opened 2018-08-06
Set first thread UID as parent if no parent exists

HLeithner:
A regex is used to normalize the THREAD output from the IMAP Server. The regex replaces parent less threads with the first uid (mostly the oldest message) as parent. Also fixed the documentation of parse_thread.


HLeithner:
Any opinion on this?

omehegan:
I just manually tested this change against 1.3.8, and it solves the problem I wanted solved, which is that emails from GitHub do not thread in RC (even though they do in every other client I use).

sjorge:
I just tested this against the 1.4-rc1 and this works fine for it as wel.

HLeithner:
@alecpl it would be great if you can merge this in to 1.4.2 I'm still using and it's working perfectly.

sjorge:
Another +1 for me, works fine and it's a bit annoying that I have to keep reapplying this after every upgrade

alecpl:
Does not look like a complete solution. It will not "fix" threads like: `((1 2)(3))`. So, some more work is needed. It would be also nice to have some unit tests.

alecpl:
Fixed with fded360d84.

HLeithner:
Thanks @alecpl

PR #6396, opened 2018-08-17
Adding support for password expiration

doktoil-makresh:


alecpl:
Thanks, but see the new password driver API.

doktoil-makresh:
Not sure to understand. I just added an optional SQL query to handle password expiry via an extra column.
And to be honest my PHP skills and my free time are not allowing me to go further

PR #6414, opened 2018-08-28
managesieve krb_authentication load_config fix

urusha:
Without this we have to specify krb_* configuration in managesieve config.inc.php


urusha:
For some reason build failed for 7.1. Seems to be temporarily failure.

PR #6429, opened 2018-09-04
Hungarian translation of $labels['logout'] is too long.

dunstn:
It doesn't display correctly under the Elastic skin.


alecpl:
Please, use Transifex service. https://www.transifex.com/roundcube/roundcube-webmail/

PR #6446, opened 2018-09-12
for reply to all, also skip VERP style Sender: address

futatuki:
Some mail lists uses VERP style E-mail address for Sender: header, like
`Sender: "List name" <listaname-bounce+foo=example.com@bar.example.org>`
to analyze bounce mail.

I'm grad if roundcubemail ommit those VERP style E-mail address for Cc:, when we uses "reply to all".

So I've added regex pattern to match them.
(Acutually, I didn't  examine this pattern in master branch, but I examined in release-1.3 branch, program/steps/mail/compose.inc)

### DIFF
--- a/program/include/rcmail_sendmail.php
+++ b/program/include/rcmail_sendmail.php
@@ -1032,7 +1032,7 @@ class rcmail_sendmail
                 if ($v = $message->headers->get('Sender', false)) {
                     // Skip common mailing lists addresses: *-bounces@ and *-request@ (#1490452)
                     if (empty($message->headers->others['list-post'])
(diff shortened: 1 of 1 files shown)


futatuki:
I also think the condition to add 'Sender:' address to cc is different from that intended.
Is it intended to skip simply if the sender of the message is not a mail list, right?
If so, the operator '||' of the diff line should be '&&'.

alecpl:
Since 0048be10a9a7 we ignore the header completely.

futatuki:
Thank you, now we can use "reply for all" safely.

PR #6449, opened 2018-09-14
Fix support for "x_frame_options" config option

faf:
The only way to embed Roundcube Webmail into iframe is to add x_frame_options into configuration:

`$config['x_frame_options'] = 'allow-from https://example.com/';`

But at the moment the code enforces the value of `sameorigin` when viewing emails regardless of configuration, while should behave that way only in case of restrictive setting.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -588,7 +588,7 @@ EOF;
         // allow (legal) iframe content to be loaded
         $iframe = $this->framed || $this->env['framed'];
         if (!headers_sent() && $iframe && ($xopt = $this->app->config->get('x_frame_options', 'sameorigin'))) {
-            if (strtolower($xopt) != 'sameorigin') {
+            if (strtolower($xopt) === 'deny') {
                 header('X-Frame-Options: sameorigin', true);
             }
         }


alecpl:
And you say it's working? `sameorigin` for an iframe inside should work.
I do not agree with this change, will have to do some tests. What browsers did you test?

faf:
I've encountered the invalid behaviour while tried to embed Roundcube into Nextcloud.

Before the change it was impossible to open emails both in (Chrome-based) Chromium 61.0.3163.100 and in (Gecko-based) SeaMonkey 2.48 (it's the current versions for my distro). I've tested in both, since Chrome-based browsers [doesn't support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#Browser_compatibility) `allow-from` value of `X-Frame-Options`, while Gecko-based does.

In the js-console there were the same errors telling that iframe was blocked due to `X-Frame-Options` was set to `sameorigin` (and response headers showed that). After the change the problem has gone.

Moreover, I've made some quick search and found that someone has already encountered and fixed this issue, but didn't provided the fix to your community. Take a look at [this code](https://github.com/Telaxus/EPESI/blob/master/modules/CRM/Roundcube/RC/program/include/rcmail_output_html.php#L521).

If you want next week I can make more tests on PC with Windows and latests stable versions of Chrome, Firefox, Opera, and even IE :), but I'm pretty sure the results will be the same.

alecpl:
I did some tests and indeed it does not work as I expected. The change is correct. Thanks.

PR #6453, opened 2018-09-17
Fix: Keep children of object tag

fjl5:
The HTML tag `<object>` optionally has embedded (child) tags that serve as an alternative (fallback) HTML representation for the object. Of course, the object and its parameters are considered harmful in HTML mail, but the alternative representation is meant for exactly this kind of situation and should be used if `<object>` is either unknown or filtered.

- By ignoring `<object>` tags, roundcube also removes all their child nodes.
- As `<object>` is not in the list of allowed `$html_elements` and `<param>` gets cleaned through `$void_elements`, they get ignored anyway, without removing the valuable child nodes.

This PR simply removes `object` from `$html_elements` and adds a test case:
```html
<div>
<object data="move.swf" type="application/x-shockwave-flash">
<param name="foo" value="bar">
<p>This alternative text should survive</p>
</object>
</div>
```
Here, the `<p>` tag should be displayed instead of the `<object>` tag.

In real world, Apple-Mail's Cocoa HTML Writer seems to create this kind of HTML when pasting e.g. tables. Thunderbird uses the fallback HTML code to display the contents, while Roundcube used to hide the whole block.


alecpl:
Please, do not mix many features in a single PR. I don't like the Print button change.

fjl5:
Uff, the second change was definitely not intended to be part of this PR. Sorry for the trouble. Should I try to revert the "print button" change for this PR, or would just cherry-pick the intended change?.

fjl5:
I've reverted the change that creped in so it is no longer part of this PR. Should be clean now.

alecpl:
Could you resolve conflicts?

PR #6454, opened 2018-09-17
Show print button in message list toolbar

fjl5:
Some users seem to miss the "print" feature in the mailbox view as the button is hidden in the "more..." submenue. This PR moves it up to the toolbar, just like when showing a single mail.


alecpl:
Printing is a rare action and there's already too many buttons, especially when you enable some plugins. So, wontfix.

PR #6455, opened 2018-09-17
Optionally prepend header info to replies

fjl5:
When replying to message, some users expect the mail body to contain fields of the original mail's headers - just like when forwarding a mail. Disabled by default.

Current body looks like this:
```
Am 2018-07-09 13:27, schrieb Herr Absender:
> Das ist der Mail-Body
```

This changes the reply body to;
```
Am 2018-07-09 13:27, schrieb Herr Absender:
> Betreff: Thema
> Von: Herr Absender <h.absender@coreboso.de>
> An: Achim Leitner <a.leitner@coreboso.de>
> 
> Das ist der Mail-Body
```
(sorry for german text sample)

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -1177,6 +1177,10 @@ $config['autoexpand_threads'] = 0;
 // 2  - place cursor above original message (top posting), but do not indent the quote
 $config['reply_mode'] = 0;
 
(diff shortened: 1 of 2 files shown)


alecpl:
I don't like this. With e4832118a40c9f you should be able to implement this as a plugin.

PR #6471, opened 2018-09-28
krb_authentication fixes

urusha:


plugins/krb_authentication/krb_authentication.php
```diff
@@ -146,6 +146,6 @@ private function gssapi_context($protocol)
              $context = $context[$protocol];
         }
 
-        return $context ?: 'host.fqdn@REALM.NAME';
+        return $context;
```
alecpl: Maybe we should log an error if context is not setup? It might be hard to debug what's going on if user did a config mistake and we do not provide a default.

Issue #6385, opened 2018-08-06
Privacy: Avoid referrer leaking


Rotzbua:
## Motivation
Avoid unnecessary information leaking about user.

Example:
Embedded image is loaded. The referrer to external resource is:
Referer: https://mail.example.com/?_task=mail&_caps=pdf%3D1%2Cflash%3D0%2Ctiff%3D0%2Cwebp%3D0&_uid=1445&_mbox=INBOX&_safe=1&_action=show

Leaked information:
- size of the inbox: _uid=1445
- name of the folder: _mbox=INBOX
- capabilities of the browser: _caps=pdf...

## Browser support?
At the moment the browser support is good: https://caniuse.com/#feat=referrer-policy .

## Backwards compatible?
Yes, older browser ignore the policy.

## Implementation status?
There are two old and pending pull requests: #5422 #5674  #6038

## Problem with the pull requests
There are different ways to implement referrer-policy:
 - meta-html-tag
 - http-header sent by php
 - http-header sent by .htaccess (would only work on apache)

There is a missing statement from maintainers which way should be chosen and what should be done to get them merged.

## Further reading

Specification:
https://www.w3.org/TR/referrer-policy/

Old Specification:
https://wiki.whatwg.org/wiki/Meta_referrer

Blog article:
https://scotthelme.co.uk/a-new-security-header-referrer-policy/

edit: add: .htaccess pr, doc

alecpl:
If you review samples in .htaccess file, you'll see it can be done on a different level. No need for a configuration option. I'm not saying we'll not add such an option, but I'd like to see it more robust, see my comment https://github.com/roundcube/roundcubemail/pull/5422#issuecomment-250823993

Rotzbua:
@alecpl After a lot of research and testing, I got following conclusion:

### Compatibility table

| Referrer-Policy   | meta-html-tag | http-header | Compatible old draft  | Works with RC referer_check | Leak to external | Full privacy |
|-------------------|:-------------:|:-----------:|:---------------------:|:---------------------------:|:----------------:|:------------:|
| None setting      |       -       |      -      |           -           |              -              |         x        |       -      |
| Same-Origin       |       -       |      x      |           -           |              x              |         -        |   partially  |
| No-Referrer       |       -       |      x      |           -           |              -              |         -        |       x      |
| Origin            |       x       |      -      |           x           |              x              |     partially    |       -      |
[...]

PR #6477, opened 2018-10-04
Add referrer policy header to disable referrer

Rotzbua:
Type | feature
------|--------
Fixes | #6385

header works on every skin, if we use meta-tag, every skin has to be changed.


alecpl:
Implemented in 186f21c4c.

Issue #6473, opened 2018-10-02
password plugin, option to ask for current password without verifying against $_SESSION['password']


conathan:
We have a token based login system which does not have access to the original password ($_SESSION['password'] is not the customer's current password), but we still want to allow users to change their password (and verify the original password, just not by roundcube).

To do so, we disabled the password check in plugins/password/password.php

else if (false && $conpwd != $newpwd) {
  $rcmail->output->command('display_message', $this->gettext('passwordinconsistency'), 'error');
}

and instead send both the current password provided, as well as the new password via the api.

There is password_confirm_current, but that hides asking for the current password (which we need)

alecpl:
I guess you rather disabled this check:
```
    else if ($confirm && $sespwd != $curpwd) {
        $rcmail->output->command('display_message', $this->gettext('passwordincorrect'), 'error');
    }
```
I think we could add such an option (simple), it could also be controlled by the driver's code (more work). Would you mind providing a pull request?

conathan:
Looks better then the one I was going to work on johndoe

alecpl:
Merged.

Issue #5040, opened 2011-12-20
Password Plugin - more password strength checking options


rcubetrac:
_Reported by fraggy on 20 Dec 2011 10:49 UTC as Trac ticket #1488268_

Hi,

we are using the password plugin to change passwords in LDAP.
It would be nice, if there are options to be more strict when choosing a new password.

required:
- Password Minimum Length: 8 chars (password_minimum_length allready implemented)
- Contain characters from 3 of the following 4 categories: English uppercase characters (A through Z), English lowercase characters (a through z), Base 10 digits (0 through 9), Non-alphabetic characters (for example, !, $, #, %)

nice to have:
- min. 2 different chars between old and new password (pam_cracklib Option difok=2)
- Not contain the user's account name or parts of the user's full name that exceed two consecutive characters

regards,
Maik

_Migrated-From: http://trac.roundcube.net/ticket/1488268_

crazy4chrissi:
I think it would be perfect if you could integrate zxcvbn, which is the password strength estimation developed at Dropbox:
https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/
https://github.com/dropbox/zxcvbn
https://github.com/bjeavons/zxcvbn-php

It allows much better password strength checks than these simple rules. And it allows to show the user a rank of his password strength as he types.

razyr:
I'd like to add my voice to the requests for ZXCVBN integration.  It does a far better job than simple rules such as "must contain at least one special character" and is relatively easy to integrate into existing portals or applications.   As crazy4chrissi mentions, it provides the user direct password strength feedback as they are entering their password and it has a nice clean UI.

![2016-12-06_09-31-54](https://cloud.githubusercontent.com/assets/8116645/20930554/d7c27218-bb9b-11e6-9d5f-58b2a9579bab.png)

torbjo:
Just a side note: When using ldap I think its more sane to implement
password policy checking on the (ldap) server.

With OpenLDAP use the ppolicy module/overlay:
* <http://www.openldap.org/doc/admin24/overlays.html>
* <http://www.zytrax.com/books/ldap/ch6/ppolicy.html>

But for the other password backends it's a nice feature to have.

alecpl:
An option could be also HaveIBeenPwned.com API use: https://haveibeenpwned.com/API/v2#SearchingPwnedPasswordsByRange

crazy4chrissi:
[...]

PR #6479, opened 2018-10-07
give password plugin drivers more power

johndoh:
Allow password drivers more control over the entire password changeing processes not just the save action. Allow them to perform old/new password comparisons and also password strength checking.

*) allow password drivers override default password comparisons (eg new is not same as current) #6473, #5118
*) allow password drivers override default strength checks (eg allow for 'not the same as last x passwords') #246 
*) allow separate password saving and strength drivers for use of strength checking services eg HaveIBeenPwned.com #5040
*) allow drivers to define password strength rules displayed to the user
*) rename password_require_nonalpha config option to password_check_strength to reflect new strength checking possibilities (added backwards compatibility)


alecpl:
Great work @johndoh. Merged with small improvements

PR #6483, opened 2018-10-16
limit skins available/shown in the settings section

dhoffend:
This change allows the admin to limit/reduce the amount of available skins presented to the user in the settings section.

Currently the part is missing that enforces the skin selection when the "current" skin is not on the list of available skins and fall back to the default one.


alecpl:
Thank you, Daniel. We need this option too.

PR #6485, opened 2018-10-19
[html2text/plaintext] add inline links behavior

nantkowiak:
i added a new feature to display links besides there text instead of a linklist. similar to other mailtools (google, thunderbird, ...) which uses html2text functionality. additionally you can now control the html2text behavior via the config.


homotechsual:
What would be needed to unstick this? It would be a fantastic addition.

program/include/rcmail.php
```diff
@@ -1924,10 +1924,11 @@ public function get_resource_content($name)
     public function html2text($html, $options = [])
     {
         $default_options = [
-            'links'   => true,
-            'width'   => 75,
-            'body'    => $html,
-            'charset' => RCUBE_CHARSET,
+            'links'             => $this->config->get('html2text_links', true),
```
alecpl, line 1927: Maybe instead of two options for link create one. I.e `html2text_links` would accept 3 values: 0 - no links, 1 - links at the end, 2 - links inline. Of course with BC where false = 0, true = 1.

program/include/rcmail.php
```diff
@@ -1924,10 +1924,11 @@ public function get_resource_content($name)
...
+            'width'             => $this->config->get('html2text_width', 75),
+            'body'              => $html,
+            'charset'           => $this->config->get('html2text_charset', RCUBE_CHARSET),
```
alecpl, line 1930: I would not allow to change this charset. Please, remove this option.

program/lib/Roundcube/rcube_html2text.php
```diff
@@ -324,19 +324,29 @@ class rcube_html2text
...
      * will instantiate with that source propagated, all that has
      * to be done it to call get_text().
      *
-     * @param string $source    HTML content
-     * @param bool   $from_file Indicates $source is a file to pull content from
-     * @param bool   $do_links  Indicate whether a table of link URLs is desired
-     * @param int    $width     Maximum width of the formatted text, 0 for no limit
+     * @param string  $source    HTML content
+     * @param boolean $from_file Indicates $source is a file to pull content from
+     * @param boolean $do_links  Indicate whether a table of link URLs is desired
```
alecpl, line 345: So, it would be bool|int. And we should create class constants LINKS_NONE=0, LINKS_END=1, LINKS_INLINE=2

alecpl (changes requested):
Also, it would be good to have some tests.

program/include/rcmail.php
```diff
@@ -1924,10 +1924,10 @@ public function get_resource_content($name)
     public function html2text($html, $options = [])
     {
         $default_options = [
-            'links'   => true,
-            'width'   => 75,
-            'body'    => $html,
-            'charset' => RCUBE_CHARSET,
+            'links'             => $this->config->get('html2text_links', rcube_html2text::LINKS_DEFAULT),
```
alecpl, line 1927: Remove extra spacing before =>.

alecpl (changes requested):

tests/Framework/Html2text.php
```diff
@@ -79,7 +79,7 @@ function data_html2text()
      */
     function test_html2text($title, $in, $out)
     {
-        $ht = new rcube_html2text(null, false, false);
+        $ht = new rcube_html2text(null, false, rcube_html2text::LINKS_NONE);
```
alecpl, line 82: You could leave `false` in a few places to have the BC check.

config/config.inc.php.sample
```diff
@@ -86,3 +86,14 @@ $config['plugins'] = [
...
+// Html2Text Link Behavior
+// options:
+// 0 - if links should be removed
+// 1 - if a table of link URLs should be listed after the text
+// 2 - if the link should be displayed to the original point in the text they appeared
+$config['html2text_links'] = 1;
+
+// Html2Text Text Width
+// Maximum width of the formatted text, in columns
+$config['html2text_width'] = 75;
```
alecpl, line 99: The documentation should go to `defaults.inc.php`.

alecpl (changes requested):

tests/Framework/Html2text.php
nantkowiak: already done. see bc tests `test_links_bc_with_boolean` and `test_links_no_list_bc_with_boolean`

PR #6486, opened 2018-10-19
cleanup login container

linushstge:
Before:
image

After:
image


alecpl:
That `&middot` fix makes sense, but I don't like the rest. Too much code for not much of an improvement.

alecpl:
I applied some improvements in c805f7eb425.

PR #6495, opened 2018-10-26
(Elastic) small compat fix

johndoh:
some JS in elastic is a little broad when it comes to setting classes on buttons. this small fix just makes sure that the core JS only interacts with the toolbar button in the core and not others (eg from plugins)


johndoh:
on seond thought i think this line could be removed completely since ab6b651. @alecpl what do you think?

alecpl:
I think so, yes. See 2d172126e7.

PR #6499, opened 2018-10-27
store menu buttos by menu id

johndoh:
after ab6b651 I am looking at extending the contextmenu plugin to work in the same way. checking through the entire menu_buttons array each time seems wasteful. this small change stores the information by menu id for quick access to check if a specific menu has any active commands.


alecpl:
The same menu can be used by more than one button. I'm not sure this change will cause issues now, but would be good to not have such limitation.

johndoh:
good point. i've added support for multiple buttons now

PR #6502, opened 2018-10-28
add php 7.3/nightly to travis and allow failure

Rotzbua:
php 7.3 is not released yet, so allow failure

But we can see if something fails with new php environment.


alecpl:
I enabled PHP 7.3 in another commit.

PR #6503, opened 2018-10-29
Create autocomplete popup in a parent window when invoked from an iframe

alecpl:
This one via PR because I need a second thought...

I need autocomplete to be created in a parent window when the input element is in an iframe (which might be inside of a dialog), otherwise it might get truncated and hard to work with (especially in Elastic mobile when the frame might be small). The only problem I have with this is assistive technologies. Though, I have no idea it's a big problem, as we already use frames anyway.

Later, I'll probably like to do the same with date-picker and color-picker widgets.

@thomascube any opinions?

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -5796,7 +5796,7 @@ function rcube_webmail()
           return;
 
         var dir = key == 38 ? 1 : 0,
-          highlight = document.getElementById('rcmkSearchItem' + this.ksearch_selected);
+          highlight = this.ksearch_pane.find('li.selected')[0];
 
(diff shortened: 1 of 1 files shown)


thomascube:
Assistive technologies should work across frames if focus is assigned correctly with JS. Especially when the popup is closed, make sure the focus goes back to the input element.

alecpl:
In our autocompletion implementation focus never leaves the input element. When we move over the list we do not focus it, we only set aria-selected and aria-activedescendant attributes. So, I guess we're fine.

PR #6504, opened 2018-10-29
Integrate markasjunk2 features into markasjunk - marking as non-junk + learning engine

alecpl:
This is not finished yet. TODO:
1. Better Non-Junk icon (at least for Elastic).
2. Get rid of `markasjunk_toolbar` option. I don't like two buttons in the toolbar. One reason is not enough room there in Elastic. I propose to use only Junk button in the toolbar. When `markasjunk_spam_only=false` we add "mark as junk" and "mark as non-junk" into Mark menu. Also making it different on the message toolbar does not make much sense, imo.

@johndoh


johndoh:
> Get rid of markasjunk_toolbar option. I don't like two buttons in the toolbar. One reason is not enough room there in Elastic. I propose to use only Junk button in the toolbar. When markasjunk_spam_only=false we add "mark as junk" and "mark as non-junk" into Mark menu. Also making it different on the message toolbar does not make much sense, imo.

Does this mean that with `markasjunk_spam_only=false` then the icon will always be on the toolbar (similar to current markasjunk plugin) and with `markasjunk_spam_only=true` then the icons will always be in the mark menu?

I like having them in the mark menu because it is a mark action. People who are used to markasjunk might be confused by the change though. I agree the different settings for the different screens is a bit much.

alecpl:
No, I think I'm a little bit lost in all these options. What I want is two modes of operation:
1. Default, only moving/marking as junk (what markasjunk plugin did originally): here we need only the "mark as junk" button in toolbar.
2. The default mode plus moving/marking as non-junk: here I propose to keep the toolbar button and additionally add "mark as junk" and "mark as non-junk" into the Mark menu.

I'm not sure we even need `markasjunk_spam_only` option at all. Maybe we could check `markasjunk_ham_mbox` and/or `markasjunk_ham_flag` to achieve these two modes. What do you think?

johndoh:
> I'm not sure we even need markasjunk_spam_only option at all. Maybe we could check markasjunk_ham_mbox and/or markasjunk_ham_flag to achieve these two modes. What do you think?

The `markasjunk_spam_only` was added recently to give markasjunk2 the same behaviour as markasjunk. Removing it would introduce a new limitation because right now the plugin could be configured with `markasjunk_ham_mbox` and `markasjunk_ham_flag` set to false so messages are not moved or flagged but a driver can still perform an action to mark the message as ham (eg the email_learn driver) which sends a copy of the message off to an email address)

johndoh:
seems good to me. just fyi you can still get both the junk and not junk buttons on the toolbar if you do a multi folder search (that is intended in MAJ2).

johndoh:
this might be a little too picky but there are lots of references to "spam" (coz in the UK its commonly called that) rather than "junk". would you like me to change all function/variable names from spam -> junk? just thinking that now, with the rewrite, it would be a good time to clean up the fairly random mix of terms.

alecpl:
I come from a country that uses 'Spam' so I don't have a strong opinion. Besides spam/ham looks better than junk/nonjunk in some places, imo. Also, there is Spamassassin, not Junkassassin ;)

Issue #6511, opened 2018-11-02
web.config gets deleted in temp folder


Virinum:
Hi!
I'm running roundcube on IIS. For disabling access to the temp folder I need to create a web.config in this folder. But it gets deleted after `temp_dir_ttl`.

I've two ideas:

1. Add a config item for setting filenames which get ignored.
2. Add common settingfilenames [here](https://github.com/roundcube/roundcubemail/blob/5542e91b21ba6267202f89ef9464f7ef8ce93dd7/program/lib/Roundcube/rcube.php#L526).

Virinum:
I know there are multiple ways to deny the access. I prefere adding the web.config file in the folder. There is also a .htaccess file in this folder. Why should a .htaccess file be possible and a web.config not?

alecpl:
Fixed by #6512.

PR #6512, opened 2018-11-04
give all temp files a constant prefix

johndoh:
this is inspired by #6511. the idea is the use a central function to generate all temp file names and to give all those files the same prefix. then in the gc code to only delete files which have that prefix. another advantage is that when used in an environment with shared temp storage roundcube will only delete its own files.

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -2235,11 +2235,11 @@ class rcmail extends rcube
 
             // generate image thumbnail for file browser in HTML editor
             if (!empty($_GET['_thumbnail'])) {
-                $temp_dir       = $this->config->get('temp_dir');
                 $thumbnail_size = 80;
                 $mimetype       = $file['mimetype'];
                 $file_ident     = $file['id'] . ':' . $file['mimetype'] . ':' . $file['size'];
-                $cache_basename = $temp_dir . '/' . md5($file_ident . ':' . $this->user->ID . ':' . $thumbnail_size);
+                $thumb_name     = md5($file_ident . ':' . $this->user->ID . ':' . $thumbnail_size) . '.thumb';
+                $cache_basename = rcube_utils::temp_filename($thumb_name, false, false);
                 $cache_file     = $cache_basename . '.thumb';
 
                 // render thumbnail image if not done yet
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -523,7 +523,7 @@ class rcube
 
         if ($tmp && ($dir = opendir($tmp))) {
             while (($fname = readdir($dir)) !== false) {
-                if ($fname[0] == '.') {
+                if (strpos($fname, RCUBE_TEMP_FILE_PREFIX) !== 0) {
                     continue;
                 }
 
@@ -1652,8 +1652,7 @@ class rcube
 
         if ($message->getParam('delay_file_io')) {
             // use common temp dir
-            $temp_dir    = $this->config->get('temp_dir');
-            $body_file   = tempnam($temp_dir, 'rcmMsg');
+            $body_file   = rcube_utils::temp_filename('msg');
             $mime_result = $message->saveMessageBody($body_file);
 
             if (is_a($mime_result, 'PEAR_Error')) {
--- a/program/lib/Roundcube/rcube_utils.php
+++ b/program/lib/Roundcube/rcube_utils.php
@@ -1388,4 +1388,40 @@ class rcube_utils
 
         return false;
     }
+
+    /**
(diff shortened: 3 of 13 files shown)


program/lib/Roundcube/rcube.php
```diff
@@ -523,7 +523,7 @@ public function gc_temp()
 
         if ($tmp && ($dir = opendir($tmp))) {
             while (($fname = readdir($dir)) !== false) {
-                if ($fname[0] == '.') {
+                if (!preg_match('/^'. RCUBE_TEMP_FILE_PREFIX .'/', $fname)) {
```
alecpl: strpos() would be simpler here.

program/lib/Roundcube/rcube_utils.php
```diff
@@ -1388,4 +1388,34 @@ public static function preg_error($error = array(), $terminate = false)
...
+        }
+
+        // On Windows tempnam() uses only the first three characters of prefix so use uniqid() and manually add the prefix
+        // Full prefix is required for garbage collection to recognise the file
+        $temp_file = $unique ? str_replace('.', '', uniqid($file_name, true)) : $file_name;
+        $temp_path = unslashify($temp_dir) . '/' . RCUBE_TEMP_FILE_PREFIX . $temp_file;
+
+        // Sanity check for unique file name
+        if ($unique && file_exists($temp_path)) {
+            $temp_path = get_temp_name($file_name);
```
alecpl: Forgot about self::?

alecpl: It should be a return statement, imo.

program/lib/Roundcube/rcube_utils.php
```diff
@@ -1388,4 +1388,34 @@ public static function preg_error($error = array(), $terminate = false)
...
+        }
+
+        return $temp_path;
```
alecpl: tempnam() creates the file to prevent from race conditions. We should do the same.

alecpl (changes requested):

program/include/rcmail.php
```diff
@@ -2235,11 +2235,11 @@ public function display_uploaded_file($file)
...
             // generate image thumbnail for file browser in HTML editor
             if (!empty($_GET['_thumbnail'])) {
-                $temp_dir       = $this->config->get('temp_dir');
                 $thumbnail_size = 80;
                 $mimetype       = $file['mimetype'];
                 $file_ident     = $file['id'] . ':' . $file['mimetype'] . ':' . $file['size'];
-                $cache_basename = $temp_dir . '/' . md5($file_ident . ':' . $this->user->ID . ':' . $thumbnail_size);
+                $thumb_name     = md5($file_ident . ':' . $this->user->ID . ':' . $thumbnail_size);
+                $cache_basename = rcube_utils::get_temp_name($thumb_name, false);
                 $cache_file     = $cache_basename . '.thumb';
```
alecpl: I would add .thumb extension to $thumb_name when calling get_temp_name(). Also, if we create the file in get_temp_name() code below will not work as intended. So, maybe when the 2nd arg is false we should not create the file automatically.

program/steps/mail/get.inc
```diff
@@ -76,9 +76,9 @@ if (!empty($_GET['_frame'])) {
 // render thumbnail of an image attachment
 if (!empty($_GET['_thumb']) && $attachment->is_valid()) {
     $thumbnail_size = $RCMAIL->config->get('image_thumbnail_size', 240);
-    $temp_dir       = $RCMAIL->config->get('temp_dir');
     $file_ident     = $attachment->ident;
-    $cache_basename = $temp_dir . '/' . md5($file_ident . ':' . $RCMAIL->user->ID . ':' . $thumbnail_size);
+    $thumb_name     = md5($file_ident . ':' . $RCMAIL->user->ID . ':' . $thumbnail_size);
+    $cache_basename = rcube_utils::get_temp_name($thumb_name, false);
     $cache_file     = $cache_basename . '.thumb';
```
alecpl: Add .thumb before calling get_temp_name().

program/lib/Roundcube/rcube_utils.php
```diff
@@ -1388,4 +1388,37 @@ public static function preg_error($error = array(), $terminate = false)
...
+
+    /**
+     * Generate a temporary file path in the Roundcube temp directory
+     *
+     * @param string $file_name String identifier for the type of temp file
+     * @param bool   $unique    Generate unique file names based on $file_name
+     *
+     * @return string temporary file path
+     */
+    public static function get_temp_name($file_name, $unique = true)
```
alecpl: Maybe `temp_filename` would be a better function name.

alecpl (changes requested):

Issue #6457, opened 2018-09-21
Space separated recipients


omarchehab98:
When editing an email's **To**, **CC**, and **BCC** fields, it's natural for someone to space separate the list of email addresses.

I am aware that the recipient fields expect a comma separated list of email address, so this is not a bug report, it's a usability suggestion.

## Steps To Reproduce

Type `alice@example.com bob@example.com charlie@example.com` in one of the recipient fields.

Roundcube: v1.3.6
Skin: larry
Plugins: managesieve, password, enigma

## Expected Behaviour 

Email is sent to the 3 email addresses `alice@example.com`, `bob@example.com`, and `charlie@example.com`.

## Actual Behaviour

Email is sent to the email address `"alice@example.com bob@example.com" <charlie@example.com>`.

thomascube:
Please not that the common format of entering recipients (mostly when using autocompletion) is "<display name><space><email-address>". With the display name prepended, the space character is already used to separate it from the actual email address and therefore cannot be used to separate recipients. With autocompletion showing suggestions, how would you allow to refine the suggestions (e.g. "John" yields 20 matches) if the space character will already terminate a (yet incomplete) recipient?

omarchehab98:
Look, I understand that this might be hard to implement because of a conflict with an existing feature. But quite frankly, it doesn't matter. I am reporting this as a serious usability issue that most people will have issues with.

Yesterday, I had one of my clients complain about this. I looked into it and it turns out they sent an email with the following **BCC:** `"a@example.com b@example.com ... n@example.com" <n@example.com>`. I don't blame them one bit. How can your system accept such a name in the first place? It's so easy to detect something is very wrong in this case. I mean, if an email address's associated name contains an email address, the least you can do is tell the user that something is seems off. In the current version of Roudcube, it just sends the email as if nothing is wrong. So, instead of sending an email to 14 recipients, the user would end up sending it to 1 with a name composed of 13 freaking email addresses-- that's insane!

If you don't want to change the fact that space doesn't separate email addresses, the least you can do is add a message that teaches the user the format of listing recipients or a pop up that warns the user that something is wrong with their recipients.

omarchehab98:
Compose a new email, add 3 email addresses to the **To** field

![alice@example.com bob@example.com charlie@example.com in the To field](https://user-images.githubusercontent.com/12089120/45923248-f479a580-beae-11e8-8b0c-9aa168726b37.png)

Save the email as a draft then open the draft you saved.

[...]

PR #6529, opened 2018-11-22
 Handle space seperated email adresses in the _to, _cc and _bcc field like seperate addresses.

intxcc:
See issue: https://github.com/roundcube/roundcubemail/issues/6457

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -4743,28 +4743,34 @@ function rcube_webmail()
   // checks the input fields before sending a message
   this.check_compose_input = function(cmd)
   {
-    var key,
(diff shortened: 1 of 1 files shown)


alecpl:
Unfortunately there's much more to it. For example:
1. In Elastic skin we have the recipient input widget that should be extended to support space separator.
2. Autocompletion does not work with space delimiter. I.e. if you enter "test@domain.tld " and now you'd start entering an another address it will not autocomplete.

So, my take on this is that we could fix this in Elastic, but for other skins it's more work probably not worth the effort.

alecpl:
Implemented by 218cd2a.

PR #6540, opened 2018-11-30
add listoptions container element

johndoh:
add a listoptions container element similar to those used in other menus to allow plugins to insert content into the menu.

### DIFF
--- a/skins/classic/templates/mail.html
+++ b/skins/classic/templates/mail.html
@@ -200,6 +200,7 @@
   </ul>
 </fieldset>
 <roundcube:endif />
+<roundcube:container name="listoptions" id="listoptionsmenu" />
 </div>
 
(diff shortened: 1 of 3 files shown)


alecpl:
Because you have no control over the dialog and Save button I say adding form elements to this menu might be tricky. What's your use-case?

johndoh:
You can interect with the dialog via the menu-open event. Save is more tricky yes but I have overloaded the rcmail.set_list_options() function to add extra form fields. My use-case is for my Swipe plugin - at the moment the content is inserted via javascript on page load. I think using a template container is more reliable, because its in the template the plugin content will always be inserted in the correct place relative to the save buttons.

Not relevent for Swipe but if you compare listoptions dialogs in Elastic and Larry inserting content via JS on the client side in the correct location is quite trickey unless you write diffrent code for each skin.

PR #6547, opened 2018-12-07
password: chpasswd: restrict allowed characters

apreiml:
An attacker can otherwise easily forge a password, which suffix is actually
a valid chpasswd input line to set the password of another user.


alecpl:
Control characters (not only \r\n) does not make sense in a password no matter which driver is used. I'd propose to do this in password::password_save() and return more use-friendly error message ('passwordforbidden').

apreiml:
@alecpl like this?

alecpl:
Yes, but with preg_match. See a32d72b661.

Issue #6565, opened 2018-12-21
login form inputs too wide in IE11


johndoh:
using latest git-master the input boxes on the login form when using IE11 are too wide:

![login](https://user-images.githubusercontent.com/88682/50330451-27a1f400-04f3-11e9-8a70-98ac6b97ed91.png)

BTW not sure if its intentional but in FF64, Edge and IE11 both login boxes have the red "invalid" marking around them on page load. It happened after 328c324

alecpl:
I fixed the border issue.

alecpl:
Fixed.

PR #6568, opened 2018-12-22
add option to build login form using divs

johndoh:
For #6565

According to https://github.com/twbs/bootstrap/pull/21569 there are some general issues with IE11, flex-box and tables. but it seems like only the login form is really affected in Elastic.

Rather than trying to hack around the shortcomings of IE11 I thought it might be easier and also most useful for the future to layout the login form using divs rather than a table. For current sinks though a table is still required so this PR allows for both. Keeping table as the default form of layout it adds a param to switch it to divs.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -2078,26 +2078,47 @@ EOF;
 
         $this->add_gui_object('loginform', $form_name);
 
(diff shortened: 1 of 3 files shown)


johndoh:
I see you fixed the IE11 problem which was the main purpose of this PR so I'll close this for now,

PR #6576, opened 2018-12-31
Hide inaccessible empty identity fields.

pabu:


alecpl:
Makes sense, but looks like indentation is broken.

pabu:
The GitHub web editor was waging a tabs vs spaces war on me it appears. Hopefully it looks better now.

alecpl:
Still the indentation does not look right. Anyway, I just noticed you used release-1.3 branch. We accept such changes to master only. Please, create a new PR.

Issue #6581, opened 2019-01-08
Domain DNS check fails when DNS servers don't support ANY as a query type


disturbio:
Hi,

After moving from DNS servers from bind to knot, roundcube started to fail validating e-mail addresses of some random, but valid domains. The reason is that `checkdnsrr($domain_part, 'ANY')` fails when asking to the knot dns server. There are reasons for DNS services to not support the ANY query type, and Cloudflare details their reasons they deprecated it [1], and also there is also work at IETF about this [2].

The effect of this failing is that the user cannot send an e-mail, as the check fails. The way to solve currently is to disable the dns_check, but would it be possible to use the check A and AAAA records directly instead of ANY and keep the function working as expected instead of breaking compatibility with the DNS Servers that will not implement that query type?

A quick way to test this is to set "1.1.1.1" as the dns server (as that's provided by cloudflare).

Thanks in advance.

[1]: https://blog.cloudflare.com/deprecating-dns-any-meta-query-type/
[2]: https://tools.ietf.org/html/draft-ietf-dnsop-refuse-any-07

alecpl:
We should check `SOA` record.

disturbio:
Checking the `SOA` record will work for the root domain, but it can fail for the subdomains (at least it will fail against bind and knot). As it's very common for universities or mailing lists, the check really should be a different one.

I looked up for a  #2175 and https://github.com/roundcube/roundcubemail/commit/e4acbbd8cc02b960bd5240538016b2c69c33d6bd I don't get a clear idea what's the goal with `ANY` there. But the idea of checking `SOA` makes me think it can be for reducing the number of queries). 

As if a domain doesn't have `MX` records set, the mail system will try the `A` record, it may be possible to remove the `MX` dns check (the mail system will use if exists, there seems to be no benefit that is there at a very superficial look), and then check for `A` || `AAAA` and if it doesn't find those try a `CNAME` one? This should minimize at least a little the possible performance problems.

alecpl:
Looks like you're right about SOA. We should try A, MX, CNAME, AAAA. In that order.

alecpl:
Fixed.

PR #6591, opened 2019-01-16
Change the query type ANY to MX/A/CNAME/AAAA for dns_check (#6581)

disturbio:
As query type ANY is not used by all dns servers, the domain validation
function checks for an A, MX, CNAME, and AAAA record (in that order).


disturbio:
This was tested against bind and knot, and has been running against 1.3 for some days without noticing any issues.

alecpl:
Thanks, slightly improved version committed in 61eb78ad64672.

PR #6594, opened 2019-01-20
[managesieve] fix: custom header & variable values are lost on submission

gabrielfin:
When adding a custom header or variable rule in any position except the first one, the value entered as the header or variable name is lost after submitting the form, which in turn causes a validation error stating that the input is required.

Steps to reproduce:
1. Create a filter
2. Add any rule
3. Add a custom header or custom variable rule
4. Enter any header or variable name, and complete the rest
5. Submit the form

Expected result:
The filter is saved correctly

Actual result:
Validation error stating that the header or variable name is required


alecpl:
Good catch, but the solution is not correct. I fixed this in fe35b38c28ca.

PR #6604, opened 2019-01-26
add theme-color meta tags

johndoh:
add theme-color meta tags to Elastic skin

### DIFF
--- a/skins/elastic/ui.js
+++ b/skins/elastic/ui.js
@@ -344,6 +344,17 @@ function rcube_elastic_ui()
         if (!$('#logo').data('src-default')) {
             $('#logo').data('src-default', $('#logo').attr('src'));
         }
+
+        // Add theme-color meta tags to color status bar on mobile devices
+        var theme_color = null;
+        $.each(['#layout > div > .header', '#layout > div.content', 'body'], function() {
+            theme_color = $(this).first().css('background-color');
+
+            if (theme_color !== undefined) {
+                $('<meta>').attr('name', 'theme-color').attr('content', theme_color).appendTo('head');
+                return false;
+            }
+        });
     };
 
     /**


program/include/rcmail_output_html.php
```diff
@@ -334,7 +334,16 @@ private function load_skin($skin_path)
         }
 
         foreach ((array) $meta['config'] as $key => $value) {
-            $this->config->set($key, $value, true);
+            // !default appended to a skin config option means it should not override main config
+            if (preg_match("/^(.*)\s!default$/", $value, $matches)) {
```
alecpl: Do this for string values only, I guess.

Will it work with skin extending? I.e. you define theme_color in main skin and its parent. Probably not, but I didn't test it.

alecpl (changes requested):

johndoh:
This needs a bit more work than I first thought. As well as the issue of extended skins (you are correct, it doesn’t work) there is also the issue of line 356, I don’t think !default options should be added to dont_override.

I had not considered the limitation of string values either. Perhaps it would be better to have 2 sections, one config which works exactly as now, and one called defaults or something similar for these overridable options?

The specific use case here is with theme_color. It is possible to change the menu bar color using _variables.less but that does not effect theme_color which would mean the skin has to be extended, I think that is overkill just to change a color so allowing it to be overridden by config seems like the natural solution.

Any thoughts you have are apricated or even if you can give me a hint how likely this is to get merged before I spend too much time on?

alecpl:
Could these meta tags be added using javascript after/on page load? Then we could just read color of  some special/hidden HTML element (or css variable?) and add meta tags. This way we'd use colors set in less.

alecpl:
What if I want different color than toolbar color? Would be better with https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables, e.g.
```
:root { --theme-color: #ddd; }
```
and read it with `document.documentElement.style.getPropertyValue('--theme-color')`. I guess we could use that to assign a few more color variables via less file, e.g. for main skin color. That might be useful in future.

Anyway, that delay might not be nice. So, maybe we could find another way.

johndoh:
Using CSS veriables wil work, browser support looks ok for it.

You cannot control theme-color directly from CSS so there would have to be some JS involved which means delay. The only way I can find around that is to have the tag in the raw HTML.

Perhaps the `favicon` template object could be extended to also set these tags. Then the colours could be overridden from `skin_logo` or possibly a new `skin_favicon` config, similar to `skin_logo` but just for the favicon and related settings e.g. the sort of thing generate by https://realfavicongenerator.net/. A downside would be having the colours in 2 places - Less and config. An upside would be it is a step towards turning Roundcube into a webapp (see https://github.com/roundcube/elastic/issues/19)

alecpl:
I would add a new "meta" object for templates that would be controlled by meta.json (not config) and would generate something like:
```
<link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
<link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png">
<link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png">
<link rel="manifest" href="/site.webmanifest">
<link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
<meta name="msapplication-TileColor" content="#da532c">
<meta name="theme-color" content="#ffffff">
<meta name="viewport" content="width=device-width, initial-scale=1.0, shrink-to-fit=no, maximum-scale=1.0">
```
We don't need all these links now, but we should add possibility to define them in meta.json or with a plugin api hook. So, for example m eta.json could look like this.
```
{
  "name": "Elastic",
  "config": {
    ...
  },
  "meta": {
    "viewport": "width=device-width, initial-scale=1.0, shrink-to-fit=no, maximum-scale=1.0",
    "theme-color": "#ffffff"
  },
  "links": {
    ...
  }
}
```
That will indeed not allow to keep colors in only one place, but I don't see it as a big problem.

PR #6629, opened 2019-02-19
Update labels.inc

syspir:
update translate


alecpl:
Fixed on https://www.transifex.com/roundcube/roundcube-webmail.

PR #6646, opened 2019-02-26
Parse abook entries more reliably

dchisolm:
Update the squirrelmail_user_copy plugin to use the same method of parsing abook entries as squirrelmail uses. 

If a user has entered something that has been quoted such as double quotes or pipes, this will
parse the address book entry better than exploding on the pipe alone


alecpl:
PHP manual states: "The locale settings are taken into account by this function. If LC_CTYPE is e.g. en_US.UTF-8, strings in one-byte encodings may be read wrongly by this function" for fgetcsv().

So, maybe converting charset first and then passing to str_getcsv() would be better?

dchisolm:
Good catch. Update applied.

Do you know which version it became possible to use international characters such as "Pelé@mail.tld"? I have access to a few older installs and this has only worked for me on this 1.4-git version.

alecpl:
It's 1.4.

PR #6664, opened 2019-03-06
address_book_type plugins fix

urusha:
If both LDAP and plugin address books are used, the LDAP one is forced as default (until changed in preferences), even if `address_book_type` is empty.  This patch makes RC check plugins for available address books first, if `address_book_type` is empty. And the plugins address books are now listed before the LDAP ones.

While this could be ignored with the larry skin, we get  annoying 'advanced search' window appearing every time we switch to the contacts page with the new elastic skin.

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -328,6 +328,9 @@ class rcmail extends rcube
             );
         }
 
+        $plugin = $this->plugins->exec_hook('addressbooks_list', array('sources' => $list));
+        $list   = $plugin['sources'];
+ 
         if (!empty($ldap_config)) {
             foreach ($ldap_config as $id => $prop) {
                 // handle misconfiguration
@@ -345,9 +348,6 @@ class rcmail extends rcube
                 );
             }
         }
-
-        $plugin = $this->plugins->exec_hook('addressbooks_list', array('sources' => $list));
-        $list   = $plugin['sources'];
 
         foreach ($list as $idx => $item) {
             // register source for shutdown function


alecpl:
Because usually LDAP addressbooks are global ones and it makes sense to have them on top.

I have no idea what "advanced search" window you're talking about.

urusha:
The order is the side effect of the patch, if we could save the original order and add an ability to set plugin's address book as default - that would be great. Actually this is a bug, since documentation says:
```
// If set to '' then no address book will be displayed or only the
// addressbook which is created by a plugin (like CardDAV).
```
But currently it's not so.

`advanced search` window is appearing after click on `Search` button, but with ldap address book having `'searchonly' => true` it appears automatically (my case) - and that's the problem.

urusha:
> Because usually LDAP addressbooks are global ones and it makes sense to have them on top.

By the way, SQL addressbook is above LDAP. Then why plugin's addressbook can't be (especially if this could lead to auto `advanced search` popups)?
But anyway, we could make this configurable, by supporting something like this:
```
// address_book_type_order ?
$config['address_book_type'] = array ('sql', 'ldap', 'plugin')
```
And if we have more than one address book engine plugin, they will be sorted in `$config['plugins']` order, I guess.

alecpl:
The advanced search popup should be fixed. That's the bug, not the addressbooks order. You can modify the order from a plugin (kolab_addressbook plugin has an option to do this). #6679.

`address_book_type=''` disables sql addressbook, but does not hide ldap, so indeed the option description is not correct. I'm not sure what was the initial intention, but looks like an issue, indeed. #6680.

PR #6667, opened 2019-03-10
Improve mail visibility 

andresbott:
I have slightly modified the css in order to make selected emails more visible, and added hover color to non selected mails.

minor changes:  
- add a line to separate mail header from body and added a pointer 
-  add a pointer to popup menu elements

Snapshot_2019-03-10_at_12 44 22


alecpl:
Thanks. PRs should target master branch and contain only relevant commits.

Anyway, the line below header could be considered, but I don't like the selection change.

AndresBott:
Thanks for response, will do the PR to master next time, 
too bad, i thought the selection makes it more usable, any plan for something similar in the future?

PR #6669, opened 2019-03-11
Fix for: Password: De-duplicate ldap drivers code

army1349:


plugins/password/drivers/ldap_simple.php
```diff
@@ -228,8 +228,8 @@ function search_userdn($rcmail, $ds)
 
         $this->_debug("S: OK");
 
-        $search_base   = rcube_ldap_password::substitute_vars($search_base);
-        $search_filter = rcube_ldap_password::substitute_vars($search_filter);
+        $search_base   = substitute_vars($search_base);
+        $search_filter = substitute_vars($search_filter);
```
alecpl: That should be self::substitute_vars().

plugins/password/drivers/ldap_simple.php
```diff
@@ -31,7 +31,7 @@ class rcube_ldap_simple_password
 {
     private $debug = false;
     private $user;
-    private $conn;
+    protected $conn;
```
alecpl: Let's make all these protected.

plugins/password/drivers/ldap_simple.php
army1349: My bad.

plugins/password/drivers/ldap_simple.php
army1349: OK

army1349:
Thanks!

PR #6672, opened 2019-03-11
favicon 64x64

iotaka:
Hello,

is possible update the icon with a dimension of 64x64, in this way on 4K displays the resulting favicon is crisp and clean.

Cheers


alecpl:
Done via 92ae70e78dfc84. It also unified the file across all skins.

PR #6674, opened 2019-03-12
Add a skip duplicates option when importing contacts

dheineman:
This PR adds a new option for users to choose what to do with duplicates when importing new contacts. By default the option (checkbox) is checked and thus will skip duplicates (as is the current behaviour).


alecpl:
Conflicts need to be resolved. I think it would be good to disable the checkbox when "Replace the entire address book" is checked.

PR #6683, opened 2019-03-14
Treelist - Default value for node.collapsed when node.children.length = 0

tpayen:
When a node has children but with length = 0 the node.collapsed value is not set so the node is considered as expanded.
This is problematic when we use a dynamic tree loaded by ajax, the node is always considered has expanded but it's isn't.
I think by default node.collapsed needs to be set to true when there is no children because it's what we expect to have


tpayen:
This is useless, sorry

PR #6691, opened 2019-03-23
create Elastic dark mode

johndoh:
With Dark Mode becoming more popular I thought I'd see what it might look like in Elastic. Currently its only supported by [Safari,](https://webkit.org/blog/8475/release-notes-for-safari-technology-preview-68/) its in development for [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=889087) and [FireFox](https://bugzilla.mozilla.org/show_bug.cgi?id=1494034).

The idea is to use the `@media (prefers-color-scheme: dark)` media query to change background and text colors for a dark look.

Current Status: This is a proof of concept only. I have not been through all the styles yet and there is some unnecessary code.

I'm interested to know if any one else thinks this is a good idea or if any one would like to help.


alecpl:
That's a lot of code. I though about using `filter:invert()` (https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/invert) to make it simpler, but I didn't have time yet to investigate it's indeed a good idea.

Also, aside of `prefers-color-scheme` use it would be good to have a mode switch in the UI.

johndoh:
Agreed. But not all of it is really needed. The reason its so long is that I started by making a list of every use of an `@color-` var but not every color will need to change so some of the code and some of the additional vars will not be needed.

I think the issue with ``filter:invert()``` will be that some colors need to change not just flip from light to dark.

BTW In the process I found some `@color-` vars which are in `colors.less` but never used:
```
@color-link-secondary
@color-taskmenu-button-special
@color-taskmenu-button-special-background
@color-taskmenu-button-selected-hover
@color-taskmenu-button-special-hover
@color-taskmenu-button-special-background-hover
@color-searchbar-icon-active
@color-popover-header
@color-popover-header-background
```
And some places where colors are hard coded:
```
styles/widgets/common.less
    Line 268:   border: 2px solid #fff;
styles/widgets/editor.less
    Line 54:    background: #fff;
    Line 410:   border-color: #c5c5c5;
    Line 431:   border-bottom-color: #fff;
    Line 857:   border-bottom-color: #fff !important;
styles/widgets/forms.less
    Line 1128:  background: #fff;
styles/widgets/messages.less
    Line 189:   color: #fff;
    Line 201:   color: #fff;
styles/widgets/toolbar.less
    Line 134:   background-color: #fbfbfb;
```

I wondered about trying create something like a `color-scheme.less` type file which could be used to override the the palette with any scheme but at first glance that did not look straight forward so I thought I'd start with just making a dark mode.

johndoh:
Another option might be to use [CSS variables](https://www.w3schools.com/css/css3_variables.asp). [Browser support](https://caniuse.com/#search=css%20var) for them is improving with modern versions of almost all browsers listed in the Elastic readme included. The main exception being IE11. I found [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill) which might help.

iambumblehead:
dark mode please! :)

mcnesium:
When inverting everything by filter, there is one possible solution to also rotate the colors back:

    filter: invert(1) hue-rotate(180deg);

mcnesium:
A general drawback of inverting everything is though, that you will need to re-invert any image (e.g. picture attachment preview) afterwards.

jayhasgin:
Inverting colors works well in elastic theme (except for the watermark.html doc),

by inserting following head style in '/program/include/rcmail_output_html.php', line ~1912

```
$page_header .= '<style>
html { 
background:#aaa; 
filter: invert(100%) !important; 
-webkit-filter: invert(100%) !important; 
}  
img, iframe { 
filter: invert(100%) !important; 
-webkit-filter: invert(100%) !important; 
}  
</style>';
```

It needs a switch though, so user can toggle between light and dark.

mcnesium:
> It needs a switch though, so user can toggle between light and dark.

the automatic switch is `@media (prefers-color-scheme: dark)` ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme))


A patch for the [Dockerfile](https://github.com/roundcube/roundcubemail-docker/blob/master/apache/Dockerfile) which includes `watermark.html` could look like this:
```
# patch dark mode
ARG CSS="@media (prefers-color-scheme: dark) { body img, body iframe, body { filter: invert(1) hue-rotate(180deg); } }"                                                      
RUN echo "${CSS}" >> /usr/src/roundcubemail/skins/elastic/styles/styles.css && \
    sed -i "12imargin:0;" /usr/src/roundcubemail/skins/elastic/watermark.html && \
    sed -i "15i${CSS}" /usr/src/roundcubemail/skins/elastic/watermark.html
```

alecpl:
@johndoh, I had to do this from scratch, but your PR was very helpful. There's still some work to be done, e.g. I'm going to work on update to TinyMCE5, then dark mode for it. There's probably some missing parts or bugs, but what's in master right now imho is quite nice. Thanks again for your great work.

https://kolabian.wordpress.com/2020/07/07/elastic-dark-mode/

gaellelo:
This is so cool! Thanks a mil for implementing it, our users will be super happy. :)

iambumblehead:
@alecpl I look forward to seeing this in my kolab now web mail :) @johndoh thank you

iambumblehead:
@alecpl politely, please bring dark mode to kolabnow web mail. It is still not found there.

iambumblehead:
@alecpl dark mode is still not found at kolabnow web mail :|

iambumblehead:
@alecpl still no kolabnow dark mode :(

alecpl:
@iambumblehead please, stop this. There was no stable release yet. It will take some time to do this upgrade in Kolab, and then another time to push it to KolabNow. It's not up to me, anyway.

PR #6724, opened 2019-04-22
Elastic: Disable subject selection in messagelist

HLeithner:
In chromium based browsers it is possible to mark the text by mistake if you want to move a mail by drag'n'drop.


alecpl:
I'm unable to reproduce.

HLeithner:
hmm strange seams to be a bug in vivaldi... you can close it or merge it depends on your policy

alecpl:
I'll consider, but maybe it should be applied to the whole `tr`, not only subject element.

HLeithner:
That didn't worked for me, all other elements can't be marked only this span can be selected...

It seams that Javascript insert user-select: auto !important from a, this gets added when start dragging.
image

HLeithner:
Thanks

PR #6730, opened 2019-04-27
Elastic: Change "All headers..." dialog to use monospace font

peterdey:
The "All headers" popup iframe uses the default font - but many (particularly multi-line) headers expect to be displayed in a monospace font.  Examples of this include the "Received" header, or rspamd's "X-Spamd-Result" header.

### DIFF
--- a/skins/elastic/styles/widgets/dialogs.less
+++ b/skins/elastic/styles/widgets/dialogs.less
@@ -253,3 +253,10 @@ html.touch .popover {
         display: none;
     }
 }
+
+/** Headers popup dialog **/
+body.action-headers {
+    font-family: Consolas, 'Courier New', Courier, monospace;
+    font-size: 90%;
+    line-height: 1.25;
+}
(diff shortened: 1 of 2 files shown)


alecpl:
I do not agree. There's nothing in Received: header format that would require monospace.

peterdey:
Could we perhaps consider merging the changes to `templates/includes/layout.html`, so that folks who might like monospaced headers only need to add a custom rule to `_styles.less`?

alecpl:
We could consider this. Some action names would however be problematic, or at least not nice looking. E.g. `plugin.managesieve-vacation` would become `action-plugin.managesieve-vacation` class (note the dot inside). Maybe we just should use asciiwords() function from lib/Roundcube/bootstrap.php.

peterdey:
Would you still want the `action-` prefix?  Or rather just have no class at all for the "none" case?

Also, I would presume the `$replace_with` parameter on asciiwords() should be set to something like "_", so that `plugin.managesieve-vacation` is still readable (i.e. becomes `action-plugin_managesieve-vacation`)

alecpl:
I would use minus sign as a replacement char. And probably `action-none` may have some use, so I'd keep it.

alecpl:
The other PR merged.

PR #6732, opened 2019-04-27
Correct upgrade database process

ChristopheAtYelty:
During my upgrade (from 1.1.2 to 1.3.8) the following error appears:

```
Updating database schema (2013052500)... [FAILED]
Error 500: Error in DDL upgrade 2013052500: [1050] Table 'cache_shared' already exists
```
I propose you to drop the table before to recreate it.

### DIFF
--- a/SQL/mysql/2013052500.sql
+++ b/SQL/mysql/2013052500.sql
@@ -1,3 +1,5 @@
+DROP TABLE `cache_shared`;
+
 CREATE TABLE `cache_shared` (
  `cache_key` varchar(255) /*!40101 CHARACTER SET ascii COLLATE ascii_general_ci */ NOT NULL,
  `created` datetime NOT NULL DEFAULT '1000-01-01 00:00:00',


alecpl:
If you really have had 1.1.2 that upgrade file (2013052500) should not be execute when upgrading to 1.3.8. There was something wrong with your database or schema version stored in system table.

I do not agree with dropping the table. You're first person with this problem, so I rather think it's not Roundcube bug, but some mess in your installation.

christophefromparis:
I'm using a docker version of Roundcube (before the official version) with the 1.1.2 and I have the version 2013011700 in my database.

alecpl:
If you get the original package from https://github.com/roundcube/roundcubemail/releases/download/1.1.2/roundcubemail-1.1.2.tar.gz you'll see 2015030800.

christophefromparis:
Ok, I'll upgrade manually in order to restore a normal situation. 
Thanks.

PR #6734, opened 2019-04-28
Allow to use unix socket for Redis server connection. 

earendilfr:
The format of the redis_hosts options should be "unix:/path/to/socket"

Tested in version 1.3.9 in my personnal server.


alecpl:
The code is ugly and I think also buggy. No long lines, no double connect() calls. What if someone's hostname is 'unix'? And we should require syntax with `unix://` prefix for consistency with unix socket arguments in other places (memcache and sqlite).

alecpl:
Done in 2dccbf28798dcc.

PR #6747, opened 2019-05-11
Fix #6746: update rcube_utils::parse_hosts

### LINKED ISSUE #6746: rcube_utils::parse_host behavior for %d (and %t, %z) should not remove domain if server has no subdomain part
In config, various hostname-related parameters such as `$default_host`, `$username_domain`, etc. have dynamic substitutions available such as `%n`, `%t`, `%d`, etc.  The options `%d`, `%t`, and `%z` are intended to return a domain by stripping off "the first part" of the server name.  If the server is a subdomain (e.g., mail.domain.tld) then this works fine... but if the server is simply a domain (e.g., domain.tld), these options strip off the domain and return only the tld, which is not desirable behavior.

In these cases, `rcube_utils::parse_host` is stripping the domain because it is not checking whether the hostname has a subdomain... it merely strips off the first element regardless.  Instead, it should ensure that it retains the domain in cases where a subdomain does not exist.

**Issue:**
If `$_SERVER['SERVER_NAME']`, `$_SERVER['HTTP_HOST']`, and/or `$host` do not have a subdomain part (e.g., only domain.tld), then rc_utils::parse_host options for `%t`, `%d`, and `%z` respectively will strip off the domain and return only tld.

**Steps to reproduce:**
1. Install rcmail in standard webhost with no subdomain (access via `https://domain.tld/roundcubemail` and **not** via `mail.domain.tld`.).

2. Set: 
```
config['username_domain'] = '%d';
config['username_domain_forced'] = true;
```

3. Log in.

**Expected behavior:** User identity should be `user@domain.tld`.
**Actual behavior:** User identity is `user@tld`, which is not correct.

**Regression:** This behavior exists in the current codebase and goes back at least as far as RC 1.1.12 (and likely since these substitutions were introduced).

**Fix:** The regexp substitutions for `$t`, `$d`, and `$z` in `rcube_utils::parse_host` need to be updated to ensure they do not strip off the actual domain and leave only the tld.  This can be done by ensuring there are at least two periods in the hostnames before substitution, and there might be an elegant way to do this within the regexp itself.

I'll try to submit a PR with the fix when I have a few minutes, if nobody else gets to it first.

cepheid666:
Fixes bug #6746 where rcube_utils::parse_hosts domain-level substitutions %t, %d, %z could return only tld when underlying hostname was just domain.tld with no subdomain (e.g., NOT mail.domain.tld). New regexps will only strip subdomain, will not strip domain without subdomain.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -36,6 +36,7 @@ CHANGELOG Roundcube Webmail
 - Fix so bin/install-jsdeps.sh returns error code on error (#6704)
 - Fix bug where bmp images couldn't be displayed on some systems (#6728)
 - Fix bug in parsing vCard data using PHP 7.3 due to an invalid regexp (#6744)
+- Fix bug in rcube_utils::parse_hosts where %t, %d, %z could return only tld (#6746)
 
 RELEASE 1.4-rc1
 ---------------
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -37,6 +37,11 @@ $config['db_dsnw'] = 'mysql://roundcube:pass@localhost/roundcubemail';
 // %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
 // %s - domain name after the '@' from e-mail address provided at login screen
 // For example %n = mail.domain.tld, %t = domain.tld
+// (N.B. %t, %d remove only the first part up to domain-level, i.e., if
+//       an FQDN is provided, they will not return anything shorter than
+//       domain.tld or domain.sd.cc. Single hostnames are returned as-is.
+//       NOTE: Currently only 2-letter 'sd' supported, be aware when using
+//             this setting if your ccSLD is longer.
 $config['default_host'] = 'localhost';
 
 // SMTP server host (for sending mails).
@@ -49,6 +54,11 @@ $config['default_host'] = 'localhost';
 // %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
 // %z - IMAP domain (IMAP hostname without the first part)
 // For example %n = mail.domain.tld, %t = domain.tld
+// (N.B. %t, %d, %z remove only the first part up to domain-level, i.e., if
+//       an FQDN is provided, they will not return anything shorter than
+//       domain.tld or domain.sd.cc. Single hostnames are returned as-is.
+//       NOTE: Currently only 2-letter 'sd' supported, be aware when using
+//             this setting if your ccSLD is longer.
 $config['smtp_server'] = 'localhost';
 
 // SMTP port (default is 25; use 587 for STARTTLS or 465 for the
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -41,7 +41,7 @@ $config['db_dsnw'] = 'mysql://roundcube:pass@localhost/roundcubemail';
 //       an FQDN is provided, they will not return anything shorter than
 //       domain.tld or domain.sd.cc. Single hostnames are returned as-is.
 //       NOTE: Currently only 2-letter 'sd' supported, be aware when using
-//             this setting if your ccSLD is longer.
+//             this setting if your ccSLD is longer.)
 $config['default_host'] = 'localhost';
 
 // SMTP server host (for sending mails).
@@ -58,7 +58,7 @@ $config['default_host'] = 'localhost';
 //       an FQDN is provided, they will not return anything shorter than
 //       domain.tld or domain.sd.cc. Single hostnames are returned as-is.
 //       NOTE: Currently only 2-letter 'sd' supported, be aware when using
-//             this setting if your ccSLD is longer.
+//             this setting if your ccSLD is longer.)
 $config['smtp_server'] = 'localhost';
 
 // SMTP port (default is 25; use 587 for STARTTLS or 465 for the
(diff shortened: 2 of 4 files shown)


johndoh:
This only works for TLDs, not for second level domains eg. example.co.uk. In this case example is still stripped just as it is now. Your proposal simply changes the existing rule from "there must be at least one dot left in the hostname" to "there must be at least 2 dos in the hostname" IMO that is more confusing as it means the behaviour is different for different domains.

If you have different URLs for your Roundcube instance then why not use `include_host_config` to provide differnt configs?

alecpl:
I don't see an issue here. Current behavior is "remove first part, no matter how many parts there are, where at least two parts exist". The new bahavior will be "remove first part, but only if at least three parts exist". It makes sense. The point is to never end with top-level domain (no dots). So:
- host -> host
- domain.tld -> domain.tld
- host.domain.tld -> domain.tld
- sub.host.domain.tld -> host.domain.tld

cepheid666:
> This only works for TLDs, not for second level domains eg. example.co.uk. In this case example is still stripped just as it is now.

You're right, thanks for catching this. This can be easily resolved by including a second clause in the negative lookahead, as follows:
`/^[^.]+\.(?![^.]+$)/` --> `/^[^.]+\.(?!(?:[^.]{3,}|(?:[^.]{2,4}\.[^.]{2}))$)/`
(tested and works)

The above mod does assume that country-code SLDs are only 2 characters, as most of them are. Unfortunately (per [Wikipedia](https://en.wikipedia.org/wiki/Second-level_domain)) it turns out there are a number of longer 2LDs, some are 7 or more characters... that isn't something that can be handled in a simple regex, and would require loading a table of valid TLDs and ccSLDs to check against, which would be a much more complex fix.  Ultimately that is the right solution, but the above simple solution is at least better than the current state.

I can incorporate this addition into the pull request if you wish.

> Your proposal simply changes the existing rule from "there must be at least one dot left in the hostname" to "there must be at least 2 dos in the hostname"

For the original proposal, that is correct, because if there is only one dot (domain.tld) then stripping the first part means you're left with only a TLD, which is not a valid domain. `rcube_utils::parse_hosts` substitutions are used in a number of places, including setting the imap host... if its default behavior allows returning an invalid host (only a tld), that is a bug.  So yes, there must be two dots in the hostname to allow stripping the first part.

Your mentioning of SLDs is a very good point -- my original consideration was very U.S.-centric but you're exactly right that example.co.uk cannot be stripped.  My modification above will correct that, at least for 2-character ccSLDs.  I can incorporate that into the PR if you would like.  Handling the cases for longer ccSLDs can't be done in a simple regex... ultimately it SHOULD be done, and I think would require a lookup table of valid matches, but this PR (or mod to the PR) at least prevents most of the grossly incorrect behavior (allowing return of just TLD or 2-letter ccSLD) that occurs with the current code.

> IMO that is more confusing as it means the behaviour is different for different domains.

Per @alecpl, I don't think it's more confusing, but you're right that it doesn't cover ALL ccSLDs.  My suggested mod above will at least cover 2-letter ccSLDs and I can incorporate that if you would like.  As mentioned, handling even longer ones will require a significantly more complicated fix.

That said -- the current behavior is absolutely not correct.  It should never be allowed to return only a TLD.  Returning a ccSLD is also incorrect but that's harder to fix more generally.  (Darn ICANN rules.)

> If you have different URLs for your Roundcube instance then why not use `include_host_config` to provide differnt configs?

That's partly covered in the discussion in #6746, but as the simplest example: in many cases, particularly for virtual hosting, mail.domain.tld and domain.tld point to the same place, and the RoundCube instance can be accessed by both; the settings should be identical for both, and requiring `include_host_config` to provide the same config is cumbersome and overkill.  Now imagine that host hundreds of such domains, and the only difference between them is literally the hostname (I want all the same configs for the rest, just different server names)... the ONLY thing in the per-host files would be to set `default_host` and/or `username_domain` and similar settings. (This is, in fact, a pretty common situation.)  Why require hundreds of per-domain config files when the same effect can be easily resolved through using a `%d` substitution in a single config?

But, more to the point, the current behavior is simply wrong, and breaks standards.  A TLD is never a valid hostname, and therefore it should not be returned by a function whose purpose is to return a valid hostname.

I admit that the fix I propose, even with the above mod, doesn't fully solve the problem for every allowable ccSLD... but it solves it for 99% of them, which is at least better than now...

Let me know if you'd like me to incorporate the regexp fix for 2-character ccSLDs, I'll add that to the PR.

cepheid666:
I've updated the PR: `rcube_utils::parse_hosts` will now handle 2-char ccSLDs per the mod I proposed above.  As mentioned earlier, handling longer ccSLDs is would need to be considered more carefully.  Some countries restrict their SLDs and allow only 3rd-level domain registration, but some countries allow any SLD to be registered (example: .ac, .ly)... so this can't be done within a simple regexp.  Most likely, the correct solution would require a lookup table of registered ccSLDs to be matched against (a match means preserve the ccSLD), but this would be a fairly complex solution and the table would need to be updated periodically from ICANN or whatever is the appropriate body.

Even the 2-letter handling above might potentially be problematic, if some countries allow arbitrary 2-letter SLDs to be registered.  For that reason, I might recommend just implementing my first commit, which simply ensures never returning a bare TLD.  It won't accommodate reserved SLDs but, per above, doing that robustly gets complicated quickly.

Nonetheless, at the very least, the first commit is imperative as returning a bare TLD is never correct under any circumstances.

alecpl:
I think I will merge first two commits only. The rest looks too much to me.

cepheid666:
Do you want me to update the config files to include a warning to those with ccSLDs that they need to be aware of this behavior?

alecpl:
No. I merged the first commit, changelog update, and added tests.

PR #6748, opened 2019-05-11
add dialog-open JS event

johndoh:
Add a `dialog-open` JS event, similar to the `menu-open` event to allow plugins to detect when a dialog is opened.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -8162,6 +8162,8 @@ function rcube_webmail()
     // Don't propagate keyboard events to the UI below the dialog (#6055)
     popup.parent().on('keydown keyup', function(e) { e.stopPropagation(); });
 
+    this.triggerEvent('dialog-open', { id:popup.attr('id') });
+
     return popup;
   };
 


program/js/app.js
```diff
@@ -8162,6 +8162,8 @@ function rcube_webmail()
     // Don't propagate keyboard events to the UI below the dialog (#6055)
     popup.parent().on('keydown keyup', function(e) { e.stopPropagation(); });
 
+    this.triggerEvent('dialog-open', { id:popup.attr('id') });
```
alecpl: I'm not sure every dialog has an id. Probably `{obj: popup}` would be better.

johndoh: jQueryUI adds this ID if there is none present (it creates a unique ID for each dialog in this case)

Note: this is the ID of the content frame of the dialog rather than of the overall dialog

alecpl:
I prefer passing popup object anyway. Done in c287dad54ef6e5

PR #6749, opened 2019-05-12
Add infrastructure to apply CSS styles to specific action dialogs/screens

peterdey:
As discussed in #6730

Adds the ability to apply CSS styles to specific action dialogs/screens.

Examples:
* action-headers - The headers popup
* action-plugin-managesieve-action - The managesieve filter creation screen


zjyysc11:
Kcash-Wallet

Issue #6450, opened 2018-09-15
enigma plugin: gpg redressing attack


hannob:
I want to report a security vulnerability in the way signed messages are displayed in the enigma plugin.

Signed messages are displayed with a green box in the message window. As this is space that is controlled by the sender of a mail it should be considered attacker controlled.

An attacker trying to fake a mail signature can simply rebuild that green box with HTML/CSS and add that to the top of the mail. (I skipped creating a PoC for this, as I already did this for several other mail clients and rebuilding things in HTML gets boring, but it should be straightforward obvious that this is possible.)

The information about a valid signature must not be displayed in attacker-controlled space. One way of fixing this is to put the signature info box above the mail headers. (This is what enigmail did after I reported it to them.)

jensvoid:
Note that this issue also applies to the way S/MIME signatures are presented to the user by Roundcube/Enigma.

alecpl:
Proposed solution: #6759.
![enigma-msg](https://user-images.githubusercontent.com/546788/57927553-a63a6c00-78ae-11e9-9c8c-1806339907c1.png)
![enigma-settings](https://user-images.githubusercontent.com/546788/57927556-a8042f80-78ae-11e9-9d65-1e53b8154480.png)

lambdafu:
As an alternative, maybe consider opening attachments in their own tab or window. There is a huge benefit from having only one signature or encryption context per message.  Then you can also verify that the scope of that context is the whole email and mitigate wrapping attacks.

MattyShires:
Has any action been done on this?

alecpl:
No. There's a PR waiting for comments and some more considerations if it's a appropriate way of fixing the issue.

JohnRDOrazio:
Wouldn't it be possible to place the message box in the same space as the "This message contains attached PGP key(s)" message box? i.e. https://github.com/roundcube/roundcubemail/blob/dec1d668ed37f8d4662eb37a37488f3fcce5fa43/plugins/enigma/lib/enigma_ui.php#L1098

JohnRDOrazio:
I tested my proposed solution on my own Roundcube instance, and it worked, so I've made PR #10007 to address this. The results on my live Roundcube instance:

<img width="622" height="252" alt="Image" src="https://github.com/user-attachments/assets/f8fec943-e935-491b-9dc9-5721dd3cc9ed" />

PR #6759, opened 2019-05-17
Enigma: Redressing attack prevention - security background (#6450)

alecpl:
See the ticket for some screenshots and discussion.

### DIFF
--- a/plugins/enigma/enigma.js
+++ b/plugins/enigma/enigma.js
@@ -63,6 +63,12 @@ window.rcmail && rcmail.addEventListener('init', function(evt) {
                 });
             }
         }
+
+        if ($('.enigma-bg-preview').length) {
+            rcmail.enigma_bg_post();
+            $('#rcmfd_enigma_bg_icon, #rcmfd_enigma_bg_scale, #rcmfd_enigma_bg_angle')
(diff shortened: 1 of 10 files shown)


dereks:

```
git fetch upstream
git rebase upstream/master
```

LBBO:
I'd like to state that this approach offers next to no security improvement (as far as I can tell by just skimming over the code), since it only modifies the box's appearance. An attacker can still include an HTML block using certain IDs and classes and all images will be set for them by Roundcube.

I agree with the opinion in #6450 that signature verification must be displayed outside of attacker's control in order to not compromise on security. Sure, the UX might suffer in some cases, but for most use cases people wouldn't tell the difference.

JohnRDOrazio:
Is it possible to display the message box in the same area as the "This message contains attached PGP key(s)." message box i.e. https://github.com/roundcube/roundcubemail/blob/dec1d668ed37f8d4662eb37a37488f3fcce5fa43/plugins/enigma/lib/enigma_ui.php#L1098? That would avoid the message box being a part of the message itself.

JohnRDOrazio:
I tested my proposed solution on my own Roundcube instance, and it worked, so I made PR #10007 with the proposed fix.

PR #6772, opened 2019-05-25
add support for SameSite cookie attribute

johndoh:
PHP 7.3.0 adds support for the SameSite cookie attribute, see https://wiki.php.net/rfc/same-site-cookie..

To use it the setcookie method must be called in a new way which is not backwards compatible so I added a PHP version check.

### DIFF
--- a/program/lib/Roundcube/rcube_utils.php
+++ b/program/lib/Roundcube/rcube_utils.php
@@ -49,10 +49,21 @@ class rcube_utils
             return;
         }
 
-        $cookie = session_get_cookie_params();
-        $secure = $cookie['secure'] || self::https_check();
+        $attrib             = session_get_cookie_params();
+        $attrib['expires']  = $exp;
+        $attrib['secure']   = $attrib['secure'] || self::https_check();
+        $attrib['httponly'] = $http_only;
 
-        setcookie($name, $value, $exp, $cookie['path'], $cookie['domain'], $secure, $http_only);
+        // session_get_cookie_params() return includes 'lifetime' but setcookie() does not use it, instead it uses 'expires'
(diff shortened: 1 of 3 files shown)


brot:
Will this patch be part of version 1.4?

cgzones:
Any update on this one?

program/lib/Roundcube/rcube_utils.php
```diff
@@ -49,10 +49,18 @@ public static function setcookie($name, $value, $exp = 0, $http_only = true)
             return;
         }
 
-        $cookie = session_get_cookie_params();
-        $secure = $cookie['secure'] || self::https_check();
+        $attrib             = session_get_cookie_params();
+        $attrib['expires']  = $exp;
+        $attrib['secure']   = $cookie['secure'] || self::https_check();
```
cgzones: this should be
`$attrib['secure']   = $attrib['secure'] || self::https_check();`

program/lib/Roundcube/rcube_utils.php
```diff
@@ -49,10 +49,18 @@ public static function setcookie($name, $value, $exp = 0, $http_only = true)
...
+        $attrib['httponly'] = $http_only;
```
cgzones: php warns about the extra 'lifetime' array entry
```
PHP Warning:  setcookie(): Unrecognized key 'lifetime' found in the options array in /var/www/sites/roundcubemail/program/lib/Roundcube/rcube_utils.php on line 59
```

 lets remove it
`unset($attrib['lifetime']);`

skiner68:
Hello, will this be released in next 1.4.*or should I wait for 1.5? Thank you!

Issue #6312, opened 2018-05-29
Improve Redis/Memcache cache performance


alecpl:
Today we use "index approach" so a normal session looks like this:
```
[1] GET 10:IMAPINDEX
[2] GET 10:IMAP:mailboxes.metadata.*.446be42f5518b0285ccbea93f318d096
[3] SET 10:IMAP:mailboxes.folder-info.Configuration {<data>}
...
[9] SET 10:IMAPINDEX {<data>}
```
This method has pros and cons:
1. We can do less GET requests as we know what's in the cache.
2. We can delete entries by prefix or delete all existing entries easily.
3. There's a big chance of a race-condition with parallel sessions. In such case the index can be overwritten with out-of-date information.

Actually in a system with multiple connections I observed the race-condition issue quite a lot, causing the cache to be not used effectively. So, we should consider a different approach, and do some perf. tests. Let's say:
1. Don't store index entries, don't delete entries (let them to expire)
2. Store update/creation timestamp with the entry data and
3. Compare that timestamp with "entries expiration timestamp", if entry is older it will be considered non-existing
We need "entries expiration timestamp" for "by prefix" and "all" entries deleting/expiration. So we'd add a special entries that we'd check (instead of the index).

Here's how that could work with examples:
1. Remove all entries in IMAP cache:
```
SET 10:IMAPEXPIRE:* 1527584756
```
2. Remove entries with "mailboxes" prefix:
```
SET 10:IMAPEXPIRE:mailboxes 1527584756
```
So now when we're asked for key `mailboxes.folder-info.Configuration` we'll do:
```
GET 10:IMAP:mailboxes.folder-info.Configuration
GET 10:IMAPEXPIRE:*
GET 10:IMAPEXPIRE:mailboxes -- only if IMAPEXPIRE:*'s timestamp < entry timestamp
GET 10:IMAPEXPIRE:mailboxes.folder-info -- only if IMAPEXPIRE:mailboxes's timestamp < entry timestamp
```
I think it will be best to limit this to max. two "hierarchy" levels. Otherwise we end up with too many GET requests, imo.

Pros and cons:
- we'll do more GET requests, but we expect cache hits than misses now, and
- we'll do less SET requests (and IMAP requests) because data will be removed/expired from cache less often. No race-condition issue, or at least less chances for it.
- this might not be suitable for attachments cache, as we'd like to remove big attachment data as soon as possible (by prefix) instead of waiting for expiration.

alecpl:
[...]

PR #6781, opened 2019-06-01
Cache refactoring (#6312)

alecpl:
A better memcache/redis cache strategy. Implements two mode of operation:
1. Indexed cache - the old solution, stores the list of keys in an index entry, used e.g. by attachment plugins.
2. Non-indexed cache - we do not store list of keys, instead we store timestamp of key creation (with the value) and timestamp of a last key deletion request. Then we compare timestamps when reading the key.


fabianbur:
We look forward to this feature.

PR #6796, opened 2019-06-09
Elastic: some visual polish-up

thomascube:
The changes include:
- Add/increase whitespace between UI elements (variables)
- Adjust font sizes for some elements
- Darker main color to increase contrast
- Compse screen: move attachments panel to the right
- Draw box around attachments list
- Distinct icons for settings and options
- Consistent icon to enable select checkboxes
- Pen icon for floating compose button

With many thanks to mittwald.de for the review and suggestions.

### DIFF
--- a/plugins/enigma/skins/elastic/templates/keys.html
+++ b/plugins/enigma/skins/elastic/templates/keys.html
@@ -6,12 +6,12 @@
 
 <!-- keys list -->
 <div id="layout-list" class="listbox selected" aria-labelledby="aria-label-enigmakeyslist">
-	<div class="header layout-box">
+	<div class="header">
 		<a class="button icon back-sidebar-button" href="#sidebar"><span class="inner"><roundcube:label name="settings" /></span></a>
 		<span id="aria-label-enigmakeyslist" class="header-title"><roundcube:label name="enigma.enigmakeys" /></span>
 		<a class="button icon toolbar-menu-button" href="#list-menu"><span class="inner"><roundcube:label name="menu" /></span></a>
 	</div>
-	<roundcube:object name="searchform" id="keysearch" wrapper="searchbar menu layout-box"
+	<roundcube:object name="searchform" id="keysearch" wrapper="searchbar menu"
 		label="keysearchform" buttontitle="findkeys" ariatag="h2" />
 	<div class="scroller">
 		<roundcube:object name="keyslist" id="keys-table" class="listing" role="listbox" noheader="true"
@@ -36,7 +36,7 @@
 <!-- key info frame -->
 <div id="layout-content" role="main">
 	<h2 id="aria-label-toolbar" class="voice"><roundcube:label name="arialabeltoolbar" /></h2>
-	<div class="header layout-box" role="toolbar" aria-labelledby="aria-label-toolbar">
+	<div class="header" role="toolbar" aria-labelledby="aria-label-toolbar">
 		<a class="button icon back-list-button" href="#back"><span class="inner"><roundcube:label name="back" /></span></a>
 		<span class="header-title"></span>
 		<!-- toolbar -->
--- a/plugins/enigma/skins/elastic/templates/keys.html
+++ b/plugins/enigma/skins/elastic/templates/keys.html
@@ -6,12 +6,12 @@
 
 <!-- keys list -->
 <div id="layout-list" class="listbox selected" aria-labelledby="aria-label-enigmakeyslist">
-	<div class="header">
+	<div class="header layout-box">
 		<a class="button icon back-sidebar-button" href="#sidebar"><span class="inner"><roundcube:label name="settings" /></span></a>
 		<span id="aria-label-enigmakeyslist" class="header-title"><roundcube:label name="enigma.enigmakeys" /></span>
 		<a class="button icon toolbar-menu-button" href="#list-menu"><span class="inner"><roundcube:label name="menu" /></span></a>
 	</div>
-	<roundcube:object name="searchform" id="keysearch" wrapper="searchbar menu"
+	<roundcube:object name="searchform" id="keysearch" wrapper="searchbar menu layout-box"
 		label="keysearchform" buttontitle="findkeys" ariatag="h2" />
 	<div class="scroller">
 		<roundcube:object name="keyslist" id="keys-table" class="listing" role="listbox" noheader="true"
@@ -36,7 +36,7 @@
 <!-- key info frame -->
 <div id="layout-content" role="main">
 	<h2 id="aria-label-toolbar" class="voice"><roundcube:label name="arialabeltoolbar" /></h2>
-	<div class="header" role="toolbar" aria-labelledby="aria-label-toolbar">
+	<div class="header layout-box" role="toolbar" aria-labelledby="aria-label-toolbar">
 		<a class="button icon back-list-button" href="#back"><span class="inner"><roundcube:label name="back" /></span></a>
 		<span class="header-title"></span>
 		<!-- toolbar -->
--- a/plugins/help/skins/elastic/templates/help.html
+++ b/plugins/help/skins/elastic/templates/help.html
@@ -2,7 +2,7 @@
 <roundcube:include file="includes/menu.html" />
 
 <div id="layout-sidebar" class="listbox" role="navigation" aria-labelledby="aria-label-helpnav">
-	<div class="header">
+	<div class="header layout-box">
 		<a class="button icon task-menu-button" href="#menu"><span class="inner"><roundcube:label name="menu" /></span></a>
 		<span id="aria-label-helpnav" class="header-title"><roundcube:label name="help.help" /></span>
 	</div>
@@ -19,7 +19,7 @@
 <h1 class="voice"><roundcube:label name="help.help" /></h1>
 
 <div id="layout-content" class="selected" role="main">
-	<div class="header">
+	<div class="header layout-box">
 		<a class="button icon task-menu-button" href="#menu"><span class="inner"><roundcube:label name="menu" /></span></a>
 		<span class="header-title"><roundcube:label name="help.help" /></span>
 	</div>
--- a/plugins/help/skins/elastic/templates/help.html
+++ b/plugins/help/skins/elastic/templates/help.html
@@ -2,7 +2,7 @@
 <roundcube:include file="includes/menu.html" />
 
 <div id="layout-sidebar" class="listbox" role="navigation" aria-labelledby="aria-label-helpnav">
-	<div class="header layout-box">
+	<div class="header">
 		<a class="button icon task-menu-button" href="#menu"><span class="inner"><roundcube:label name="menu" /></span></a>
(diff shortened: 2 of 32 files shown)


alecpl:
Some really nice changes there. However, things I don't like or are just bugs:
1. that margin-right on buttons in toolbar, there should be no margin (in desktop mode). Especially in Mail we have many buttons in toolbar (consider markasjunk and archive), but even with the default set it looks kind of scattered.
2. Inconsistent left and right padding in searchbar. I understand you moved the search icon to match the padding and icons below and above, but... the options icon has different padding. Also, on messages list you don't really have left-icons on list records.  Also, on mobile the searchbar icons are not aligned with anything. It looks like that lists padding changes break more than improve.
3. different main (blue) color for logo/taskmenu and the rest
4. I'm not sure I  like padding in popup menus. And it is inconsistent e.g. with padding of folder selector popup.
5. Font-size in mobile header is too big. Sounds like a personal preference, but really it does not have to be that big. `1.2rem` is max. (in sync with font side on lists). `1rem` is my preference.
6. `.layout-box` class - I'd definitely want it to be removed and assigned padding to elements that need it in styles (these are mostly only headers and some menus which are generally part of the layout which should not require additional classes, at least that's my idea of the main structure).
7. With small screen width the width is precious (tablet or mobile with taskmenu still visible), I think taskmenu could be using smaller width.
8. Inconsistent padding in taskmenu and other menus on mobile.
9. On messages list td.flags and td.subject (in threaded mode) does not have to have such big left padding.
10. Visit Mail or Contacts with screen width ~1080-768px. See padding issue on icons in the list header.
11. Padding on lists is inconsistent in different screen sizes. In some sizes I see the new (bigger) padding in some I see none. E.g. the searchbar icon has three different positions when you resize the screen from desktop to mobile width.
12. Attachments list box style is good, but do we need it in mail compose?
13. Mobile menu records' height was consistent with other lists, now it is not. Also font-size.

thomascube:
> Some really nice changes there. However, things I don't like or are just bugs:

First of all I'd like to mention that the proposed changes are implementations of the suggestions we received from a screen designer who volunteered to review the Elastic theme. While style is always a matter of taste, I think we should trust his professional eye. Also, his perception is much more objective than ours who have been staring at the Elastic screens for months now.

>     1. that margin-right on buttons in toolbar, there should be no margin (in desktop mode). Especially in Mail we have many buttons in toolbar (consider markasjunk and archive), but even with the default set it looks kind of scattered.

This was the first point in the review. The margin helps separating the buttons from each others, especially the carets to open an options menu are now more clear to which button they belong. I think there should be some margin, we can maybe reduce it a bit.

>     2. Inconsistent left and right padding in searchbar. I understand you moved the search icon to match the padding and icons below and above, but... the options icon has different padding. Also, on messages list you don't really have left-icons on list records.

The right margin is 0 in order to have no space between the options button (when hovering).

>  Also, on mobile the searchbar icons are not aligned with anything. It looks like that lists padding changes break more than improve.

You are right, that should be corrected.

>     3. different main (blue) color for logo/taskmenu and the rest

The adjustment was made to increase the contrast of the primary buttons and text links. The color is different on white and dark background for just that reason. And who says that it has to be the same color as in the logo?

>     4. I'm not sure I  like padding in popup menus. And it is inconsistent e.g. with padding of folder selector popup.

I don't understand what you mean here.

>     5. Font-size in mobile header is too big. Sounds like a personal preference, but really it does not have to be that big. `1.2rem` is max. (in sync with font side on lists). `1rem` is my preference.

I increased it because the text looked tiny inside that header and besides the rather large icon buttons left and right. We can agree on 1.2rem or consider making the icons a bit smaller or even the entire header. But I strongly suggest some approximation of the header elements.

>     6. `.layout-box` class - I'd definitely want it to be removed and assigned padding to elements that need it in styles (these are mostly only headers and some menus which are generally part of the layout which should not require additional classes, at least that's my idea of the main structure).

OK.

>     7. With small screen width the width is precious (tablet or mobile with taskmenu still visible), I think taskmenu could be using smaller width.

This is the main navigation and therefore I think it deserves the additional 6px I gave it.

>     8. Inconsistent padding in taskmenu and other menus on mobile.

That's not intentional and should be fixed.

>     9. On messages list td.flags and td.subject (in threaded mode) does not have to have such big left padding.

It's just the same padding as in other lists. You desire consistency, right?

>     10. Visit Mail or Contacts with screen width ~1080-768px. See padding issue on icons in the list header.

I see. Will fix this.

>     11. Padding on lists is inconsistent in different screen sizes. In some sizes I see the new (bigger) padding in some I see none. E.g. the searchbar icon has three different positions when you resize the screen from desktop to mobile width.

This is because we have three classes of screen sizes we're addressing. Don't compare them to each others while resizing the window but look at each individually. The paddings should reflect the available space we have in the three different layouts (1,, 2 or 3 columns). The tablet view with 2 columns certainly is the trickiest one because we squeeze in a lot of content. Therefore the padding is small here. In mobile view with only one column we can again afford a bit more space in order to not have all elements just sticking right at the border of the screen.

>     12. Attachments list box style is good, but do we need it in mail compose?

Maybe it doesn't but I think it's OK.

>     13. Mobile menu records' height was consistent with other lists, now it is not. Also font-size.

Feedback from the designer and I also compared the menu font size with other mobile apps (e.g. Gmail) which all use smaller fonts.

I'll follow up with commits fixing the obvious bugs mentioned here.

alecpl:
Please, note that we have had also other designers looking at this and they didn't come with the same suggestions. Also, I saw that designer feedback and to me it wasn't clear what he means exactly.
Now, replying to your responses... ;)
1. There's padding already, maybe make it very lightly bigger, but really I think toolbar buttons have enough spacing.
2. And you're happy with the result?
3. I've been learned less colors is better. Of course we use shades/derivatives, but to me these two blue colors does not look like coming from the same palette.
4. In Mail goto More... > Copy to.... Now you see two popup menus with different padding.
5. Note that we display long text there sometimes (folder name) or add sub-text (calendar timezone). I'm checking with real phone and for me font-size as well as buttons and the header element sizing is ok. I made sure to work with real device when choosing layout element sizes.
9. I desire consistency, but this is the "record internal structure". Other lists do not have this. And internal structure can have smaller paddings to display more information.
11. I meant that the padding is some elements is not consistent with other elements using the same screen size.
13. If the size is ok for the menu then it should be ok for other lists and should be made consistent. I choose the record height using my mobile and my fingers, but I could agree to a couple pixels less, but make it consistent.

thomascube:
> Please, note that we have had also other designers looking at this and they didn't come with the same suggestions. Also, I saw that designer feedback and to me it wasn't clear what he means exactly.
> Now, replying to your responses... ;)
> 
>     1. There's padding already, maybe make it very lightly bigger, but really I think toolbar buttons have enough spacing.

Unfortuntely I disagree on this.

>     2. And you're happy with the result?

Yes.

>     3. I've been learned less colors is better. Of course we use shades/derivatives, but to me these two blue colors does not look like coming from the same palette.

We have very few colors on our palette. We might indeed try to adjust those two colors but I hope you agree that the very same color doesn't show the same contrast on both white and dark backgrounds. And for links and primary buttons contrast and visibility is essential.

>     4. In Mail goto More... > Copy to.... Now you see two popup menus with different padding.

The paddings of the folder list are set by code to represent their indentation level. I don't know how to overcome that issue in the best way.

>     5. Note that we display long text there sometimes (folder name) or add sub-text (calendar timezone). I'm checking with real phone and for me font-size as well as buttons and the header element sizing is ok. I made sure to work with real device when choosing layout element sizes.

Same here. I tested on an iOS device and the text really looks ridiculously small between those icons.

Photo 19-06-13 14-51-17 2265

>     6. I desire consistency, but this is the "record internal structure". Other lists do not have this. And internal structure can have smaller paddings to display more information.
> 
>     7. I meant that the padding is some elements is not consistent with other elements using the same screen size.

That's certainly not desired and I'm trying to fix this.

>     8. If the size is ok for the menu then it should be ok for other lists and should be made consistent. I choose the record height using my mobile and my fingers, but I could agree to a couple pixels less, but make it consistent.

I'm on it.

alecpl:
Here's some more on issues in big screen mode:
14. That wider taskmenu does not make sense. There are text-labels that have different length (also it depends on locale). So, setting a padding for one does not really work for others.
14a. Taskmenu items have broken padding. Different on top and bottom. I mean visually, in css the top and bottom padding is the same but the icon position makes it look wrong.
15. The padding is still not consistent. E.g. the iframe padding (on `.formcontent`) is 18.2px, while the list record padding (on `.listing li a`) is 14px. Even if you add 2px for the left border focus indicator it's still not the same.
16. This different padding on left and top really does not look good to me:
padding
16a. This bigger left-padding is definitely not needed in popup menus. They are "separated elements" and don't need to have the same padding. I.e. the record icon should have the same padding on left and top and bottom. This way it looks best in menus especially.
16b. Can't we make that left-padding the same as the distance between button label and bottom of the toolbar? Which is about 12px. Minus 2px it is 10px for padding-left on list records. I checked that and it looks much better.
17. That margin-right for toolbar elements is problematic also for one reason. There are elements that are hidden in desktop mode (e.g. `select menulink hidden-big hidden-large active`), but the still have that margin. That can be fixed by applying the hidden-* classes to parent element, but I really thing the margin should be removed (maybe increasing padding instead).
Also, that margin on li.spacer is kind of redundant.
18. `li` elements in toolbar have height too small. It makes a visual issue with hover effect.
19. The padding on messages list, considering the expand/collapse icon is really bad. As you can see on the folders list it looks better if these icons does not have an extra padding on right.
20. These new submenu indicators (right-angle icons) have two issues. They should be moved more to the right and they should become white on hover.
21. Still don't like the searchbar padding. Please, enter something into the search input to see the Cancel button. Now, these buttons imo should be close to each other and have the same size. (Note I may add a third button there in the future).
22. Broken "Export selected" icon in Contacts toolbar.

Now tablet mode looks not that bad.

23. In tablet mode (1024px) the right-side buttons should be closer to the right side.
tablet-issue
Also that menu button does not fit the searchbar options button position below.
24. Paddings in tablet mode didn't change much. Because of that I think that iframe padding might be too big there as it does not match with anything.

On mobile:

25. Toolbar icons should be aligned with list icons below:
mobile-issue
26. Iframe padding (e.g. `.formcontent`) should be smaller. The space is precious here. ... Actually I see that the extra padding on `html.iframe .formcontent > .propform` makes it too big. See Preferences iframe. The same for `.popupmenu.propform`, see messages list Options dialog.
27. Expand icon on messages list is not aligned vertically to the center.

thomascube:
I give up on this.

alecpl:
I spend almost 2 years polishing positions and sizes in Elastic, especially considering all these dependencies between various elements on the page. I knew it will not be simple to do changes in that regard. I hope that you at least see most of my points valid and not only my subjective opinions.

The designer said spacing is not uniform, but didn't really gave answers on where what. And indeed it's not that easy as you have find out here. If you don't have time to finish that I may try or at least get some good ideas from it.

PR #6798, opened 2019-06-10
Fix nl_NL label shortheaderfrom in dev/elastic

OliverHe:
Found in 1.4 rc1, no idea how it got there. Wasn't there in the 1.4 beta.

Expected: "Van Afzender op een datum"
Actual: "Van **$to** op een datum"


alecpl:
This have to be fixed via Transifex. https://www.transifex.com/roundcube/roundcube-webmail

alecpl:
This is already fixed in Transifex and git master. Closing.

dev/elastic branch is not up to date, we're using it for some development occasionally and could remove it at any time.

PR #6805, opened 2019-06-15
Adding ssha512 password_algorithm

netzding:
To natively support the ssha512 hashing algorithm e.g. used by dovecot I've added a new case for it by copying the existing ssha case and doing some minor adjustments to it.


alecpl:
Looks good, but these two empty lines you removed on copy/paste were actually useful.

netzding:
Added the two lines back in I accidentally dropped on copy pasting.

PR #6811, opened 2019-06-19
Put html meta content type right after the <head>

jfcherng:
Currently, the `<meta http-equiv="content-type" content="text/html; charset=UTF-8" />` is put after `<title>...</title>`. While the title could very possibly be non-ASCII, it should be put right after the `<head>`.

Reference: https://www.w3.org/TR/html4/charset.html#h-5.2.2

> The META declaration must only be used when the character encoding is organized such that ASCII-valued bytes stand for ASCII characters (at least until the META element is parsed). META declarations should appear as early as possible in the HEAD element.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1897,7 +1897,7 @@ EOF;
             $page_header = "<head>\n<title>$page_title</title>\n$page_header\n</head>\n";
         }
 
-        // add page hader
+        // add page header
         if ($hpos) {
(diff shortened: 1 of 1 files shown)


alecpl:
We should also consider removing use of `pagetitle` object from templates and add automatically the title at the end of `meta` object output.

jfcherng:
Closed because of a991e3181cc77ed2efdec7a2edfb88cd4635de02.

PR #6817, opened 2019-06-23
Fix variables defined in variable.less cannot be overwritten

jfcherng:
Variables defined in `variables.less` cannot be overwritten by `_variables.less`.

For example, re-defining `@page-font-size` in `_variables.less` is not working.


alecpl:
dev/elastic branch is old. We acceept pull requests to master only.

PR #6822, opened 2019-06-25
Make Apache prefer to brotli over gzip when possible

jfcherng:
# Problem

I use `brotli` (rather than `gzip`) on my server and I occasionally found that it did not be activated on my RC's site. I soon found that there is a `.htaccess` in RC's root dir which forces Apache to use `gzip`.

# Proposal

`brotli` is a better (than `gzip`) compression algorithm derived by Google. It's available in Apache 2.4.26 and later. Maybe we should prefer `brotli` over `gzip` if it's available?

# References

- https://en.wikipedia.org/wiki/Brotli
- https://httpd.apache.org/docs/2.4/mod/mod_brotli.html

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -19,6 +19,13 @@ RewriteRule /?(README.*|meta\.json|composer\..*|jsdeps.json)$ - [F]
 SetOutputFilter DEFLATE
 </IfModule>
 
+# prefer to brotli over gzip if brotli is available
+<IfModule mod_brotli.c>
+SetOutputFilter BROTLI_COMPRESS
+# some assets have been compressed, so no need to do it again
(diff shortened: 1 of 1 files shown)


alecpl:
Browser support is not good enough, https://caniuse.com/#feat=brotli

jfcherng:
> Browser support is not good enough, https://caniuse.com/#feat=brotli

If the browser not support brotli, it should not claim it supports. So I don't think it's a problem?

Chrome: `accept-encoding: gzip, deflate, br`
IE 11: `Accept-Encoding	gzip, deflate`

alecpl:
Of course, you're right. The only possible issue I see is environments with nginx proxy between user and Roundcube than may have trouble with this. Other than that it's fine, but I'll postpone this to 1.5.

PR #6828, opened 2019-06-29
Fix visibility of rcube_ldap_exop_password::$debug

jfcherng:
`$debug` is `protected` in its parent class `rcube_ldap_simple_password`.

```
PHP Fatal error:  Access level to rcube_ldap_exop_password::$debug must be protected (as in class rcube_ldap_simple_password) or weaker in /var/www/roundcubemail/plugins/password/drivers/ldap_exop.php on line 0
PHP Warning:  Zend OPcache could not compile file /var/www/roundcubemail/plugins/password/drivers/ldap_exop.php in /var/www/roundcubemail/vendor/preload.php on line 47
```


alecpl:
Even better is to remove that line. Fixed.

Issue #6818, opened 2019-06-23
Nested conditioning skin markup behaves weirdly


jfcherng:
Environment
------------

- RC: official git repo here, master branch
- PHP: both 5.6.40 and 7.4.0-dev are tested with the same results

Problem
--------

Since I don't see a note in [Wiki](https://github.com/roundcube/roundcubemail/wiki/Skin-Markup#roundcubeifelseifelseendif) that RC's skin markup does not allow doing this common template feature, I consider this as a bug.

![image](https://user-images.githubusercontent.com/6594915/59975092-b13b9700-95e6-11e9-9a2f-e2e644b48673.png) => outputs `A C` (Good)

![image](https://user-images.githubusercontent.com/6594915/59975070-67eb4780-95e6-11e9-9349-ed2ef1910af1.png) => outputs `D` (**Bad**, should be `B C`)

![image](https://user-images.githubusercontent.com/6594915/59975106-ce706580-95e6-11e9-8876-b4abe7ad68a9.png) => outputs `B C D` (**Bad**, should be `D`)

![image](https://user-images.githubusercontent.com/6594915/59975116-efd15180-95e6-11e9-9688-d6c9368a33f6.png) => outputs `B C D` (**Bad**, should be `D`)

Reproducer
------------

I replace the whole `skins/elastic/templates/login.html` with the following codes and visit the login page to do above tests.

```html
<roundcube:if condition="0" />
	<roundcube:if condition="0" />
		A
	<roundcube:else />
		B
	<roundcube:endif />
		C
<roundcube:else />
	D
<roundcube:endif />
```

alecpl:
This is indeed a limitation of our template engine. We should document that. Fixing this is not trivial and also not a high prio.

PrachiChi:
project

PrachiChi:
how to pull project

vaso123:
This is a must have feature. Maybe that could be an option, if you should tag your condition with a name. 

For example:

```
<roundcube:if condition="config:something_enabled=='1'" name="something" />
Say something here
    <roundcube:if condition="config:anything_enabled=='1'" name="anything" />
        Say anything here
    <roundcube:endif name="anything" />
    More something
<roundcube:endif name="something" />
```
It is insane that now I feel this template engine is good for nothing, and I always should use objects, mixing languages and do the HTML output from PHP.

vaso123:
[...]

PR #6829, opened 2019-06-29
add support for nested if conditions

johndoh:
possible solution for #6818


alecpl:
It would be good to add some tests.

jfcherng:
LGTM in my use cases.

alecpl:
Please, move tests to tests/Rcmail directory (just created in master).

PR #6839, opened 2019-07-11
Elastic: small improvements to compose screen

thomascube:
- Move attachments panel to the right
- Use pen icon for floating compose button


alecpl:
Thank you for this work. I had no time to do this by my own as I promised, but I will make sure these PRs land in master asap.

This one is merged with a small fix.

PR #6851, opened 2019-07-16
Fix the initial SQL schemas to support ES-419 language.

drrtuy:
I got SQL error whilst trying to set Spanish(Latin America) language in the user preferences. It boils down to the width of the language field that is 5 in all schemas thus isn't big enough to accomodate the code 'es-419'.
The one approach is to remove es-419 at all from here:
```
[drrtuy@intmacsta roundcubemail]$ ag -a 'es_419'
program/localization/index.inc
99:  'es_419' => 'Spanish (Latin America)',
```

Another is to resize language field in all schemas. I picked the second approach b/c I need this language. 
I need to alter the field type to varchar(6) that seems to be relatively lightweight operation in all supported databases also and  thus I have a question on schema migration files naming. How should I name incremental migration files?

### DIFF
--- a/SQL/mssql.initial.sql
+++ b/SQL/mssql.initial.sql
@@ -102,14 +102,14 @@ CREATE TABLE [dbo].[users] (
 	[last_login] [datetime] NULL ,
 	[failed_login] [datetime] NULL ,
 	[failed_login_counter] [int] NULL ,
-	[language] [varchar] (5) COLLATE Latin1_General_CI_AI NULL ,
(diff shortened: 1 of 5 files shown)


alecpl:
The name should be in format YYYYMMDDXX (where XX is a number indicating order if there's more than one update for specified day). So, you can use 2019071700. Don't forget to update the roundcube-version string to the same value at the end of *.initial.sql files

alecpl:
And this should target git-master.

alecpl:
Fixed in 50d6ea30e0d

PR #6852, opened 2019-07-16
Elastic: icons cleanup

thomascube:
- Distinct icons for settings and options
- Distinct icons for expand actions (threads)
- Consistent icon to enable select checkboxes

### DIFF
--- a/skins/elastic/styles/widgets/buttons.less
+++ b/skins/elastic/styles/widgets/buttons.less
@@ -114,15 +114,18 @@ button.btn {
     &.folders:before {
         content: @fa-var-folder-open;
     }
+    &.options:before {
+        content: @fa-var-sliders-h;
+    }
     &.tools:before,
     &.settings:before {
-        content: @fa-var-wrench;
+        content: @fa-var-cog;
     }
     &.properties:before {
         content: @fa-var-info-circle;
     }
     &.select:before {
-        .font-icon-regular(@fa-var-check-circle);
+        .font-icon-regular(@fa-var-check-square);
     }
     &.insert.recipient:before {
         content: @fa-var-user-plus;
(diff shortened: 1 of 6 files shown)


alecpl:
1. This breaks "Export selected" icon in Contacts toolbar menu because it uses `.selection` class that changed font-weight in your PR.
2. I'm not sure I like changes in Threads menu, but it is not a strong objection. Also, the "Expand unread" should probably use the same icon as "Unread" option in selection menu.
3. What do you think about changing the "Invert" icon in selection menu to https://fontawesome.com/icons/check-square?style=solid ? I.e. it would be invertion of the Selection icon. Other ideas?

thomascube:
>     1. This breaks "Export selected" icon in Contacts toolbar menu because it uses `.selection` class that changed font-weight in your PR.

I'll check on that.

>     2. I'm not sure I like changes in Threads menu, but it is not a strong objection. Also, the "Expand unread" should probably use the same icon as "Unread" option in selection menu.

I think it's still better than having the same icon for three different actions.

>     3. What do you think about changing the "Invert" icon in selection menu to https://fontawesome.com/icons/check-square?style=solid ? I.e. it would be invertion of the Selection icon. Other ideas?

I don't think this icon would be an improvement but unfortunately I don't have a better suggestion.

thomascube:
@alecpl Do you think this PR is eligible to be merged with the fix for the export selected menu icon?

alecpl:
Sorry for the delay, I was on my vacation. We can ignore point 3, and 1. is done, but for 2. I'm not sure. the "Expand unread" icon is inconsistent. Also, if you see Reply all or Forward or Download menu, you'll see that it's not that unusual to have the same icon for all options. Personally I prefer this over use of some possibly not clear icons.

thomascube:
@alecpl Icon for "expand unread" fixed. Using the same icon for different actions more seems like a bug to me. If we can't find distinct icons, we better do not show any at all.

PR #6860, opened 2019-07-20
add expand/collapse all treelist methods

johndoh:
add methods to the rcube_treelist_widget JS class to support expanding/collapsing all nodes in single action

note: I removed the "private methods" as it makes no sense to me, not all the method are private and similar comments are not used in other JS classes

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -1776,6 +1776,9 @@ function rcube_webmail()
 
   this.folder_collapsed = function(node)
   {
+    if (this.folder_collapsed_timer)
+      clearTimeout(this.folder_collapsed_timer);
+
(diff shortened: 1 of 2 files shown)


johndoh:
The second commit is to fix a problem when saving the list state back to user prefs.

The issue appears to be when the collapse/expand treelist event is triggered multiple times near simultaneously then the code in `folder_collapsed()` in app.js cannot keep up.

Happens in current code when using the `collapse()` method with the recursive param.

alecpl:
I don't like `set` argument in `collapse_all()`. Please make both these methods without arguments.

johndoh:
done in latest commit

PR #6866, opened 2019-07-24
Fix PHP 7.4 deprecation: implode() wrong parameter order

jfcherng:
It has been [deprecated in PHP 7.4](https://wiki.php.net/rfc/deprecations_php_7_4#implode_parameter_order_mix).

Such as PHP deprecated:  implode(): Passing glue string after array is deprecated. Swap the parameters in `/var/www/roundcubemail/program/lib/Roundcube/rcube_db.php` on line 917

NOTE: This PR should probably also be applied to other branches that are still under maintain since formally speaking this is a wrong usage of `implode()` imo.

### DIFF
--- a/bin/installto.sh
+++ b/bin/installto.sh
@@ -130,7 +130,7 @@ if (strtolower($input) == 'y') {
     }
 
     if (!empty($adds)) {
-        echo implode($adds, "\n") . "\n\n";
+        echo implode("\n", $adds) . "\n\n";
     }
 
     echo "Running update script at target...\n";
--- a/plugins/enigma/lib/enigma_ui.php
+++ b/plugins/enigma/lib/enigma_ui.php
@@ -752,7 +752,7 @@ class enigma_ui
         }
 
         $table->add('title', html::label('key-name', rcube::Q($this->enigma->gettext('newkeyident'))));
-        $table->add(null, html::tag('ul', 'proplist', implode($identities, "\n")));
+        $table->add(null, html::tag('ul', 'proplist', implode("\n", $identities)));
 
         // Key size
         $select = new html_select(array('name' => 'size', 'id' => 'key-size'));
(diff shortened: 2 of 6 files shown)


navossoc:
My address book search is not working anymore after updating to 1.3.10.
I think this can be the issue. I'm using php 7.3.9.

jfcherng:
That would be interesting since what this PR done is simple and solid. :confused:

alecpl:
Works for me. So, you'll have to be more specific. And better to create a ticket than comment on an irrelevant commit. What you describe sounds like the issue fixed in 1.3.10, maybe you did upgrade improperly.

LossAngeles:
I recently upgraded Ubuntu 18.x to 20.x which included a Roundcube package upgrade from 1.3 to 1.4. From what I can tell this bug looks fixed. Was this fix part of the 1.4 package? Is this an error I can ignore? Appreciate any guidance. 

roundcube: PHP Deprecated:  implode(): Passing glue string after array is deprecated. Swap the parameters in /var/www/sites/roundcube/program/lib/Roundcube/rcube_db.php on line 917

jfcherng:
> Was this fix part of the 1.4 package?

The first release included this PR is `1.4-rc2`.

LossAngeles:
Hi 

> > Was this fix part of the 1.4 package?
> 
> The first release included this PR is `1.4-rc2`.

Hi and thanks for helping out. Because this fix was written well over a year ago I expected it to be in this Ubuntu LTS release that 20.x installed : 

Installed version | 1.4.3+dfsg.1-1
-- | --


Is PR 1.4-rc2 part of the above installed package? If it is why would my logs be hit by this error? Thanks again. -Greg

jfcherng:
> Is PR 1.4-rc2 part of the above installed package? If it is why would my logs be hit by this error? Thanks again. -Greg

It should be. And that file in https://launchpad.net/debian/+archive/primary/+sourcefiles/roundcube/1.4.3+dfsg.1-1/roundcube_1.4.3+dfsg.1.orig.tar.xz looks good to me.

LossAngeles:
Well chit lol .... that bug is definitely in the package Ubuntu 20,x is delivering on Google Cloud. As a work around I can keep my manual edit and look for a new release down the road. I have no interest in figuring out why that package has a stale file. Cheers =)

PR #6868, opened 2019-07-24
Elastic: Add indicator for popover menu items that open a submenu

thomascube:

### DIFF
--- a/skins/elastic/styles/widgets/menu.less
+++ b/skins/elastic/styles/widgets/menu.less
@@ -52,6 +52,10 @@
                     margin-right: .5em;
                 }
             }
+
+            &:hover:after {
+                color: @color-menu-hover;
+            }
         }
     }
 
--- a/skins/elastic/styles/widgets/menu.less
+++ b/skins/elastic/styles/widgets/menu.less
@@ -35,6 +35,27 @@
         }
     }
 
+    .popover & {
+        a.menulink {
+            position: realtive;
+
+            &:after {
+                .font-icon-class();
+                content: @fa-var-chevron-right;
+                float: right;
+                width: 1em;
+                margin-right: 0;
+                color: @color-black-shade-text;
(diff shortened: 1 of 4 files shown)


johndoh:
@thomascube there is something wrong with the `:hover` styling on the updated items, the font and background colours do not change the same as they do for the existing items like forward or reply-all.

thomascube:
@johndoh Hover color fixed in recent commit. However, these icons are not the same as the ones for forward or reply.

alecpl:
There's some more work needed. 
1. These new elements must look the same (on mobile) as `a.dropdown`. Same icon and color. On the other hand they are slightly different in functionality. Maybe `a.dropdown` should have some left-border-like style to "separate" it from the main button action.
2. I'm not sure doing this by a class on `a` element is the best approach. There's at least one case where this does not work (will require some javascript). See what zipdownload plugin does with the "Download" action. It adds a submenu, so it should just add `menulink` class to the link element. However, this is not enough because when the link action changes state the class is lost. We would need to re-register the button. Maybe we could/should use `[aria-haspopup=true]` selector instead of a class?

@thomascube what do you think? I could do the work, but does it make sense?

thomascube:
It's actually on purpose that the indicators do not look the same as the already existing dropdown icons because they are not the same. But I admit that the visual distinction might be too vague. The idea with a border is good. We might also consider to choose a different icon for the dropdown link, e.g. three dots denoting "more options" which is actually what that particular link does: it opens a submenu with more options related to the according action item.

Indeed, using `aria-haspopup` as selector is more generic.

alecpl:
I gave it a try. A border alone didn't look good enough, so I tried background color and different icons.

menus

I think I prefer the first from left.

thomascube:
@alecpl I have the same preference. Nice!

alecpl:
Done in bdd1b2054f4.

PR #6894, opened 2019-08-20
Line spacing deleted

daynka:
Messages can be written without default line spacing

### DIFF
--- a/program/js/editor.js
+++ b/program/js/editor.js
@@ -56,6 +56,7 @@ function rcube_text_editor(config, id)
       paste_webkit_style: "color font-size font-family",
       paste_data_images: true,
       browser_spellcheck: true,
+      forced_root_block: "<br>",
       anchor_bottom: false,
       anchor_top: false
     };


alecpl:
Could you elaborate? The TinyMCE documentation says: "Not using p elements as root block can severely cripple the functionality of the editor." So, we should be careful with changes like this.

daynka:
Hello Aleksander Machniak, 

i've edited the source code on my server for a few months ago. On my
server about 50 users using roundcube on a regular basis. The users had
so far with the change in the source code neither problems or crashes. I
unserstand that must act cautiously due to the emergency of ThinyMCE who
needs. ThinyMCE also writes that it has already worked in a previous
version so adapted and only from version 3.0 the default was changed. I
would very much in favor out the adaptation, since there is no
possibility of adjusting the line spacing in the integrated ThinyMCE
editor. With the <br> element, the line spacing can be defined more
flexibly. In addition, all other email providers also work with the <br>
element.

-- 

DANIEL GRETZ

Am 21.08.2019 13:28, schrieb Aleksander Machniak:

> Could you elaborate? The TinyMCE documentation says: "Not using p elements as root block can severely cripple the functionality of the editor." So, we should be careful with changes like this. 
> 
> --
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1]
https://github.com/roundcube/roundcubemail/pull/6894?email_source=notifications&amp;email_token=AALNIFXO7EMBP67WPKFL6OTQFURFVA5CNFSM4IN52F22YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD4ZLCDI#issuecomment-523415821
[2]

alecpl:
You can modify the editor config using a Roundcube plugin (`editor-init` hook). A simple plugin that does this, but using a different (older) technique you can find at https://git.kolab.org/diffusion/RPK/browse/master/plugins/tinymce_config/tinymce_config.php

As a core feature we don't want this.

PR #6908, opened 2019-08-31
a few tweaks to elastic popup menus

johndoh:
There are a couple of edge cases with the popup menus that I have tried to improve on with this PR. This is based on some CSS from the Contextmenu plugin.

1) If you have a new menu item which does not have an icon then it does not align correctly. For this I have added a blank placeholder icon
2) Wrapping of long menu items with popups. Currently the popup chevron disappears and the height of the item is increased. For this I reworking the existing CSS a little to prevent the line wrapping
3) I added min-widths to the before/after to prevent browsers from squashing them when there is long text and messing up the alignment

One extra thing I did which I think is more personal choice than anything else, there is a blank placeholder on the right (for the chevron) as well as the left now. I think it looks nicer than having the text go right up to the right hand edge but I'm happy to remove it if having the maximum possible space for text is more important

I attach a before and after image for reference.

menu

### DIFF
--- a/skins/elastic/styles/widgets/menu.less
+++ b/skins/elastic/styles/widgets/menu.less
@@ -24,6 +24,10 @@
             &:extend(.font-icon-class);
         }
 
+        &:not(.no-icon):before {
+            content: "\00a0"; // blank place holder
+        }
+
         &:hover,
         &:focus {
             outline: 0;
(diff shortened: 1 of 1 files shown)


alecpl:
1. I prefer no blank placeholder on right.
2. Maybe add a class for cases when someone wants a menu with all options with no icons (the blank icon is unwanted).
3. I didn't check, so please verify if folder selector menu (Copy To/Move To) is not broken by this change

johndoh:
> 1. I prefer no blank placeholder on right.

I've removed it

> 2. Maybe add a class for cases when someone wants a menu with all options with no icons (the blank icon is unwanted).

Added class `no-icon` to remove the blank icon

> 3. I didn't check, so please verify if folder selector menu (Copy To/Move To) is not broken by this change

I confirm the folder selector menu is not broken by this change

PR #6913, opened 2019-09-05
Adjust the #mainscreen bottom margin

leofeyer:
This PR decreases the bottom margin from 20px to 10px, so the content area can be rendered slightly higher. The layout still looks good, because now the bottom margin matches the left and right margin.

I am aware that this is an opinionated change, so feel free to close the PR if you do not agree. 😄


alecpl:
We're with this margin for quite a long time.

PR #6950, opened 2019-10-04
managesieve: Add support for 'spamtest' extension

bubble75:
Only RFC3685 for now with slightly correct emulation of what RFC5235 offers.


alecpl:
Nice. We'll need some tests for the parser.

bubble75:
I've added a snippet, but I'm not familiar with PHP unit testing, sorry.

PR #6952, opened 2019-10-05
improve JS Edge version detection, add support for Edge Chromium

johndoh:
currently the JS browser version detection for Edge is not working quite right as it gets confused by other browser engines mentioned in the user agent. eg:

`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36 Edge/18.18938`

is detected as version `537.36` rather than `18.18938`. This fixes that.

also in the Edge Chromium the user agent string has changed slightly to:

`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3763.0 Safari/537.36 Edg/75.0.131.0`

so i've added support for that as well.

### DIFF
--- a/program/js/common.js
+++ b/program/js/common.js
@@ -58,7 +58,7 @@ function roundcube_browser()
   this.dom = document.getElementById ? true : false;
   this.dom2 = document.addEventListener && document.removeEventListener;
 
-  this.edge = this.agent_lc.indexOf(' edge/') > 0;
+  this.edge = this.agent_lc.indexOf(' edge/') > 0 || this.agent_lc.indexOf(' edg/') > 0; // "edg" in Chromium based Edge
   this.webkit = !this.edge && this.agent_lc.indexOf('applewebkit') > 0;
   this.ie = (document.all && !window.opera) || (this.win && this.agent_lc.indexOf('trident/') > 0);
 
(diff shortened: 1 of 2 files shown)


program/js/common.js
```diff
@@ -77,8 +77,14 @@ function roundcube_browser()
   }
 
   if (!this.vendver) {
-    // common version strings
-    this.vendver = /(opera|opr|khtml|chrome|safari|applewebkit|msie)(\s|\/)([0-9\.]+)/.test(this.agent_lc) ? parseFloat(RegExp.$3) : 0;
```
alecpl: Wouldn't just `/(edge?|opera|...` work here?

alecpl (commented):
Question is if we still want to detect Edge Chromium as Edge not Chrome ;)

program/js/common.js
johndoh: I tired that but for me it always picks up the `applewebkit` hit. may be I'm doing something wrong

johndoh:
> Question is if we still want to detect Edge Chromium as Edge not Chrome ;)

a fair question. i guess this is a similar case to how safari is handled. where safari is set in JS but the class `chrome` is added to the html. the same could be done for Edge

program/js/common.js
alecpl: Ah, I think it takes the longest match not the first match.

alecpl: or first match looking at the input not the regexp

johndoh:
Second attempt. I've rewritten the browser version code as I found errors in multiple browsers with the current code. new code is tested with latest user-agents from:

* IE10
* IE11
* Edge
* Edge Chromium
* Opera
* Konqueror
* Safari
* Chrome
* Firefox

I've also set the html class to `chrome` for Edge-Chromium as I think this makes the most sense for CSS.

Issue #6981, opened 2019-10-22
[1.4rc2] get_template_logo() adds unnessecary ":" to search pattern


MASHtm:
Using a config like this
```
$config['skin_logo'] = array("*" => "/images/roundcube_logo.png",
                             "elastic:*" => "/images/favicon.png",
                             "elastic:small" => "/images/logo.svg",
                             ":favicon" => "/images/favicon-uni.png",
);
```
fails for "elastic:small" because the search pattern is built with
```
        $template_names = array(
            $this->skin_name . ':' . $name,
            $this->skin_name . ':*',
            $name,
            '*',
        );
```
But `$name` already contains the leading `':'` in all the routines searching skin_logo. This patch fixes it
```
--- program/include/rcmail_output_html.php.orig	2019-10-22 15:33:57.867444609 +0200
+++ program/include/rcmail_output_html.php	2019-10-22 15:34:18.807756733 +0200
@@ -2464,7 +2464,7 @@
         }
 
         $template_names = array(
-            $this->skin_name . ':' . $name,
+            $this->skin_name . $name,
             $this->skin_name . ':*',
             $name,
             '*',
```
at least for me.

johndoh:
The correct config would be like this:

```php
$config['skin_logo'] = array(
  "*" => "/images/roundcube_logo.png",
  "elastic:*" => "/images/favicon.png",
  ":small" => "/images/logo.svg",
  ":favicon" => "/images/favicon-uni.png",
);
```

or this (though currently only the Elastic skin support the `:small` option)

```php
$config['skin_logo'] = array(
  "*" => "/images/roundcube_logo.png",
  "elastic:*" => "/images/favicon.png",
  "elastic::small" => "/images/logo.svg",
  ":favicon" => "/images/favicon-uni.png",
);
```

MASHtm:
then this needs to be documented -vvv. defaults.inc.php didn't do the job obviously.

alecpl:
We have now three special keywords (`print`, `favicon`, `small`) which could be used alone but also with a skin name prefix. I don't like the double colon syntax. I also think that the colon prefix is redundant.  The chances we ever add another keyword or there's a skin with such name is low.

So, I propose to get rid of the colon prefix, i.e. `favicon`, `small`, `print` instead of `:favicon`, `:small`, `:print`. Which means with skin prefix `elastic:small`, without skin prefix just `small`.

@johndoh could you create a PR asap (so we have it before the stable release)?

[...]

PR #6983, opened 2019-10-23
rewrite skin_logo config

johndoh:
for #6981 

various improvements to the skin_logo config option. the following keys are now supported:

```
skin:template[type]
skin:template
skin:*[type]
skin:*
skin[type]
*:template[type]
*:template
template[type]
template
*[type]
*
[type]
```

also I've improved the documentation in `defaults.inc.php`

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1353,7 +1353,7 @@ EOF;
                 else if ($object == 'logo') {
                     $attrib += array('alt' => $this->xml_command(array('', 'object', 'name="productname"')));
 
-                    if (!empty($attrib['type']) && ($template_logo = $this->get_template_logo(':' . $attrib['type'], true)) !== null) {
+                    if (!empty($attrib['type']) && ($template_logo = $this->get_template_logo(null, $attrib['type'], true)) !== null) {
                         $attrib['src'] = $template_logo;
                     }
                     else if (($template_logo = $this->get_template_logo()) !== null) {
@@ -1363,7 +1363,7 @@ EOF;
                     // process alternative logos (eg for Elastic small screen)
                     foreach ($attrib as $key => $value) {
                         if (preg_match('/data-src-(.*)/', $key, $matches)) {
-                            if (($template_logo = $this->get_template_logo(':' . $matches[1], true)) !== null) {
+                            if (($template_logo = $this->get_template_logo(null, $matches[1], true)) !== null) {
                                 $attrib[$key] = $template_logo;
                             }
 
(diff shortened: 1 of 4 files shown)


tests/Rcmail/OutputHtml.php
```diff
@@ -0,0 +1,142 @@
...
+        $parameters = array('mail', 'small');
+        $result = $get_template_logo->invokeArgs($output, $parameters);
+        $this->assertSame('img00', $result);
+
+        $rcmail->config->set('skin_logo', array(
+             "elastic:login[small]" => "img01",
+             "elastic:login"        => "img02",
+             "elastic:*[small]"     => "img03",
+             "larry:*"              => "img04",
+             "elastic[small]"       => "img05",
```
alecpl: `elastic[small]` is wrong. In this case you look for "elastic" template. This is confusing. I propose to allow skin name prefix with colon only. I.e. setting `[small]` logo in Elastic only would be `elastic:[small]` or `elastic:*[small]`

program/include/rcmail_output_html.php
```diff
@@ -2449,11 +2449,13 @@ protected function about_content($attrib)
...
      * @param string  $name     Name of the logo to check for
      *                          default is current template
+     * @param string  $type     Type of the logo to check for (e.g. 'print' or 'small')
+     *                          default is null (no special type)
      * @param boolean $strict   True if logo should only be returned for specific template
      *
      * @return string image URL
      */
-    protected function get_template_logo($name = null, $strict = false)
+    protected function get_template_logo($name = null, $type = null, $strict = false)
```
alecpl: The $name argument is not used. I propose to get rid of it.

alecpl (changes requested):

program/include/rcmail_output_html.php
johndoh: good point. I'll correct this

tests/Rcmail/OutputHtml.php
johndoh: This PR makes `elastic[small]` and `elastic:*[small]` both valid. My aim was to allow for variations that people might use. but I understand your point may be `elastic:[small]` is more obvious than `elastic[small]`. Or may be all 3?

alecpl: The `elastic[small]` option is confusing, because without colon it looks like `template[type]`. "Skin prefix has always a colon" rule should solve that.

program/include/rcmail_output_html.php
```diff
@@ -2447,35 +2447,45 @@ protected function about_content($attrib)
...
-     * @param string  $name     Name of the logo to check for
-     *                          default is current template
+     * @param string  $type     Type of the logo to check for (e.g. 'print' or 'small')
+     *                          default is null (no special type)
      * @param boolean $strict   True if logo should only be returned for specific template
      *
      * @return string image URL
      */
-    protected function get_template_logo($name = null, $strict = false)
+    protected function get_template_logo($type = null, $strict = false)
```
alecpl: Now, while we're on it. The $strict argument does not make sense to me. Could we remove it?

johndoh: right, any time `$type` is set we can assume a strict match

PR #6993, opened 2019-10-30
added source to image file to deploy on openshift

ombre8:
This file allows Openshift to deploy and install Roundcube automaticly on Openshift.

### DIFF
--- a/.s2i/bin/assemble
+++ b/.s2i/bin/assemble
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# Do Bash Magic to get dot files moved aswell
+shopt -s dotglob
+
+# move the application source
+mv /tmp/src/* $HOME/
+
+cd $HOME/
+
+# get composer
+wget https://getcomposer.org/download/1.9.0/composer.phar
+
+# Install deps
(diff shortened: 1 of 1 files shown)


alecpl:
Thanks, but we'll rather not merge it.

ombre8:
Hi @alecpl 
is there a reason why?
Greetings

alecpl:
If we add something we have to support it. That's why we also remove docker files from the repo.

ombre8:
Ok I get that, so for Documentation purpose, if someone else would like to deploy roundcube on Openshift here is what you can do:

Configure the build config as followed:
```
[...]
    source:
    git:
      ref: release-1.4
      uri: https://github.com/roundcube/roundcubemail.git
    type: Git
  strategy:
    sourceStrategy:
      from:
        kind: ImageStreamTag
        name: php:7.1
        namespace: openshift
      scripts: https://raw.githubusercontent.com/roundcube/roundcubemail/d7b60c4d08ebfc4c7bf97b6126d09fd472df92bf/.s2i/bin/
    type: Source
[...]
```

Issue #7007, opened 2019-11-09
[1.4] Unable to leave sieve filter rule


HLeithner:
While testing final 1.4 I found a problem with manage sieve plugin. Iirc this doesn't exists in 1.4-rc2. 

After selecting a rule from the filters list it's not longer possible to click on "Mail", "Compose", "contacts".... and also the Settings buttons are not working. You can only change the "Actions".

If you click on Save it's possible again to go to mail overview, same is true if you reload the page.

alecpl:
Confirmed.

alecpl:
Fixed.

HLeithner:
Fix confirmed, thank you very much for this great release!

sffax:
For some reason, this fix did not work for me. I replaced: *plugins/managesieve/managesieve.js* with the patched version, restarted the web server, cleared the cache, same behaviour. Does this fix require any file other than the one I referenced above?

alecpl:
Did you remove managesieve.min.js?

sffax:
Apologies, I forgot about that! All good now. Thanks for the help!

PR #7022, opened 2019-11-12
Release 1.4/ plugins for managesieve

intoqh:
Hello,
Today, I used the latest v1.4.0 for the first time. When I configured the managesieve module and deleted a new filter rule, I could not delete it, and the page did not respond. I found that there might be a problem with the program/js/app.min.js file after the operation, but it did not exist in the rc2 version.

by Ricky


alecpl:
See #7007 and don't create dummy pull requests.

intoqh:
OK. 
Then I will be the file: program/js/app.min.js into rc2 later problems disappear, in this, at the same time, I will be two versions of the app.min.js file, found that there are differences. 

THX.

PR #7051, opened 2019-11-19
always update folder count after purge command

johndoh:
Please set the folder message counts regardless of if the user is currently in the folder being purged or not. This is related to https://github.com/johndoh/roundcube-contextmenu/issues/120. The behavior of these 2 JS commands does not change depending on the current folder.

### DIFF
--- a/program/steps/mail/folders.inc
+++ b/program/steps/mail/folders.inc
@@ -59,6 +59,13 @@ else if ($RCMAIL->action == 'purge') {
 
         if ($success) {
             $OUTPUT->show_message('folderpurged', 'confirmation');
+            $OUTPUT->command('set_unread_count', $mbox, 0);
+            rcmail_set_unseen_count($mbox, 0);
+
+            // set trash folder state
+            if ($mbox === $trash_mbox) {
+                $OUTPUT->command('set_trash_count', 0);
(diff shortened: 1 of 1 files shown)


program/steps/mail/folders.inc
```diff
@@ -59,21 +59,21 @@ else if ($RCMAIL->action == 'purge') {
...
 
             if (!empty($_REQUEST['_reload'])) {
                 $OUTPUT->set_env('messagecount', 0);
                 $OUTPUT->set_env('pagecount', 0);
                 $OUTPUT->set_env('exists', 0);
                 $OUTPUT->command('message_list.clear');
                 $OUTPUT->command('set_rowcount', rcmail_get_messagecount_text(), $mbox);
-                $OUTPUT->command('set_unread_count', $mbox, 0);
                 $OUTPUT->command('set_quota', $RCMAIL->quota_content(null, $mbox));
                 rcmail_set_unseen_count($mbox, 0);
```
alecpl: I would also move `rcmail_set_unseen_count($mbox, 0);`. Other than that it looks good.

smares:
Just noticed that the icon is not updated and still shows a full trash can for example.

PR #7076, opened 2019-11-26
filter unwanted subdomains like mail. pop. imap.

f1-outsourcing:
the popup does not take into account that the $config['default_host']  has subdomains like mail or imap or pop. Those are of course not wanted to display in the domain of the email address.
Maybe you can move this delarr into the config?


alecpl:
Use `username_domain` option for this.

PR #7091, opened 2019-11-29
Fix ISO-2022-JP-MS encoding convert

shirosaki:
ISO-2022-JP-MS characters are removed by iconv on sending mail.
Use mbstring functions on sending.
A message with ISO-2022-JP-MS characters currupts by wordwrap.
Use ISO-2022-JP-MS charset in wordwrap.

### DIFF
--- a/program/include/rcmail_sendmail.php
+++ b/program/include/rcmail_sendmail.php
@@ -671,6 +671,10 @@ class rcmail_sendmail
      */
     public function email_input_format($mailto, $count = false, $check = true)
     {
+        // convert to UTF-8 to preserve \x2c(,) and \x3b(;) used in ISO-2022-JP;
+        $charset = $this->options['charset'];
+        $mailto = rcube_charset::convert($mailto, $charset, RCUBE_CHARSET);
+
         // simplified email regexp, supporting quoted local part
         $email_regexp = '(\S+|("[^"]+"))@\S+';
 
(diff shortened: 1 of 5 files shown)


alecpl:
Is it a PHP bug? Could you provide some samples? Create some unit tests?

shirosaki:
This is not a PHP bug.
Japanese OS environment would be required to reproduce this issue.

Related to this issue.
https://github.com/roundcube/roundcubemail/pull/5668

Use this plugin.
https://github.com/joniburn/sendcharset/blob/master/sendcharset.php

Change send charset to ISO-2022-JP.

If I send a message with subject, body or to/cc address name that contain Japanese characters including ISO-2022-JP-MS characters like '①②③', Japanese characters are removed or corrupted.

I'll try to add unit tests.

shirosaki:
I've split patches and add a test for each change.

program/include/rcmail_sendmail.php
```diff
@@ -702,7 +706,21 @@ public function email_input_format($mailto, $count = false, $check = true)
...
+                    $head_encoding_mode = 'B';
+                    $name = rcube_charset::convert($name, RCUBE_CHARSET, $charset);
+                    $mb_charset = $charset;
+                    if (rcube_charset::$mb_aliases[$charset]) {
+                        $mb_charset = rcube_charset::$mb_aliases[$charset];
+                    }
+                    mb_internal_encoding($mb_charset);
+                    $name = preg_replace('/^"(.*)"$/', '$1', $name);
+                    $name = mb_encode_mimeheader($name, $mb_charset, $head_encoding_mode, "\r\n", 8);
+                    mb_internal_encoding(RCUBE_CHARSET);
```
shirosaki: This part could be replaced by `Mail_mimePart::encodeMB()` if 
https://github.com/pear/Mail_Mime/pull/18 was accepted.

alecpl:
Whay don't you just set the sendcharset to ISO-2022-JP-MS?

shirosaki:
On Wed, Dec 4, 2019 at 20:16 Aleksander Machniak <notifications@github.com>
wrote:

> Whay don't you just set the sendcharset to ISO-2022-JP-MS
>

Mailer such as Thunderbird cannot decode Base64 header string contains
=?ISO-2022-JP-MS?B? .
ISO-2022-JP-MS looks like PHP specific.

> --
Hiroshi Shirosaki

shirosaki:
@alecpl We have upgraded our roundcube to 1.4 with the patches. It works fine.
Other mailers and web mail services support this. So same behavior is expected in Japan.

This company seems to fix some Japanese encoding problems for enterprise support.
https://www.designet.co.jp/system/support/maintenance/roundcube.html

sanmai:
[See my comment here.](https://github.com/pear/Mail_Mime/pull/18#issuecomment-580528487) 

@shirosaki Am I passing your point along all right?

shirosaki:
@sanmai You are right.
We have to use iso-2022-jp-ms for iso-2022-jp variant characters.

PR #7143, opened 2020-01-03
Fix: Composer missing icons when 'htmleditor' is in config['dont_override']

Rawk:
`html_editor_init()` in `skins/elastic/ui.js` requires a `<select name="editorSelector">` element after the `<textarea>`. But this select element is not rendered by the template if `'htmleditor'` is in `config['dont_override']`.

This leads to `html_editor_init()` not adding the `'html-editor'` class, making the html/plaintext icons invisible.

This pull request fixes this by always rendering the editorSelector. I don't know the reason for not rendering it, so maybe this will break something.

### DIFF
--- a/skins/elastic/templates/compose.html
+++ b/skins/elastic/templates/compose.html
@@ -60,14 +60,12 @@
 				</div>
 			</div>
 			<roundcube:endif />
-			<roundcube:if condition="!in_array('htmleditor', (array)config:dont_override)" />
-				<div class="form-group row hidden">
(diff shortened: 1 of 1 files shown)


alecpl:
There's indeed a bug, however... I think the intention of "locking" htmleditor was to disable editor mode switching at all (not only set the initial state). At least that's how old skins have it implemented. That's why I think the way to fix it will be to remove the broken button from the editor toolbar.

alecpl:
Fixed.

PR #6762, opened 2019-05-19
TinyMCE 5

alecpl:
Because of many TinyMCE 5 bugs I think we should not merge it into Roundcube 1.4 yet.

Here's a list of most prominent issues:
- https://github.com/tinymce/tinymce/issues/4865
- https://github.com/tinymce/tinymce/issues/4863
- https://github.com/tinymce/tinymce/issues/4862
- https://github.com/tinymce/tinymce/issues/5003

Maybe none of them is serious, but all together give impression of unfinished product. There are [more tickets](https://github.com/tinymce/tinymce/issues/created_by/alecpl) I created there.

https://kolabian.wordpress.com/2019/05/23/tinymce-5/

PR #7144, opened 2020-01-04
Update to TinyMCE 5.1

EwoutH:
- Update TinyMCE from 4.8.2 to 5.1.5 in jsdeps.json
 - Update SHA-1 hash

As far as I can see none of the [breaking changes](https://www.tiny.cloud/docs/migration-from-4x/) in TinyMCE 5 affect RoundCube, but please help testing, especially the media file selector since it uses some custom CSS.

### DIFF
--- a/jsdeps.json
+++ b/jsdeps.json
@@ -36,12 +36,12 @@
     },
     {
       "lib": "tinymce",
-      "version": "4.8.2",
+      "version": "5.1.5",
       "url": "https://download.tiny.cloud/tinymce/community/tinymce_$v.zip",
       "dest": "program/js",
-      "sha1": "d7fced05acdeeb78299585ea9909b0de2b3d759d",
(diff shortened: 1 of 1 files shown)


alecpl:
There are breaking changes. See #6762.

EwoutH:
@alecpl Thank you for the quick reply! All 4 issues listed in PR 6762 are now closed. In my repo it also passed the CI, something it didn't do in the earlier PR.

EwoutH:
I now see there are a lot more open issues with TinyMCE 5, a lot of which you reported. Thanks for all your work on RoundCube so far!

alecpl:
Some issues are irrelevant as I decided we'll not use mobile theme (which is deprecated). Still there's some things to consider and work on, but I'm on it. I also am waiting for TinyMCE 5.2. We have time to do the switch in 1.5, it will not be backported to 1.4 anyway.

PR #7152, opened 2020-01-08
Fix de_DE translation for errcsrfprotectionexplain

nilclass:
Previously the translation stated that the resources was protected *with* CSRF, while in reality it is protected *against* it.

The [`en_US` version of this sentence](https://github.com/roundcube/roundcubemail/blob/fded360d843d9d4ba16a4dee93ed8e798aa804cb/program/localization/en_US/messages.inc#L214) states: "For your protection, access to this resource is secured against CSRF."


alecpl:
I fixed this on Transifex it will be pushed next time we do localizations update.

Next time please use https://www.transifex.com/roundcube/roundcube-webmail

PR #7168, opened 2020-01-12
Update labels.inc

Sdarot:
flip dollar signs


alecpl:
Next time use https://www.transifex.com/roundcube/roundcube-webmail/, please. I corrected it there. It will be pushed with next localization update.

PR #7179, opened 2020-01-16
Enable change of login form field type in config.

petrkle:
If input type is set to 'email' browser validate input before submit. Also on mobile client keyboard with @ is displayed.


alecpl:
Maybe instead of a new option we should implement that as part of `login_username_filter`. It could accept special value "email". This way we have one config option less, and we have syntax check for the posted value. I'm not sure yet, but having a config option only for the input type is not what we normally do.

alecpl:
Implemented in 1e1ea25.

Issue #7193, opened 2020-01-23
Reading Pane - Understanding the issue


FUGithubSupportFF:
I use about 12 different mail clients, every one of them has an option to turn a preview pane on and off.

Unfortunately my hosting company has implemented Roundcube mail for all my domains and I can't believe you do not have this.

It seems that earlier versions have it because they used to talk about a config file option.

It is all about workflow, obviously if you receive a lot of mail you want to start with a list and do not want to open each one or even have it preview as it wastes time,  Some you will drag and drop into folders, but it is more than that, you want the screen space that the preview is taking up

There are very few config options compared to other webmail, nothing that gives a users the ability to change the awful layout and design 

As if this was not bad enough "white space" issue has crept in, what is the white space issue,? It is when developers have huge screens and create loads of unnecessary white space which may be fine for them but is awful on the most common screen format; the 15" laptop.  I can see 17 emails on my ISP email and just 4 on this awful roundcube mail client.

People want an "at a glance" view, plus the ability to rapidly decide what mails are important by the configurable list of fields on a list of their inbox.  
 
What shocked me even  more was when the author said 

"No such feature. As the skin author I'm really not interested in having it. Such feature would increase complexity much and Elastic's code is already quite big."

How bloody inconsiderate and arrogant.  Not Interested!  How about your users!  

So I will ask again, and yes your code is quite big, in fact it has a memory leak so if you use this client for any length of time it uses up gigabytes of memory and eventually crashed the browser.  The only way I could stop this was to stop it refreshing.

I would like to congratulate you on designing the most sloppy client I have seen in a long while.

I hope you manage to find some professional pride and self respect, then sort this out.

If you need to know what it should look like just google disable X reading pane where X is the mail client, e.g. Yahoo, Gmail, Thunderbird, Outlook, Hotmail, GMX etc, even Squirrel had this feature.

A similar request is below but promptly closed because someone was not interested because they already wrote bloated code.
https://github.com/roundcube/roundcubemail/issues/6458

alecpl:
Thank you for your opinion. Use different skin, write a plugin, whatever. I don't have time. Provide a pull request then I may re-consider.

Issue #7161, opened 2020-01-10
Elastic: Missing 'Layout' feature


dennisscholing:
Hi,

Within the Larry skin, the 'Layout' option is present under 'Mailbox view'. This cannot be found within Elastic.

Is it possible to retrieve this? Or is that already planned? The 'List' and 'Desktop' option will be missed within the current new skin!

alecpl:
This feature is not planned for implementation.

dennisscholing:
It was available 1.4.0. Why remove it? :)

alecpl:
No, it was never available in Elastic.

PR #7195, opened 2020-01-25
Elastic: support list type layout

johndoh:
First attempt at adding a list type layout mode to the Elastic skin (no preview pane)
demo
I tried a few different layouts for the combined toolbar above the message list. I think I prefer option 2 from the selection below.
toolbars
Most of the changes are confined to the Elastic skin itself but there were a couple of updates needed in the core.

1. allow skins to device which layout modes they support - this PR adds support for the `list` layout mode but not `desktop`. So I needed a way for skins to define which layouts they support. I have added `supported-layouts` to the skin `meta.json` file and an internal config option called `skin_layouts` which controls which options are shown on the settings screen.
2. separation of skin layout and message list layout options -currently the value of `layout` is used both for the skin layout and for the layout of the fields in the message list. That's good for Classic and Larry but not for Elastic ~~so by setting `layout_messagelist` the layout of the message list can be defined separately. If `layout_messagelist` is not defined then `layout` is used.~~ I also made a small tweak in `app.js` to ensure that `add_message_row()` uses the same message row layout that the PHP does when generating the row. Edit 1: I've added a new JS env var called list_layout which can be used to set the list layout separately to the screen layout, if this env var is not present then the `layout` env is still used.

UPDATE: I've added support for "list mode" message list on large screens. 

I know this feature is not a priority for the devs so it would be great if people who opened issues (e.g. #7161, #7193) about hiding the preview pane could help test this PR. All feedback welcome. Thanks.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -371,6 +371,18 @@ EOF;
             }
         }
 
+        $layouts = $this->config->get('skin_layouts');
+        if (!empty($meta['supported-layouts'])) {
+            $layouts = $meta['supported-layouts'];
+            $this->config->set('skin_layouts', $meta['supported-layouts'], true);
+        }
+
+        // make sure current layout config is supported by current skin
+        // if not then fallback to first defined layout
+        if (!in_array($this->config->get('layout', 'widescreen'), $layouts)) {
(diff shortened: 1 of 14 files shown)


johndoh:
closing this. see the discussion in #7222

Issue #7206, opened 2020-02-04
Markasjunk: Moving to Junk fails on messages selected with Select > All


dddns:
Hi Alec,
To help, I've saw this on the logs, have no idea how it happen or who's the user, but the line number may help you.
RC 1.4.1

Feb  4 10:04:12 fastweb roundcube: PHP Warning:  Invalid argument supplied for foreach() in /home/hosting/host.pt/webmail/plugins/markasjunk/markasjunk.php on line 278

alecpl:
This is a real bug. It happens when you mark all messages in a folder and try to mark them as junk. Looks like that part of the code works only with "simple" selection (which is if you select one or more messages, but not all).

alecpl:
Actually, all drivers assume that $uids is an array, but it could be also "*". So, it's not that trivial to fix. @johndoh, maybe you would like to give it a try?

alecpl:
Fixed.

PR #7219, opened 2020-02-09
markasjunk: handle select all case (#7206)

johndoh:
pr for #7206

### DIFF
--- a/plugins/markasjunk/markasjunk.php
+++ b/plugins/markasjunk/markasjunk.php
@@ -42,6 +42,7 @@ class markasjunk extends rcube_plugin
         'JUNK'    => 'Junk',
         'NONJUNK' => 'NonJunk'
     );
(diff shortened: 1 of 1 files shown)


johndoh:
alternatively support for this case could be added into `rcmail::get_uids()` then other plugins etc could use it too.

alecpl:
There are cases when we don't need to resolve "*" into an array of uids. rcube_imap supports "*" as an input. That's why I think get_uids() does not need to resolve that (maybe it could be an option).

Also, if you didn't enable any driver in markasjunk (you really need only move-to-junk functionality) we could also skip that resolution, but we'd need to modify markasjunk_move method.

johndoh:
Agreed the requirement to expect the UID list is specific to the drivers. I've refactored the code a little to take this into account.

PR #7232, opened 2020-02-20
MySQL charset change

army1349:
There is no reason to use "SET NAMES", when charset connection parameter is used.
This way charset can be easily configured in db_dsnw or forced on server side.
Motivation for this is use of utf8mb4 charset.

### DIFF
--- a/program/lib/Roundcube/db/mysql.php
+++ b/program/lib/Roundcube/db/mysql.php
@@ -47,17 +47,6 @@ class rcube_db_mysql extends rcube_db
     }
 
     /**
-     * Driver-specific configuration of database connection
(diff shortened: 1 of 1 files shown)


alecpl:
We actually switched to utf8mb4, so it was a leftover. Fixed in b7410ffe89f. Thanks.

army1349:
> 
> 
> We actually switched to utf8mb4, so it was a leftover. Fixed in [b7410ff](https://github.com/roundcube/roundcubemail/commit/b7410ffe89f33b7c55267f70c8e463664f7d9ab3). Thanks.

Nice work, thanks.

PR #7235, opened 2020-02-22
allow skins to define which layout options they support

johndoh:
Elastic currently sets the `layout` config option to `dont_override`. This is a proposal for an alternative approach which allows a skin to define which of the core layout modes a skin supports. If only 1 mode is supported then the option is hidden on the settings screen.

This allows, for example, for a new skin to extend Elastic and add new layout options.

(excerpt from #7195)

### DIFF
--- a/program/lib/Roundcube/rcube_config.php
+++ b/program/lib/Roundcube/rcube_config.php
@@ -257,6 +257,9 @@ class rcube_config
             ini_set('error_log', $error_log);
         }
 
+        // set default screen layouts
+        $this->prop['supported_layouts'] = array('widescreen', 'desktop', 'list');
+
         // remove deprecated properties
         unset($this->prop['dst_active']);
     }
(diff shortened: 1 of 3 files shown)


alecpl:
I don't like the changes in rcmail_output_html.php at all. I think a single `skin_layouts` (or just `layouts` or `supported_layouts`) setting should be enough.

johndoh:
Is it the the entire change or only the `base_skin` param bit? because the `base_skin`. Its purpose was to prevent the layout option being incorrectly overwritten when extending an extending skin to support new layouts but its not technically required. The second part which checks the user's configured `layout` option is supported by the skin is required but may be there is a better place for that code.

johndoh:
Perhaps the code to ensure the `layout` option is within `skin_layouts` could be moved to the rcube_config::get() method.

alecpl:
I just think this is a wrong place to do that. This part of the code should not deal with specific config options like this one.

If we make skin_layouts (supported_layouts) a normal config option, we don't need the supported_layout metadata. I think that specifying this in Elastic should be enough:
```
"config": {
   "supported_layouts": ["widescreen"]
   ... // no "layout" here
}
```
Then you can override that from a skin as well as from a plugin.

PR #7236, opened 2020-02-22
create possibility to seperate screen and list layout options

johndoh:
Extend current UI layout behavior to allow a skin to use a different layout mode for the screen and for the message list. Currently if the screen is in widescreen mode then the list is also in widescreen mode.

(excerpt from #7195)

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -2364,7 +2364,7 @@ function rcube_webmail()
       row.cols.push(col);
     }
 
-    if (this.env.layout == 'widescreen')
+    if (flags.layout == 'widescreen')
       row = this.widescreen_message_row(row, uid, message);
 
     list.insert_row(row, attop);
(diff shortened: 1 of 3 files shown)


alecpl:
Do we really need this?

johndoh:
My use case. In the current code the layout of the list is tied to the `layout` and you use the `list` layout. On a big screen this is fine, lots of space for all the columns. But on a small screen, say a phone, showing all the columns is not practial and so the `widescreen` list layout is preferable. By separating the `layout` option from the list a skin author can which layout they want for the list depending on what layouts and devices they want their skin to support.

johndoh:
Another idea might be to remove the `layout` option from the PHP code for the message list and instead return a JSON object for each message row with the cols required for the `widescreen` layout as a minimum plus any other cols configured for display and then construct the `<tr>` element in app.js so that plugins/skins can manipulate the layout of the row if they wish. May be this would cause a performance hit though, I'm not sure.

johndoh:
I'll resubmit a new idea.

PR #7237, opened 2020-02-22
improvements to options menu link template object

johndoh:
Here I propose 2 changes.

1) is to allow template object attributes to contain conditions or expressions as well as static values.
2) is to add some more parameters to the rcmail_options_menu_link template object and improve the flexibility for use both as part of the message list template object and as a standalone element.

These changes are backwards compatible.

(excerpt from #7195)

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1529,12 +1529,17 @@ EOF;
      */
     protected function prepare_object_attribs(&$attribs)
     {
-        // Localize data-label-* attributes
-        array_walk($attribs, function(&$value, $key, $rcube) {
+        foreach ($attribs as $key => &$value) {
             if (strpos($key, 'data-label-') === 0) {
-                $value = $rcube->gettext($value);
+                // Localize data-label-* attributes
+                $value = $this->app->gettext($value);
             }
-        }, $this->app);
+            elseif ($key[0] == ':') {
+                // Evaluate attributes with expressions and remove special character from attribute name
+                $attribs[substr($key, 1)] = $this->eval_expression($value);
+                unset($attribs[$key]);
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


program/include/rcmail_output_html.php
```diff
@@ -1529,11 +1529,19 @@ protected function xml_command($matches)
      */
     protected function prepare_object_attribs(&$attribs)
     {
-        // Localize data-label-* attributes
         array_walk($attribs, function(&$value, $key, $rcube) {
             if (strpos($key, 'data-label-') === 0) {
+                // Localize data-label-* attributes
                 $value = $rcube->gettext($value);
             }
+            elseif (strpos($value, 'condition:') === 0) {
```
alecpl: I would make these compare the `$key` not `$value`. Also, maybe it would be more convenient to use `exp:` and `cond:` prefix. And, because check_condition is the same as eval_expression, I think we only need one, `exp:`. Or maybe even as Vue does, only the colon character.

program/steps/mail/func.inc
```diff
@@ -673,8 +685,8 @@ function rcmail_options_menu_link($attrib = array())
     return html::a(array(
             'href'     => '#list-options',
             'onclick'  => $onclick,
-            'class'    => isset($attrib['class']) ? $attrib['class'] : 'listmenu',
-            'id'       => 'listmenulink',
+            'class'    => $attrib['class'] ?: 'listmenu',
```
alecpl: Well, that's not exactly backward compatible, but I think we can leave it this way.

alecpl (changes requested):

program/include/rcmail_output_html.php
johndoh: I think `attribute-name="exp:!in_array('list_cols', (array)config:dont_override)"` is cleaner looking than `exp:attribute-name="!in_array('list_cols', (array)config:dont_override)"`

alecpl: Using this on key is safer, i.e. using "exp:..." as a value that should be send to the browser will be problematic. Let's do ":attribute-name", some people should be familiar with such a way.

johndoh: Changes done. I know its a slightly bigger change now foreach is faster than array_walk.

PR #7239, opened 2020-02-22
allow possibility to define the container of a table with a fixed header

johndoh:
Current code assumes that the container of a table with a fixed header is always the `window`. This adds the ability for a skin or plugin to define another container.

(excerpt from #7195)

### DIFF
--- a/program/js/list.js
+++ b/program/js/list.js
@@ -259,12 +259,18 @@ init_header: function()
   }
 },
 
+/**
+ * Set the scrollable parent object for the table's fixed header
+ */
+container: window,
+
 init_fixed_header: function()
 {
   var clone = $(this.list.tHead).clone();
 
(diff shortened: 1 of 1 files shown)


program/js/list.js
```diff
@@ -259,12 +259,18 @@ init_header: function()
   }
 },
 
+/**
+ * Set the scrollable parent object for the table's fixed header
+ */
+container: window,
```
alecpl: I'd move this to rcube_list_widget() where all vars are initialized. Other than that it looks good.

alecpl (changes requested):

program/js/list.js
johndoh: I placed it here because if its in rcube_list_widget() I cannot find a way to override the setting before the event listener is added to the container during obj initialization.

PR #7240, opened 2020-02-22
add some events and public methods to elastic ui.js

johndoh:
I'm not 100% if I've done this right but currently extending the Elastic skin is very tricky by adding a few public methods and JS even triggers I am to simplify that a bit.

If you do not like this approach please close this PR so I know to rethink.

(excerpt from #7195)

### DIFF
--- a/skins/elastic/ui.js
+++ b/skins/elastic/ui.js
@@ -82,7 +82,10 @@ function rcube_elastic_ui()
     this.pretty_select = pretty_select;
     this.datepicker_init = datepicker_init;
     this.bootstrap_style = bootstrap_style;
-
+    this.toggle_list_selection = toggle_list_selection;
+    this.get_screen_mode = get_screen_mode;
+    this.is_mobile = is_mobile;
+    this.is_touch = is_touch;
 
     // Detect screen size/mode
     screen_mode();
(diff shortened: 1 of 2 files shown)


skins/elastic/ui.js
```diff
@@ -2467,6 +2487,15 @@ function rcube_elastic_ui()
             return parent.UI.menu_toggle(p);
         }
 
+        var ret = undefined;
```
alecpl: `= undefined` is redundant.

skins/elastic/ui.js
```diff
@@ -2467,6 +2487,15 @@ function rcube_elastic_ui()
...
+        if ((ret = rcmail.triggerEvent('skin-menu-toggle', { props: p })) !== undefined) {
```
alecpl: What use-case do you have for this whole code change?

alecpl (changes requested):
The general idea is all right. I'm not sure about one part, see inline comment.

skins/elastic/ui.js
johndoh: I would like to add option to the `messagelistmenu` dialog. To do that I need to be able to override the default handling so that I can set the current value of my field when the dialog is opened and read the value when the save button is clicked.

alecpl: I think you can do that using menu-open and menu-close events.

johndoh: menu-open might work but you have fish around to find the dialog and technically its fired before the dialog is created by Elastic (coz it also relies on this trigger). menu-close is not fired when the save button is clicked.

johndoh:
I think you are right about the `skin-menu-toggle` trigger it was duplicating functionality which already exists so I will find a different solution. In the mean time I added a small sanity check to the `menu_toggle()` function so that it does not error when there is a `menu-open` event with custom content (eg a menu added by a plugin).

PR #7242, opened 2020-02-23
Small layout changes for elastic skin

alexte:
Small layout changes to improve usability on small screens


alecpl:
I don't like the part about messagestack move.

alexte:
The reason is "usability" for me. The message stack hides the delete button for some seconds. Which slows down deleting mails.

But there might be a nicer solution.

Issue #7216, opened 2020-02-08
Add more options to the layout of the archive folder


mildred:
Hi,

I'd like to suggest to add more options to the archive folder layout (preferences > special folders > archive) :

- Year then original folder (Archives/YYYY/original/folder)
- Month then original folder (Archives/YYYY/MM/original/folder)

These options are available in Thunderbird, and for compatibility it would make sense to include them.

alecpl:
Should be easy. Pull requests appreciated.

mildred:
See linked PR for implementation.

alecpl:
Implemented. Thanks for the PR.

PR #7266, opened 2020-03-04
Add new archival types to keep the folder structure

mildred:
Add archival type folderyear and foldermonth to keep the original folder
structure after a YYYY or a YYYY/MM prefix.

Implements #7216


mildred:
I only added translation strings for French and English

alecpl:
Adding localizations other than en_US is not needed. They have to be applied on Transifex anyway, and could get lost on localization update.

alecpl (approved):

PR #7282, opened 2020-03-13
Change line 82 from 'junk' to 'markasjunk.junk'

keysuck:
corrected wrong Spam button label


alecpl:
What is wrong with the label? There's no such label in the plugin, that's why the "core" label is used.

keysuck:
It loads translation for 'Junk' inbox without the mod.

alecpl:
There's no `junk` label in markasjunk plugin localization. And that's the point, the label should be localized. Is this an issue in some specific localization?

PR #7283, opened 2020-03-15
program/js/app.js: Remove forcing _mbox to INBOX on task switch.

iazz:
There is no need to force the _mbox query parameter to "INBOX" when
switching task to "mail", since the actual selection of the current
folder is taken care of by rcmail_init_env() in
program/steps/mail/func.php.  In addition, by letting the latter
manage the current folder, going back to the "mail" task always goes
back to the last opened folder, which is more intuitive.

Signed-off-by: Ignacy Gawędzki <bugs@qult.net>

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -1582,9 +1582,7 @@ function rcube_webmail()
     if (action)
       url += '&_action=' + action;
 
-    if (task == 'mail')
-      url += '&_mbox=INBOX';
-    else if (task == 'logout') {
+    if (task == 'logout') {
       url = this.secure_url(url);
       this.clear_compose_data();
     }


alecpl:
Not many people complained about this (and it is like this since forever). So, I wouldn't be so sure it's better.

iazz:
> Not many people complained about this (and it is like this since forever). So, I wouldn't be so sure it's better.

Not many people seem to need to have a startup folder different than INBOX either.  But I needed that and was quite surprised to discover that no such feature is present in the stock software nor in any publicly available plugin.  So I just wrote such a plugin and while testing it I discovered that the open folder is force-switched to INBOX when going back from "settings" or "addressbook" task.  I dove into the intricacies of the PHP and JS code only to come to the conclusion that this forcing makes not much sense.  The currently open folder is kept in $_SESSION and if it cannot be determined, then it is reset to INBOX anyway.

My pull request doesn't remove anything for users who don't complain, it only adds a feature for those who do.

alecpl:
My point is that now users expect to land in INBOX when switching back from other places. So, "doesn't remove anything" is not actually true.

iazz:
Oh I see, in fact my commit removes a "feature", not a bug.  A "feature" that looks like it landed in the code by pure chance.  The problem is that my "startup folder" plugin doesn't work properly if the switch_task function is left unchanged.  I suppose that if that "feature" is so useful, there should be an option to enable/disable it in the settings.

PR #7287, opened 2020-03-16
move all message list entry generation to client side

johndoh:
This is a rethink of PR #7236.

Rather than changing which cols are returned by `rcmail_js_message_list()` the idea is to always return the user's configured columns plus the hard coded cols required to create the widescreen layout of the message list and then allow the `add_message_row()` JS function to decide what to display and the layout.

I've also added a new JS even to allow skins/plugins to override the defaults and define how a message list row should be rendered.

### DIFF
--- a/program/steps/mail/func.inc
+++ b/program/steps/mail/func.inc
@@ -448,10 +448,7 @@ function rcmail_js_message_list($a_headers, $insert_top=false, $a_show_cols=null
     $thead = $head_replace ? rcmail_message_list_head($_SESSION['list_attrib'], $allcols) : NULL;
 
     // get name of smart From/To column in folder context
-    if (array_search('fromto', $a_show_cols) !== false) {
-        $smart_col = rcmail_message_list_smart_column_name();
-    }
-
+    $smart_col = rcmail_message_list_smart_column_name();
     $OUTPUT->command('set_message_coltypes', array_values($a_show_cols), $thead, $smart_col);
 
     if ($multifolder && $_SESSION['search_scope'] == 'all') {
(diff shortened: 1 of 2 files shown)


alecpl:
Not bad, but
1. I'm pretty sure we don't need a new event. You can just change env.layout and env.listcols, right?
2. Tests fail, so there might be some issue to fix.

johndoh:
How about adding the layout option as a param to add_message_row if a new hook is too much? the reason for not changing rcmail.env.layout is that the main aim here is to get some separation between screen layout and the list layout.

i'll work on the failing test.

program/steps/mail/func.inc
```diff
@@ -446,15 +441,11 @@ function rcmail_js_message_list($a_headers, $insert_top=false, $a_show_cols=null
...
-    }
+    // make sure minimum required columns are present (needed for widescreen layout)
+    $allcols = array_merge($a_show_cols, array('threads', 'subject', 'fromto', 'date', 'flag', 'attachment'));
+    $allcols = array_unique($allcols);
 
-    $thead = $head_replace ? rcmail_message_list_head($_SESSION['list_attrib'], $a_show_cols) : NULL;
+    $thead = $head_replace ? rcmail_message_list_head($_SESSION['list_attrib'], $allcols) : NULL;
 
     // get name of smart From/To column in folder context
     if (array_search('fromto', $a_show_cols) !== false) {
```
haroflow, line 451: Using `$a_show_cols` on this line hides the Sender e-mail on widescreen view for me, because `$smart_col` is left empty.
Correct me if i'm wrong, but looks like it should be `$allcols`, fixed for me anyways...

johndoh: good catch thanks. happens when the smart col is not included in your col choices for list mode but actually the check here is pointless we can just pass the col name to the JS and it will use it if required.

PR #7296, opened 2020-03-23
allow array in smtp_host config

johndoh:
it is possible to configure an array of imap hosts but not smtp hosts so if not all your providers use the same naming conventions for their servers you can configure the imap but the smtp.

inspired by https://www.roundcubeforum.net/index.php/topic,28819.msg73532.html#msg73532

if this is accepted I'll update the config information in the wiki to include the array option, similar to what is there for the default_host option - i did not update the comments in `defaults.inc.php` because there is nothing there about the array option for the imap so I think the wiki is enough.

I've also fixed a bug in the installer which meant that the `%h` macro of `smtp_server` was never expanded and always returned blank.


alecpl:
Actually it is mentioned in "Leave blank to show a textbox at login, give a list of hosts to display a pulldown menu or set one host as string.". I'd like to have the array option mentioned in defaults.inc.php. Other than that it looks fine.

PR #7317, opened 2020-04-07
fix a bug in % Chars

SCR-IR:
Change % to $ in some lines


alecpl:
Fixed on https://www.transifex.com/roundcube/roundcube-webmail/translate/#fa_IR/messages/9050154?q=text%3Acode

Issue #7251, opened 2020-02-26
Session should not be destoryed with window close


HighlyFavoredBA:
With Version 1.4.0, and Elastic, I noticed that Firefox, and most phone browsers log me out of Email when I close, and re open the browsers, if I restart my phone, etc., and that was also happening on my last phone that I recently had.  

In a dated R C version install that I have, the (melanie2_larry_mobile) skin has been great about keeping me logged in!   :--)

I prefer 7200 minutes of no activity.

// Session lifetime in minutes
$config['session_lifetime'] = 7200;

alecpl:
What's your point exactly?

HighlyFavoredBA:
Anybody that has a session span of more than a few hours would prefer to not be auto logged out, simply from restarting phones, etc.!  Lots of Email software are excellent about keeping users logged in.  :--)

HighlyFavoredBA:
My 1.2.7 install with the (melanie2_larry_mobile) skin is great at keeping me logged in when all the mobile browsers that I've tested!  As long as the (melanie2_larry_mobile) skin keeps running, I plan to keep using that on the side, until more of the kinks are removed from the 1.4 versions, and the Elastic skin.

Is the auto logging out when closing mobile browsers, a 1.4.* issue, or is it a Elastic skin issue?  :--)

alecpl:
I just noticed that closing the browser window ends a session, or it looks like that, because when you open it again you'll be "forced" to log in again. I don't think this is intentional. I agree that it work differently before. This need to be investigated.

alecpl:
I did some investigation. I thought it might be related to 403d8453c, but if I comment out the added line the behavior is the same. So, maybe something has changed in PHP or browsers recently. I cannot find another change that would be related.

Here's what I had to do to make it working as expected:
```diff
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -463,8 +463,7 @@ class rcube
             ini_set('session.gc_maxlifetime', $lifetime * 2);
         }
 
-        // set session cookie lifetime so it never expires (#5961)
-        ini_set('session.cookie_lifetime', 0);
+        ini_set('session.cookie_lifetime', $lifetime);
         ini_set('session.cookie_secure', $is_secure);
         ini_set('session.name', $sess_name ?: 'roundcube_sessid');
         ini_set('session.use_cookies', 1);
[...]

PR #7326, opened 2020-04-15
7251 Add option to keep sessions when browser restarts

mkllnk:
This is a proposal to solve #7251 with backwards compatibility. The new configuration variable `session_lifetime_max` sets an expiry date on the session cookie. It will survive browser restarts. But since PHP is not renewing session cookies, it also limits the maximum time a user can be logged in. After that amount of time the user gets kicked out. This is different to the default configuration which will let the session cookie live until the browser is closed, which could be never.

### Draft

I see a possible issues here:

It could be very inconvenient to get kicked out while you are logged in, potentially writing an email. In order to prevent that, we would need to renew the session cookie. Is that something worthwhile?

Reviewing this situation, I'm not sure if the additional authentication cookie is necessary. Its purpose is better security but I'm wondering if we could renew the session cookie in a similar fashion.

The timeslot logic is a bit complex and doesn't add any security. An attacker could steal an old cookie and guess the new timestamp after observing the session lifetime. I'm not sure I have enough time for a re-implementation though.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -488,7 +488,18 @@ $config['skin_include_php'] = false;
 $config['display_product_info'] = 1;
 
 // Session lifetime in minutes
+// While the browser is active, the session's life is constantly extended. But
+// when the browser is inactive, the session expires after this amount of
+// minutes.
 $config['session_lifetime'] = 10;
+
+// Maximum session lifetime in minutes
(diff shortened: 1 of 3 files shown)


mkllnk:
I'm closing this because I don't think that the "maximum session lifetime" option is a good improvement. It would be better to implement a refresh mechanism of the session cookie.

PR #7342, opened 2020-04-23
Option to make message too big errors more human friendly

nward:
When a user attempts to add an attachment which puts the message over the max message size limit, the error shown states the limit as the "real" maximum size. This is confusing for most people, as their attachments will always add up to far less than that maximum size.

This change adds an option, which is disabled by default, to make the maximum message size in these errors show up as the maximum "practical" size, calculated the same as we calculate the maximum attachment size - dividing by 1.33 to accomodate base64 overhead.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -703,6 +703,10 @@ $config['sendmail_delay'] = 0;
 // Size in bytes (possible unit suffix: K, M, G)
 $config['max_message_size'] = '100M';
 
+// Show message size by the attachable size in errors. i.e. the above limit
+// divided by 1.33 to account for base64 encoding.
+$config['show_message_size_by_attachment'] = false;
+
 // Maximum number of recipients per message (including To, Cc, Bcc).
 // Default: 0 (no limit)
 $config['max_recipients'] = 0;
--- a/program/actions/mail/attachment_upload.php
+++ b/program/actions/mail/attachment_upload.php
@@ -275,7 +275,12 @@ class rcmail_action_mail_attachment_upload extends rcmail_action_mail_index
         $size  += $filesize * $multip;
 
         if ($size > $limit) {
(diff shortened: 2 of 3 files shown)


alecpl:
I don't like the idea of another config option. We should probably make that default behavior.

nward:
I would be happy with that - I am proposing it as a config option so that it doesn't change behaviour unless someone opts in, but, I think is reasonable to have as a non-optional change.

Happy to amend this PR to remove the optional component - not sure what the culture is on this project - am I looking for feedback from others before doing that?

alecpl:
No, I'm the boss here ;) Though, I'm not yet sure what I will do about this PR. That was just a first thought.

piuma:
I think it must be the default behavior

atomindtechnology:
Me Too!

I think that as the label above the "upload attachment" button, shows the correct number of MB divided by 1.33;
the error message should also do it by default with no options.

alecpl:
"the label above the "upload attachment" button, shows the correct number of MB divided by 1.33;" is not true. It displays maximum file size for upload (as defined in PHP config).

It wasn't a problem until we added the option for max message size, which has to take the base64 encoding factor into account.

alecpl:
I revert my previous comment. The hint will indeed display the size divided by 1.33, but only when the defined max message size (divided by 1.33) is smaller than the max upload size. This makes it even more complicated. The fact that message/rfc822 attachments aren't encoded is another thing.

So, indeed the fact we display different values needs to be fixed, but in case the size in the form label is not divided by 1.33 it becomes not that obvious.

nward:
@alecpl I've updated this against the current master. We've been running this on 1.4 and 1.5 for a few years. I'm updating to 1.6 at the moment so will report test results from that shortly, once I've applied some internal patches.

Issue #5069, opened 2013-06-01
Drag & Drop for Contacts from "To" to "CC"


rcubetrac:
_Reported by backi on 1 Jun 2013 19:50 UTC as Trac ticket #1489141_

I dont know if this is a browser issue or a possible feature for roundcube. Mark Emailadresses/contacts from "To" or an other field and move or copy them to CC or BCC

_Keywords: Drag&Drop_
_Migrated-From: http://trac.roundcube.net/ticket/1489141_

chilek:
Copy&paste shortcuts do not work for ELASTIC skin so it's mission impossible for that.
You have to copy mail address using web browser dom inspector :(

chilek:
Ohhh I use release-1.4 branch HEAD.

alecpl:
Done for Elastic.

PR #7345, opened 2020-04-24
enhancement: elastic skin: allow to drag & drop addresses between add…

chilek:
…ress input fields during compose message

#5069

### DIFF
--- a/skins/elastic/ui.js
+++ b/skins/elastic/ui.js
@@ -235,6 +235,20 @@ function rcube_elastic_ui()
         // This have to be after mail compose feature above
         $('[data-recipient-input]').each(function() { recipient_input(this); });
 
+        var recipient_inputs = $('#_to, #_cc, #_bcc, #_replyto, #_followupto', $('.compose-headers')).siblings('.recipient-input');
+        recipient_inputs.each(function() {
+            $(this).sortable({
+                appendTo: document.body,
+                items: "> li.recipient",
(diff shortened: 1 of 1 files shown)


alecpl:
The code can be simplified if you move this to recipient_input().

chilek:
@alecpl thanks a lot for very fast feedback.

skins/elastic/ui.js
```diff
@@ -3252,7 +3252,19 @@ function rcube_elastic_ui()
...
             .on('mouseup', function () { selection = window.getSelection().toString(); })
-            .on('click', function() { if (!selection.length) input.focus(); });
+            .on('click', function() { if (!selection.length) input.focus(); })
+            .sortable({
+                appendTo: document.body,
+                items: "> .recipient",
+                connectWith: '.recipient-input',
+                receive: function(event, ui) {
+                    input.parent().appendTo(list);
+                    input.trigger($.Event('keydown', { key: "," }));
```
chilek, line 3262: This line can be replaced by:
```js
update_func();
```
Am I right?

alecpl: Yes, but `apply_func()` might be enough for this case. And for ui.sender I would use 'change' event instead of 'keydown' below.

alecpl:
Thanks, there were some issues still, but I fixed them. Merged.

Issue #7157, opened 2020-01-09
Use preferred/main keyring from Mailvelope


Frederick888:
Currently Roundcube uses the [unique user id](https://github.com/roundcube/roundcubemail/blob/a9abe62d58df8a321591af30a9e02d1bd6b0f5dc/program/lib/Roundcube/rcube_user.php#L256) as the keyring identifier in [Mailvelope integration](https://github.com/roundcube/roundcubemail/blob/a9abe62d58df8a321591af30a9e02d1bd6b0f5dc/program/js/app.js#L3718). However, as I use GnuPG with YubiKey, it's impossible for me to import my private key into the Roundcube keyring and there doesn't seem to be a way of wiring the GnuPG keys into other keyrings in Mailvelope, either.

Is it possible to allow users to select the main/preferred keyrings from Mailvelope? I think it could also be more convenient for users in some cases if they simply want to share the same keys for different users/websites.

![image](https://user-images.githubusercontent.com/4507647/72030099-24370180-32dc-11ea-8bc1-a8ed7db57fbd.png)

thomascube:
At the time of writing that code, the Mailvelope API didn't allow access to the main keyring. I'll check if that changed in the meantime...

Frederick888:
@thomascube Ah yeah, indeed its client API doesn't allow accessing the main and GnuPG keyrings. The identifiers of these 2 keyrings are `localhost|#|mailvelope` and `localhost|#|gnupg` respectively however in the client content script, it's explicitly [banned](https://github.com/mailvelope/mailvelope/blob/7c9279c/src/content-scripts/clientAPI.js#L58) to provide `host` as part of the identifier.

It's probably a design for better security but I think there should be a setting available to make them, especially the GnuPG one, accessible to client API consumers. I guess I can file an issue there to see how the Mailvelope developers think.

alecpl:
The `mailvelope_main_keyring` option already is implemented. I think this ticket should be now to make it available to the user through Preferences.

alecpl:
Implemented.

Issue #7017, opened 2019-11-11
Unsaved changes warning not used in Mailvelope mode


alecpl:
Steps:
1. Goto mail composer.
2. Click Encrypt button (to switch to Mailvelope mode).
3. Write anything to the mail body textarea (Mailvelope frame).
4. Click on Edit Identities button (or any other task switching button).

Result: Browser opens Settings > Identities interface without warning about unsaved changes.
Expected: A warning is displayed.

So, this is imo a bug. @thomascube you might have more info on `ref.compose_skip_unsavedcheck = true` in Mailvelope mode.

thomascube:
@alecpl I rememebr the problem being that we cannot query the Mailvelope API for the actual contents of the widget and therefore cannot detect whether it has changed since the last saving. You may argue that we can simply request the encrypted message content and compare that but this triggers actions like warnings if a recipient's public key is missing inside the widget. Such warnings are desired and expected when the user hits the "save" button but are likely unexpected and disturbing when triggered by a background task like the unsaved check.

alecpl:
Thanks. Agreed, if we're talking about background tasks. However, for switching task action we should always assume there are unsaved changes and display the warning. That will be better than not showing it and cause the content to be lost.

alecpl:
Fixed.

PR #7348, opened 2020-04-26
Several Mailvelope-related fixes and features

gurnec:
Each commit is listed below. I'm open to any disagreements/critiques, so please fire away!

 * ~~24d0cf5~~ *Merged.* Let Mailvelope use sender's address to find pubkeys to check signatures

    Also remove the showExternalContent option which was removed from the Mailvelope API January 2019.

 * cec240d Fix size of Mailvelope iframe for PGP-inlined mail

    Emails with PGP attachments appear correctly [like this](https://user-images.githubusercontent.com/2966862/80312069-38324080-87b1-11ea-8ce9-88e263e1931b.png), but emails with the PGP in the message body before this commit looked squished [like this](https://user-images.githubusercontent.com/2966862/80312089-55670f00-87b1-11ea-8944-f0510c642a28.png). Now they both appear correctly. (Only tested in Chromium and Firefox 75, sorry.)

 * ~~c3d4598~~ *Merged.* Add missing `\`s to regexes in `rcube_check_email()`

    An unrelated minor bug fix, I can remove it if you'd prefer.

 * 7f64384 Avoid initializing Mailvelope if we don't need to

    Small efficiency fix, but it also makes the next two commits a bit smaller in scope.

 * cc3779f Add config option `mailvelope_main_keyring` to use Mailvelope Main Keyring

    Defaults to false for backwards compatibility. It's probably only useful for new installs. Related: #7157

 * 2f71928 Add config option `mailvelope_keysize` for size of new Mailvelope keys

    Defaults to 4096 for backwards compatibility.

 * a121960 Always ask before discarding composed Mailvelopes

    The unpatched code appears to be intentionally *not* asking before discarding emails currently being composed. This does the exact opposite, which can be minorly annoying if one saves and then leaves while composing, but is much safer in other cases IMO. If others disagree, perhaps this could be a config option instead? Related: #7017

 * 9d39c0a Don't warn to re-add attachments when restoring a Mailvelope draft

    Currently, if you save a Mailvelope email draft, and then select to edit it from your Drafts folder, the "Already uploaded attachments cannot be encrypted"... warning message is unnecessarily displayed.

    edit: I just (Apr 29) amended this last commit to remove the use of `Object.values()` which isn't supported by IE. `Object.keys()` apparently [is supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Browser_compatibility).

 * ~~40ecd47~~ *Merged.* Show Encrypt button w/Mailvelope, even if disabled (added Apr 30)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -70,6 +70,7 @@ RELEASE 1.4.4
 - Security: Fix local file inclusion (and code execution) via crafted 'plugins' option [CVE-2020-12640]
 - Security: Fix CSRF bypass that could be used to log out an authenticated user [CVE-2020-12626] (#7302)
 - Mailvelope: Fix size of iframe for PGP-inlined mail (#7348)
+- Mailvelope: Add config option to use Main Keyring (#7348)
 
 RELEASE 1.4.3
 -------------
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -71,6 +71,7 @@ RELEASE 1.4.4
 - Security: Fix CSRF bypass that could be used to log out an authenticated user [CVE-2020-12626] (#7302)
 - Mailvelope: Fix size of iframe for PGP-inlined mail (#7348)
 - Mailvelope: Add config option to use Main Keyring (#7348)
+- Mailvelope: Add config option to set the size for new keys (#7348)
 
 RELEASE 1.4.3
 -------------
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -72,6 +72,7 @@ RELEASE 1.4.4
 - Mailvelope: Fix size of iframe for PGP-inlined mail (#7348)
 - Mailvelope: Add config option to use Main Keyring (#7348)
 - Mailvelope: Add config option to set the size for new keys (#7348)
+- Mailvelope: Always ask before discarding email currently being composed (#7348)
 
 RELEASE 1.4.3
 -------------
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -69,6 +69,7 @@ RELEASE 1.4.4
 - Security: Fix remote code execution via crafted 'im_convert_path' or 'im_identify_path' settings [CVE-2020-12641]
 - Security: Fix local file inclusion (and code execution) via crafted 'plugins' option [CVE-2020-12640]
 - Security: Fix CSRF bypass that could be used to log out an authenticated user [CVE-2020-12626] (#7302)
+- Mailvelope: Fix size of iframe for PGP-inlined mail (#7348)
 
 RELEASE 1.4.3
 -------------
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -73,6 +73,7 @@ RELEASE 1.4.4
 - Mailvelope: Add config option to use Main Keyring (#7348)
 - Mailvelope: Add config option to set the size for new keys (#7348)
 - Mailvelope: Always ask before discarding email currently being composed (#7348)
+- Mailvelope: Fix unnecessary warning to re-add attachments when restoring a draft (#7348)
 
 RELEASE 1.4.3
 -------------
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -86,3 +86,7 @@ $config['plugins'] = array(
 
 // skin name: folder from skins/
 $config['skin'] = 'elastic';
+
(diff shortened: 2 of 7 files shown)


alecpl:
Maybe we should make `mailvelope_main_keyring` a user preference. Assigning @thomascube for review.

ryan77627:
Commit 2367aa7 is also working for Firefox 75, just verified it.

alecpl:
I cherry-picked fixes from here and backported to 1.4. The rest is a feature request. Please, rebase

gurnec:
@alecpl Thanks! Rebased, and I'll update the initial comment.

gurnec:
@alecpl One other quick note: I believe this one also qualifies as only-a-bugfix:

> * 9c38a44 Fix size of Mailvelope iframe for PGP-inlined mail
>
>    Emails with PGP *attachments* appear correctly [like this](https://user-images.githubusercontent.com/2966862/80312069-38324080-87b1-11ea-8ce9-88e263e1931b.png), but emails with the PGP *in the message body* before this commit looked squished [like this](https://user-images.githubusercontent.com/2966862/80312089-55670f00-87b1-11ea-8944-f0510c642a28.png). Now they both appear correctly. (Only tested in Chromium and Firefox 75, sorry.)

thomascube:
> Maybe we should make `mailvelope_main_keyring` a user preference. Assigning @thomascube for review.

@alecpl, @gurnec I agree this might be worth a user preference and not a fixed setting for all.
I wasn't aware that using the main keyring is now possible with Mailvelope. This would then fix #7157.

thomascube:
Question is just to which preference section that option could be added to...

thomascube (approved):

thomascube:
I suggest to merge this into master and then find the right place to add that user preference. The other changes are very valuable and should go into 1.5-beta

trasherdk:
Am I right in believing the `Mailvelope` thing is a browser extension? And that the usage is optional, as in a plugin?

ryan77627:
Yes, Mailvelope is a browser plugin that does what the Enigma plugin does but client side (PGP Encryption and Decryption). It has an API that Roundcube integrates. Usage is only activated if the Mailvelope plugin is installed and your Roundcube domain has the API option turned on.

CHANGELOG
```diff
@@ -69,6 +69,11 @@ RELEASE 1.4.4
 - Security: Fix remote code execution via crafted 'im_convert_path' or 'im_identify_path' settings [CVE-2020-12641]
 - Security: Fix local file inclusion (and code execution) via crafted 'plugins' option [CVE-2020-12640]
 - Security: Fix CSRF bypass that could be used to log out an authenticated user [CVE-2020-12626] (#7302)
+- Mailvelope: Fix size of iframe for PGP-inlined mail (#7348)
```
alecpl, line 72: Please, rebase and put these on top of the changelog.

config/config.inc.php.sample
```diff
@@ -86,3 +86,7 @@ $config['plugins'] = array(
 
 // skin name: folder from skins/
 $config['skin'] = 'elastic';
+
+// Use the Main Keyring in Mailvelope? If not, use (creating if required)
```
alecpl, line 90: Remove this. We want to have the sample config short.

program/js/app.js
```diff
@@ -4021,7 +4031,8 @@ function rcube_webmail()
       }
 
       ref.hide_message(msgid);
-      $(selector).addClass('mailvelope').children().not('iframe').hide();
+      $(selector).children().not('iframe').hide();
+      $('#messagebody').addClass('mailvelope');
```
alecpl, line 4035: You should not use skin-specific selectors. If needed add messagebody property to env via steps/mail/show.inc::rcmail_message_body().

program/js/app.js
```diff
@@ -10016,6 +10027,19 @@ function rcube_webmail()
     return 0;
   };
 
+  // check for mailvelope API
+  this.pgpmime_support_check = function(action)
+  {
+    if (typeof window.mailvelope !== 'undefined')
```
alecpl, line 10033: wouldn't be better to just do `if (window.mailvelope)` ?

alecpl (changes requested):

alecpl:
@gurnec Have you seen my last comments, will you be able to work on these?

PR #7363, opened 2020-05-01
Fix invalid handling of certain hash schemes in dovecot method

### LINKED ISSUE #7362: 'dovecot' method broken in password plugin for some hash schemes
When using the 'dovecot' method in the password plugin, it complains that the encryption function was not found, when using the argon2i (and likely certain other) hash schemes.

The problem lies in the preg_match comparison, which simply checks for false, which is an incorrect usage, according to the php documentation, as it can return a value that evaluates to false, even if a match was found.

A viable and likely more efficient solution is a call to mb_strpos, with a comparison to 0, since the method should be at the start of the string.

`if(!(mb_strpos($crypted,'{' . $method . '}') === 0)) {
                return false;
            }`

Pull request with this fix will be in shortly.

dodexahedron:
This resolves #7362 

The comparison to false is invalid, in the original function, because preg_match can return a false-equivalent value when a match was, in fact, made.

### DIFF
--- a/plugins/password/password.php
+++ b/plugins/password/password.php
@@ -712,6 +712,8 @@ class password extends rcube_plugin
                 $method = 'CRAM-MD5';
             }
 
+            $method = strtoupper($method);
+
             $spec = array(0 => array('pipe', 'r'), 1 => array('pipe', 'w'), 2 => array('file', '/dev/null', 'a'));
             $pipe = proc_open("$dovecotpw -s '$method'", $spec, $pipes);
 
@@ -729,7 +731,7 @@ class password extends rcube_plugin
             fclose($pipes[1]);
             proc_close($pipe);
 
-            if (!preg_match('/^\{' . $method . '\}/', $crypted)) {
+            if (!(mb_strpos('{' . $method . '}', $crypted) === 0) {
                 return false;
             }
 


plugins/password/password.php
```diff
@@ -712,6 +712,8 @@ static function hash_password($password, $method = '', $prefixed = true)
                 $method = 'CRAM-MD5';
             }
 
+            $method = strtoupper($method);
```
dodexahedron, line 715: This is not directly related to the bug, but it does allow this configuration directive to be case-insensitive, which seems desirable.

alecpl:
It does not make sense to me. "preg_match() returns 1 if the pattern matches given subject, 0 if it does not, or FALSE if an error occurred."

What is your real case that didn't work? What's in `$method` and what's in `$crypted`?

dodexahedron:
As described in the issue, argon2i fails reliably, for any input text.
It also fails if it is provided in all caps as ARGON2I, in the configuration.

alecpl:
I don't see how is that possible and I don't have dovecot installed. So, it would help if I know a sample `$crypted` value that causes issues.

What PHP version?

dodexahedron:
PHP 7.4
Honestly not sure what frame of mind I was in, last night...
In any case, I've extracted that code out of the method and am testing it in isolation to see what happened and why.
The behavior did not make sense to me, either, but the strpos change fixed it, last night.
I'm also closing the pull request until I've figured it out with more certainty. I'll leave the issue open.

PR #7370, opened 2020-05-03
attachment filesize includes units

johndoh:
small fix for 377239fa8e066e290b50f094c6024d0f4f6d4a79.

the line `if ($filesize == 0) {` expects `$filesize` it be an `int` but its a `string`. `$RCMAIL->message_part_size($attach_prop);` uses `rcmail::show_bytes()` which always adds at least a unit, in the case of 0 that's going to be a B.


alecpl:
Thanks, a better fix in 7ead0bb227a.

PR #7375, opened 2020-05-07
Specify 13px font-size for .message-part div.pre

rnkn:
Default monospace fonts are larger than proportional fonts and so make the plain text message look oddly large, upsetting the aesthetics of the page.

12px font sizes brings the Elastic skin into consistency with Larry and Classic.

### DIFF
--- a/skins/elastic/styles/styles.less
+++ b/skins/elastic/styles/styles.less
@@ -338,6 +338,7 @@ body.task-error-login #layout {
 
     div.pre {
         font-family: monospace;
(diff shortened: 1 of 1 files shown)


alecpl:
In my opinion 12px is too small, maybe 13px, but I'm not sure.

rnkn:
Hi Aleksander, thanks for making this great skin.
Yeah 12px is probably too small in the context of the other elements. 13px looks better.

PR #7380, opened 2020-05-10
add basic error message on fatal PHP error

johndoh:
rather than showing a blank screen on a fatal PHP error this adds a simple error message:
Untitled-1
this will catch common errors like typos in the config file which cause a parse error or missing PHP extensions such as Mbstring and give admins a hint to check the log files for info.


alecpl:
I wouldn't use "Roundcube" word on that page. How about "For administrators: Please, check the application and/or server error log for more information."?

johndoh:
Good point. I've changed the message and squashed the commits.

Note: I left the Roundcube name in the CLI error message because only admins will see that,

PR #7402, opened 2020-05-23
IMAP Servers may timeout for a login request.

hefee:
When connection fails in a recoverable way, we should retry to connect
to the server again. The code in rcube_imap.php looks, like, that is
already prepared for a retry, but `data['retry']` is not used anywhere.

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -148,6 +148,10 @@ class rcube_imap extends rcube_storage
 
         $attempt = 0;
         do {
+            if ($attempt > 0) {
+                usleep(rand(1000, 100000));
+            }
+            rcube::write_log('imap','Connecting to IMAP server attempt:' . $attempt);
             $data = $this->plugins->exec_hook('storage_connect',
                 array_merge($this->options, array('host' => $host, 'user' => $user,
                     'attempt' => ++$attempt)));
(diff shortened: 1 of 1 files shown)


alecpl:
That 'retry' item was there to give plugins opportunity to e.g. retry using another imap server or sth else. I don't think we should retry by default.

hefee:
> That 'retry' item was there to give plugins opportunity to e.g. retry using another imap server or sth else. I don't think we should retry by default.

@alecpl: I see your point, that this is code, that my solution fixes a specific issue on a very generic point and move this to a plugin makes more sense. But how do I write such a plugin? I'm happy to implement. But I see that it does not affect a lot of setups, otherwise this issue would attract more attention. I searched in previous issues and found, that also others have/had this issue, too. about imap sever, that just does not answering #5393.

alecpl:
Sample plugin code that uses `storage_connect` hook: https://git.kolab.org/diffusion/RPK/browse/master/plugins/kolab_sso/kolab_sso.php$169 (put that in `init()` method) and https://git.kolab.org/diffusion/RPK/browse/master/plugins/kolab_sso/kolab_sso.php$201

ibizara:
There's a patch issued in the Roundcube debian package found at https://packages.debian.org/buster-backports/roundcube named retry_to_reach_imap_server.patch, however this causes the issue https://github.com/roundcube/roundcubemail/issues/7754

PR #7405, opened 2020-05-26
Typos

milvito:


alecpl:
Localization need to be updated via https://www.transifex.com/roundcube/roundcube-webmail/

Issue #7224, opened 2020-02-14
Elastic: Improve mail header "detailed mode"


syzop:
With the new elastic skin you can no longer see instantly if there is 1 recipient or multiple. For every message you have to manually click on ```Details``` in order to see the recipients. Since there is no hint if there is 1 or more recipients this means you basically have to do this for *every message*. I got several complaints about that from staff and I agree with them that this is a major issue. I see it is also noted by https://github.com/roundcube/roundcubemail/issues/7222#issuecomment-586258114 and possibly elsewhere but I did not see a separate Issue for it.

I suggest to show the first 5 recipients by default and if there are more then an 'and more' or '...', like most email clients do.

This is on roundcube 1.4.2 by the way.

Perhaps the writers of the skin did not take into account that in certain (possibly even many) languages, regions or organizations the email netiquette is quite informal. Emails start simply with "Hi," and similar phrases which give no clue otherwise as to the recipient. In my country, such information is often seen as redundant as the header already shows to who the email is addressed.
Also, in our organization - and hopefully in others as well - we have clear rules as to what ```To:``` means (you are expected to respond) and ```Cc:``` (you may read and optionally reply, but there is no expectation for you to respond). By not showing out-of-the-box whether I am in To or in Cc, that too causes confusion, especially since the text of the email may very well be entirely in the singular form when there is 1 person in To and 1 person in Cc.

johndoh:
Perhaps the state of the state of the Details button could be remembered so the next time a message is previewed then the details are displayed.

alecpl:
Probably it could, but I also have some ideas to improve that widget.
1. After you click "Details" link:
    - the link should disappear, instead we could add a close icon in right-top of the div.message-partheaders element.
    - also div.header-subject should disappear (as it contains duplicate information).
2. Or we could design something completely different instead of the Details/Close button, e.g. some right/left arrow (or more/close) on side, or I don't know. But the idea is to make the whole header look better in "detailed mode".

rdacn:
I confirm many users are confused by the current "Details" link, the lack of "To" immediately visible field (especially to see at once if they are the main recipient) , and information redundancy between simple mode and "Details" (From and Date).  #7437 would be a great improvement !

rdacn:
Another complaint is that the From field only displays the username, lots of users would prefer username + email address to be displayed at once

johndoh:
[...]

PR #7437, opened 2020-06-14
elastic: message headers view changes

johndoh:
reference #7224

based on comments in the referenced issue this pr moves the "details" link up to the envelope summary line and changes the icon for expand/collapse. it also makes the status of the envelope block remembered between page loads.

Untitled-1

### DIFF
--- a/program/localization/en_US/labels.inc
+++ b/program/localization/en_US/labels.inc
@@ -338,6 +338,7 @@ $labels['alwaysallow'] = 'Always allow from $sender';
 $labels['isdraft'] = 'This is a draft message.';
 $labels['andnmore'] = '$nr more...';
 $labels['details'] = 'Details';
+$labels['summary'] = 'Summary';
 $labels['headers'] = 'Headers';
 $labels['allheaders'] = 'All headers...';
 $labels['togglemoreheaders'] = 'Show more message headers';
--- a/skins/elastic/styles/colors.less
+++ b/skins/elastic/styles/colors.less
@@ -186,6 +186,9 @@
 @color-mail-signature:              @color-black-shade-text;
 @color-mail-headers:                @color-black-shade-text;
 
+@color-messagepart-border:          @color-black-shade-bg;
+@color-messagepart-background:      #fcfcfc;
+
 @color-spellcheck-link:             @color-error;
 
 @color-table-border:                @color-layout-border;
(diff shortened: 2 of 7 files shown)


rdacn:
This update will be surely well regarded by our users, however there is still another complaint : they would like to see the username and the email address in the From field, instead of just the username

alecpl:
The envelope icon without a visible label is not that clear.

I'd try with leaving Details link where/how it was, but change the behavior so it hides the `.header-subject` element when `.headers-table` is visible and vice-versa.

johndoh:
hows this?

Untitled-1

alecpl:
Now, when the "extended" header can be used in "always on" mode, it would make sense think about improving it, e.g. get rid of the border. Also, maybe it would be better to put "All headers" ("Headers") in the same line as "Summary". I'm not sure what icon we could use. How about  `envelope` for "Headers" and `info` for Summary/Details?

johndoh:
Hows this?
Untitled-1

alecpl:
Much better. I'm not sure yet I like the icons.

alecpl:
I think that Headers link could be always visible now, i.e. also in short header mode.

johndoh:
Making the Headere link always visible is a simple CSS change but I'll leave this for now and take some more time to think about the icons.

johndoh:
What about keeping the envelope icon for details as it is currently and using the info icon for headers?
Untitled-1

alecpl:
It could work if we don't find a better option. I guess, if we replace Headers and Details icons it would work as well.

johndoh:
The info icon is only available in solid type, that's why I kept them that way around.

alecpl:
Looking at the code only... You removed some styling from `.message-partheaders` which is also used to display headers on message/rfc822 parts. I think this needs some fixes. We should probably not use that class inside `.header-envelope`. We probably don't need the `.headers-table` parent at all.

johndoh:
The display of message/rfc822 parts looks fine to me. it uses the same message template as the regular display so it has the same changes as that. unless I am misunderstanding.

About the current DOM structure with `.short-header`/`.header-envelope`/`.message-partheaders`/`.headers-table`. I was trying to keep my changes as minimal as possible but if you are happy with something a bit more radical then I could work on it.

You like the Details/Summary, Headers, Plain text/HTML buttons in a row at the bottom right (pending decision on the icons)? So this would be more about cleaning up the DOM than any more visual changes.

Issue #7414, opened 2020-06-03
Problem forwarding an email with jpg image attached received from iOS mail app


SamsamM38:
I receive an email from iOS mail app containing a jpg image attached.
I can open the image, all works normally.
When I forward this email, the jpg image attached disappear, it not appear in body either.

All works fine if the file attached is something else like a pdf file.
Roundcube version : 1.4.5

alecpl:
Provide a sample message, please.

SamsamM38:
[mail.txt](https://github.com/roundcube/roundcubemail/files/4722450/mail.txt)

alecpl:
The attachment is marked as inline. I think we have already tickets for similar issues, but I'll leave it open. I hope to have time to work on these for 1.5.

Your workaround is to use "forward as attachment" feature.

SamsamM38:
Thank you

SamsamM38:
The question is : why the problem appear only with an email from iOS app mail ?

alecpl:
Most clients do not use `Content-Disposition: inline` on the image part for messages with a plain text body and an image attachment.

SamsamM38:
If I used the "forward as attachment" feature, a file attachment containing the email appear + the jpg image file that did not appear with the simple "forward" command.

So to summarize : forward command, jpg file disappear, "forward as attachment" jpg file appear.
Call Mulder and Scully !

roberthr74:
We have the same issue using Rouncube Web client. Forward as attachment works in this case.

tomsommer:
We are also getting reports of this

lucasmarin:
@alecpl 

The problem is at compose.inc:997
```
            if ($part->disposition == 'inline' && $part->mimetype != 'application/pdf') {
                if (!$bodyIsHtml) {
                    continue;
                }

                $idx = $part->content_id ? ('cid:' . $part->content_id) : $part->content_location;

                if ($idx && isset($CID_MAP[$idx]) && strpos($message_body, $CID_MAP[$idx]) !== false) {
                    $replace = $CID_MAP[$idx];
                }
                else {
                    continue;
                }
            }
```

I saw that code above was developed by issue https://github.com/roundcube/roundcubemail/issues/7122 and removing else condition on it fix the problem on this issue. 

I will investigate it, fix on my RC locally and open a PR here.

Regards.

alecpl:
Fixed.

Neustradamus:
Linked to:
[...]

Issue #7440, opened 2020-06-17
iPhone only email attachments wont forward


roberta0967:
Hi Guys,

Odd one, so:

If a user emails in using outlook/roudcube or even a samsung mobile phone with an image/s attached. If you "forward" that email, the images will attach to the forwarding email.

However, if you email those images via an iphone using any account, and select "forward" on the email sent to you from an iphone - those images will NOT attach to the forwarding email.

I appears since cPanel upgrade to version 1.4.4 

Thanks
Robert.

alecpl:
Duplicate of #7414.

Issue #7293, opened 2020-03-19
Problem with forwarding/replying to the messages containing multiple inline images.


Bojo78:
Roundcube Webmail 1.4.3
pear/mail_mime 1.10.7

When forwarding or replying to a message containing multiple inline images, only the first image is attached multiple times. 
Original message
![image](https://user-images.githubusercontent.com/44122211/77112262-5f1e8980-6a31-11ea-8e99-0bc4c0249589.png)
Reply to the same message in the compose window
![image](https://user-images.githubusercontent.com/44122211/77112324-79586780-6a31-11ea-981c-87f843d51911.png)

Maybe is related with #7122

alecpl:
Confirmed with git-master.

alecpl:
Fixed.

Neustradamus:
Linked to:
- https://github.com/roundcube/roundcubemail/pull/9150
- https://github.com/roundcube/roundcubemail/issues/5051
- https://github.com/roundcube/roundcubemail/issues/6279
- https://github.com/roundcube/roundcubemail/issues/7117
- https://github.com/roundcube/roundcubemail/issues/7122
- https://github.com/roundcube/roundcubemail/issues/7214
- https://github.com/roundcube/roundcubemail/issues/7293
- https://github.com/roundcube/roundcubemail/issues/7414
- https://github.com/roundcube/roundcubemail/issues/7440
- https://github.com/roundcube/roundcubemail/pull/7441

Issue #7214, opened 2020-02-07
Problems previewing some attached imagesin Roundcube Version 1.4.2


informas:
Hi,

Since updating Roundcube from version 1.3.6 to 1.4.2 I have problems with 
some emails from GMAIL generated from a mobile phone with an attached image.

When viewing the email, Roundcube does not detect that it has an attachment, 
but if I reply or forward that email it does.

I have detected that the problem appears when the attachment has a 
"Content-ID:" valued header and is not used as an inline image.

For example, if the attached image is defined like this, Roundcube does not 
detect the attachment:
```
Content-Type: image / jpeg; name = "image1.jpg"
Content-Disposition: attachment; filename = "image1.jpg"
Content-Transfer-Encoding: base64
Content-ID: <f_k6cdcld40>
```

On the other hand, if it is defined without Content-ID, Roundcube does 
detect the attachment and preview the image:
```
Content-Type: image / jpeg; name = "image1.jpg"
Content-Disposition: attachment; filename = "image1.jpg"
Content-Transfer-Encoding: base64
```
A simplified version of the raw email to reproduce the problem is as follows:

```
MIME-Version: 1.0
From: FROMPERSON <fromabc@gmail.com>
Date: Fri, 7 Feb 2020 13:13:25 -0300
Subject: IM4 - with X-Attachment-Id AND with Content-ID
To: "DDESTINATION" <email@domain.com>
Content-Type: multipart/mixed; boundary="000000000000134e5e059dfeac92"
X-PMFLAGS: 570949760 0 5 YZD01IM4.CNM                       

--000000000000134e5e059dfeac92
Content-Type: multipart/alternative; boundary="000000000000134e5b059dfeac90"

--000000000000134e5b059dfeac90
Content-Type: text/plain; charset="UTF-8"

Text plain content

--000000000000134e5b059dfeac90
Content-Type: text/html; charset="UTF-8"

<div dir="ltr">HTML text content</div>

--000000000000134e5b059dfeac90--
--000000000000134e5e059dfeac92
Content-Type: image/jpeg; name="image1.jpg"
Content-Disposition: attachment; filename="image1.jpg"
Content-Transfer-Encoding: base64
Content-ID: <f_k6cdcld40>

/9j/4AAQSkZJRgABAQEASABIAAD/2wBDAAEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEB
[...]

PR #7441, opened 2020-06-17
Apple Mail send image files marked as inline even if it is not on body message (#7414)

lucasmarin:
Fix to reported issue https://github.com/roundcube/roundcubemail/issues/7414

Attached two e-mails with reported problem:

[mail.txt](https://github.com/roundcube/roundcubemail/files/4795204/mail.txt)
[Comprovante de Up Grade.txt](https://github.com/roundcube/roundcubemail/files/4795209/Comprovante.de.Up.Grade.txt)

### DIFF
--- a/program/steps/mail/compose.inc
+++ b/program/steps/mail/compose.inc
@@ -994,7 +994,8 @@ function rcmail_write_compose_attachments(&$message, $bodyIsHtml, &$message_body
                 if ($idx && isset($CID_MAP[$idx]) && strpos($message_body, $CID_MAP[$idx]) !== false) {
                     $replace = $CID_MAP[$idx];
                 }
-                else {
+                // Note: Apple Mail send image files marked as inline even if it is not on body message #7414
+                elseif (!isset($part->ctype_parameters['x-apple-part-url'])) {
                     continue;
                 }
             }


alecpl:
Thanks, we might apply that change as a quick workaround, but I think we need a more general solution for handling images that are marked as inline, but are not used in the html body.

lucasmarin:
I agreed. I don't know if this will be default behavior of apple client and there are another mail clients with that behavior too.

alecpl:
I just noticed that these messages do not have a HTML part, so checking if HTML part exists might be a better fix.

alecpl:
Fixed via ec2e713.

Neustradamus:
Linked to:
- https://github.com/roundcube/roundcubemail/pull/9150
- https://github.com/roundcube/roundcubemail/issues/5051
- https://github.com/roundcube/roundcubemail/issues/6279
- https://github.com/roundcube/roundcubemail/issues/7117
- https://github.com/roundcube/roundcubemail/issues/7122
- https://github.com/roundcube/roundcubemail/issues/7214
- https://github.com/roundcube/roundcubemail/issues/7293
- https://github.com/roundcube/roundcubemail/issues/7414
- https://github.com/roundcube/roundcubemail/issues/7440
- https://github.com/roundcube/roundcubemail/pull/7441

PR #7445, opened 2020-06-21
support providing skins for plugins inside the main skin folder

johndoh:
I have been [working on](https://github.com/roundcube/plugin-installer/pull/24) the `plugin-installer`, trying to fix the reported issues. Then I thought it would be good if skins could be delivered through composer just like plugins. I added support for a package type of `roundcube-skin` and the installer moves it into the `skins/` folder rather than the `plugins/` folder.

If someone is going to create a new skin though they will want to add skin assets for plugins. At the moment that has to be in the skins folder of a plugin. So I [made](https://github.com/roundcube/plugin-installer/pull/24/commits/3cf1462ae6a860ea8f97753a9d79106eaba741df) the `plugin-installer` copy plugin skins into the `plugins/` folder.

I think this method of copying files around is rather messy so then I thought what if skins could hold plugin assets themselves. A lot of the existing code for extended skins can be leveraged to achieve this.

This is a non-breaking change (assuming properly coded plugins), the only special case is the `jqueryui` plugin which uses a `themes` dir rather than `skins` (ref #266) so I added support for custom dir names and skin names (because of the skin mapping feature of the `jqueryui` plugin).

Any assets in the `plugins/myplugin/skins/myskin/` folder take priority but if none are found then the code checks in `skins/myskin/plugins/myplugin/`. It also works with extended skins.

### DIFF
--- a/plugins/jqueryui/jqueryui.php
+++ b/plugins/jqueryui/jqueryui.php
@@ -18,6 +18,7 @@ class jqueryui extends rcube_plugin
 
     private static $features = [];
     private static $ui_theme;
+    private static $css_path;
     private static $skin_map = [
         'larry'   => 'larry',
         'default' => 'elastic',
@@ -47,10 +48,12 @@ class jqueryui extends rcube_plugin
         $skins[]  = 'elastic';
 
         foreach ($skins as $skin) {
-            self::$ui_theme = $ui_theme = !empty($ui_map[$skin]) ? $ui_map[$skin] : $skin;
+            self::$ui_theme = !empty($ui_map[$skin]) ? $ui_map[$skin] : $skin;
+            self::$css_path = $this->local_skin_path('themes', self::$ui_theme);
 
-            if (self::asset_exists("themes/$ui_theme/jquery-ui.css")) {
-                $this->include_stylesheet("themes/$ui_theme/jquery-ui.css");
+            $css = self::$css_path . '/jquery-ui.css';
+            if (self::asset_exists($css)) {
+                $this->include_stylesheet($css);
                 break;
             }
         }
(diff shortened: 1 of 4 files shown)


thomascube:
I think this makes sense.

plugins/jqueryui/jqueryui.php
```diff
@@ -45,9 +46,11 @@ public function init()
 
         foreach ($skins as $skin) {
             self::$ui_theme = $ui_theme = $ui_map[$skin] ?: $skin;
```
alecpl, line 48: So, we don't need that local $ui_theme variable. I'd remove it everywhere.

program/lib/Roundcube/rcube_plugin.php
```diff
@@ -369,7 +369,7 @@ public function url($fn)
      */
     private function resource_url($fn)
     {
-        if ($fn[0] != '/' && !preg_match('|^https?://|i', $fn)) {
+        if ($fn[0] != '/' && !preg_match('|^https?://|i', $fn) && strpos($fn, $this->ID . '/') === false) {
```
alecpl, line 372: I think, it would be better to use `!== 0` here, instead of `=== false`.

program/lib/Roundcube/rcube_plugin_api.php
```diff
@@ -718,7 +719,7 @@ protected function template_container_hook($attrib)
      */
     protected function resource_url($fn)
     {
-        if ($fn[0] != '/' && !preg_match('|^https?://|i', $fn))
+        if ($fn[0] != '/' && !preg_match('|^https?://|i', $fn) && strpos($fn, '/' . $this->url) === false)
```
alecpl, line 722: This change looks fishy to me.

program/lib/Roundcube/rcube_plugin.php
```diff
@@ -381,21 +381,43 @@ private function resource_url($fn)
...
             $skins = (array) $rcube->config->get('skin');
         }
 
+        $dirs = array('skins');
+        if (!empty($extra_dir)) {
+            array_unshift($dirs, $extra_dir);
+        }
+
+        if (!empty($skin_name)) {
+            $skins = (array) $skin_name;
```
alecpl, line 403: Shouldn't that extend the `$skins` array, instead of replacing it? The comment says "Additional skin name".

alecpl (changes requested):

program/lib/Roundcube/rcube_plugin_api.php
johndoh: I think I can change `strpos($fn, '/' . $this->url) === false` to `strpos($fn, 'skins/') !== 0`, is that better?

program/lib/Roundcube/rcube_plugin.php
johndoh: !== will not work because of all the different possible paths e.g.:
```
skins/larry/style.css
plugin.js
js/plugin.js
skins/elastic/plugin.css
skins/elastic/plugins/plugin/plugin.css
skins/elastic/plugins/plugin/plugin.js
```
the idea is to make sure the plugin folder only gets added into the path if its not already there but I guess people could add a subfolder in their plugin folder with the same name and then this would fail.

perhaps this would be better `!preg_match("|^skins/[a-z0-9-_]+/plugins/$this->ID/|i", $fn)`?

program/lib/Roundcube/rcube_plugin.php
johndoh: the only use for this is in the jqueryui plugin which has its own skin map. its the comment which is misleading, I've corrected it.

dereks:

```
git fetch upstream
git rebase upstream/master
```

PR #7458, opened 2020-06-27
Fixed a minor typo in ko-KR.

in2etv:
** $labels['shortheaderfrom']: $to -> $from


alecpl:
Fixed on https://www.transifex.com/roundcube/roundcube-webmail/translate/#ko_KR/labels where also the `From:` label is localized

PR #7459, opened 2020-06-28
[Bugfix] Re-init plugins when changing task

pulsejet:
When roundcube initializes on a different task, plugins for just the
login task do not get initialized, and hence do not get executed when
the exec_hook call is made

One example is if check_auth fails, the login page will render,
but the plugins with 'login' task will not be fired

## Quick reproduction
1) Get any plugin with only the 'login' (and say 'logout') task that does something visible on the login page (e.g. [roundcube-oidc](https://packagist.org/packages/radialapps/roundcube-oidc), which adds a "Login with OIDC" link)
2) Login to roundcube with username and password
3) Change the first character of `roundcube_sessauth` cookie to something else (deliberately fail `check_auth`)
4) Refresh the page. The plugin will not be fired.

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -161,6 +161,11 @@ class rcmail extends rcube
             $task = asciiwords($task, true) ?: 'mail';
         }
 
+        // Re-initialize plugins if task is changing
+        if (!empty($this->task) && $this->task != $task) {
+            $this->plugins->init($this, $task);
+        }
+
         $this->task      = $task;
         $this->comm_path = $this->url(array('task' => $this->task));
 


alecpl:
It might be better to put this code in `rcmail::kill_session()`.

pulsejet:
@alecpl can do that, but this semantically makes more sense to me here, as we are _changing_ the task just below it (this is the crucial part). Putting it in `rcmail::kill_session()` makes us "assume" that since we've killed the session, the task is going to be `login` now.

What do you think?

pulsejet:
@alecpl can do that, but this semantically makes more sense to me here, as we are _changing_ the task just below it (this is the crucial part). Putting it in `rcmail::kill_session()` makes us "assume" that since we've killed the session, the task is going to be `login` now.

What do you think?

alecpl:
Indeed, I'm not 100% sure. On the other hand a plugins initialization makes sense to be "internal" to `rcmail`. It also could be put in `rcmail::set_task()`. @thomascube what do you think?

pulsejet:
Ahh yes that is perfect. I'll add a check in `rcmail::set_task()` to check if we are changing the task, and reinitialize the plugins if that is the case.

pulsejet:
Wait that might not work as expected. Our `exec_hook` call to `unauthenticated` is above the `rcmail::set_task()` call, so if a plugin with the `login` task is listening on `unauthenticated`, then it won't work (correct me if I'm wrong).

alecpl:
We're changing behavior, so there might be some BC breaks for some plugins, but... I think the impact should be small. Most authentication plugins do not bind to any task. And if there's a plugin that uses 'unauthenticated' hook, it should really not be bound to any task too. If it was binding to 'logout' task only then it might be a problem for that plugin.

Btw, let's not forget about plugins that have no task binding. Will we re-initialize them too?

pulsejet:
If that is the case then it makes most sense to put in `rcmail::set_task()`. I'll do that shortly.

Plugins that have no task binding won't be re-initialized, since we check if `plugins_initialized` contains the plugin first
https://github.com/roundcube/roundcubemail/blob/279ae66120884262c419528b72e7aa4b908e0514/program/lib/Roundcube/rcube_plugin_api.php#L99

pulsejet:
Moved to `rcmail::set_task()`

pulsejet:
@alecpl does this look okay now?

Issue #7456, opened 2020-06-26
Setting 'login' in config['skin_logo'] changes also the favicon accordingly


gianlucagiacometti:
To reproduce the issue (only in the login page):

```
$config['skin_logo'] = array(
  'login' = '[a valid image]'
);
```

sets the logo in the login page but also the favicon with the same image

or with

```
$config['skin_logo'] = array(
  'login' = 'whatever which is not a valid image, also this text, but not an empty string'
);

```
No favicon displayed, the alt text 'Logo' shows.

No way to visualise the correct favicon when login is set in the array.

Last version in master branch used (downloaded today)

johndoh:
> No way to visualise the correct favicon when login is set in the array.

Set the logo for `favicon` type e.g.:

```
'[favicon]' => '/my/favicon.png'
```

As for matching the logo template this may be a regression caused by 75125cbac16b6ad714e3bd019f218e1d1794f9e6 which removed the strict matching param. I'm not certain of the best approach. This line:
https://github.com/roundcube/roundcubemail/blob/98881a0cbd74936240d7f9afcba7223d60d64241/program/include/rcmail_output_html.php#L2517-L2520
Could be changed to:
```php
$template_names = preg_grep("/\]$/", $template_names);
```
so that only entries matching the specified type are used. That makes sense for the `[favicon]` and I think also for `[small]` the only one I'm not certain of is `[print]` if I specify a logo for a skin then I might expect the print logo to also change without additional config.

alecpl:
I'm also not sure about this.

BTW, @johndoh,  isn't the regexp wrong in `$template_names = preg_grep("/\*$/", $template_names, PREG_GREP_INVERT);`? There's only one name that ends with a star. Also, the other regexp, has inconsistent use of backslash.

johndoh:
I think when I did 75125cb I got that line wrong and that it does not make sense any more. Like you said there is only 2 entries which end in a * and thinking about it today removing only those when a logo type is specified seems arbitrary and kind of pointless.

I think that line could be removed completely or replaced with something that limits the matches to those that specify the type. Both approaches have advantages and disadvantages.

May be I need to bring back the strict matching parameter or perhaps allow the logo object to specify multiple types so it define a fallback.

johndoh:
[...]

PR #7470, opened 2020-07-04
fix skin_logo config, again

johndoh:
I apologise for how much trouble the skin_logo config option has been since i first modified it.

I've fixed the bug raised in #7456 by adding a new param which accounts for the special case of the print icon and allows for similar cases in future. I've also added a bunch more tests to make sure all the different uses are properly covered.

### DIFF
--- a/skins/elastic/templates/includes/menu.html
+++ b/skins/elastic/templates/includes/menu.html
@@ -1,7 +1,7 @@
 <div id="layout-menu">
 	<h2 id="aria-label-tasknav" class="voice"><roundcube:label name="arialabeltasknav" /></h2>
 	<div class="popover-header">
-		<roundcube:object name="logo" src="/images/logo.svg" data-src-small="0" id="logo" alt="Logo" />
+		<roundcube:object name="logo" src="/images/logo.svg" alt-logos="dark,small,small-dark" id="logo" alt="Logo" />
 		<a class="button icon cancel"><span class="inner"><roundcube:label name="close" /></span></a>
 	</div>
 	<div id="taskmenu" class="menu toolbar" role="navigation" aria-labelledby="aria-label-tasknav">
--- a/skins/elastic/templates/includes/menu.html
+++ b/skins/elastic/templates/includes/menu.html
@@ -1,7 +1,7 @@
 <div id="layout-menu">
 	<h2 id="aria-label-tasknav" class="voice"><roundcube:label name="arialabeltasknav" /></h2>
 	<div class="popover-header">
-		<roundcube:object name="logo" src="/images/logo.svg" alt-logos="dark,small,small-dark" id="logo" alt="Logo" />
+		<roundcube:object name="logo" src="/images/logo.svg" id="logo" alt="Logo" />
 		<a class="button icon cancel"><span class="inner"><roundcube:label name="close" /></span></a>
 	</div>
 	<div id="taskmenu" class="menu toolbar" role="navigation" aria-labelledby="aria-label-tasknav">
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -406,7 +406,8 @@ $config['support_url'] = '';
 // - logo type - it is used for logos used on multiple templates
 //   the available types include '[favicon]' for favicon, '[print]' for logo on all print
 //   templates (e.g. messageprint, contactprint) and '[small]' for small screen logo in supported skins
-//   '[dark]' and '[small-dark]' for dark mode logo in supported skins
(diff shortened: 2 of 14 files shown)


skins/elastic/templates/includes/menu.html
```diff
@@ -1,7 +1,7 @@
 <div id="layout-menu">
 	<h2 id="aria-label-tasknav" class="voice"><roundcube:label name="arialabeltasknav" /></h2>
 	<div class="popover-header">
-		<roundcube:object name="logo" src="/images/logo.svg" data-src-small="0" id="logo" alt="Logo" />
+		<roundcube:object name="logo" src="/images/logo.svg" alt-logos="dark,small,small-dark" id="logo" alt="Logo" />
```
alecpl, line 4: Do we need `alt-logos` attribute at all?

johndoh: TBH I'm not sure. I thought this way allowed skins to define which alt types they support.

johndoh: Alternatively I could add something like `supported_logo_types` in meta.json and not have this attribute

alecpl:
I'd prefer to have the watermark thing separated. I'm not sure this is the best way to go with watermark images. Maybe we should rather make the watermark.html a template (with normal img element) parsed by the templates engine.

gianlucagiacometti:
I agree with @alecpl on the watermark.
At the moment the best way I found to customise it is to rewrite in a temp dir the watermark.html with the customised image and change the env var watermark accordingly. Having the image directly customisable in the core would be wonderful.

johndoh:
In the past Alec has said he didn't want to move `watermark.html` into the template system and it is already possible for a plugin to replace `watermark.html` with anything it likes. I saw the changes made for dark mode support and I thought it offered an opportunity to change the logo without the need for `watermark.html` to use the template system.

My thinking was that people can replace the Roundcube logo with any image they like and not only by config but also simply by replacing `logo.svg` with their own file but what looks good on a light background does not necessarily look good on a dark one and so it would be useful to be able to change the logo on the `watermark.html` screen for light/dark modes just like in the rest of Roundcube (eg. login screen vs mail screen, light vs dark backgrounds). I totally understand there are different ways of achieving this and dark mode in Roundcube is very new and still a work in progress and as things develop better approaches can emerge.

gianlucagiacometti:
It's true that dark mode is very new, but it's a sort of standard now, which will last for a long time.
The simplest way might be adding one field to the existing skin_logo array

```
$config['skin_logo'] = array(
  "elastic:dark:*" => "/images/logo_dark.png",
  "elastic:light:*" => "/images/logo_light.png",
  "elastic:*:*[small]" => "/images/logo_small_dark and_light.png",
  ...
  );
```
Old skins can simpoly use * instead of dark or light

Moreover, when your server hosts more than one domain, it would be nice to customise logo and watermark also per domain. So a further improvement might be
```

$config['skin_logo'] = array(
  "domainA:elastic:dark:*" => "/images/logoDomainA_dark.png",
  "domainA:elastic:light:*" => "/images/logoDomainA_light.png",
  "domainB:elastic:dark:*[small]" => "/images/logoDomainB_small_dark.png",
  "*:elastic:*:*" => "/images/default_logo_for all_domains_dark_and_light.png",
  ...
  );

```

alecpl:
I'm not yet 100% happy with current way the dark mode is implemented. E.g. with devel_mode=0 there's visible blinking effect on page load (especially on pages with forms). As for the watermark, the fact that we apply some fancy css rules on the logo in watermark.html does not make it simpler. In the past I was against making it a template, but I think this is still on the table. I don't have my preference yet, though.

johndoh:
Your first part about the light and dark. This PR makes that possible everywhere except `watermark.html` without the need to change the existing syntax.

The second part I disagree with i think that makes the `skin_logo` option differnt to every other config option and that makes it harder to understand when such setups probably already have handling for multiple domains in place.

alecpl:
Also, the fact that we extended `skin_logo` setting in master moves slightly the balance in the way I think about the watermark.

johndoh:
Another approach I though of was to include a CSS file in `watermark.html` and then the logo could come from a variable in the Less, and could be customised thanks to `_variables.less` but I have not tried it yet so I'm not sure if its practial or not.

gianlucagiacometti:
I like the idea of the css. It would allow good manipulation.
I also perfectly understand to keep the configuration as simplest as possible. Customising by domain may stay in charge of a plugin.

alecpl:
@johndoh Looks good. For backward compatibility I'd keep the attribute name `type` instead of `logo-type`.

PR #7483, opened 2020-07-10
solves concatenation problem between help_source and help_index_map

ghost:
Hi, roundcube community.

I've been experiencing some problems with the help plugin. 
The plugin was wrongly concatenating the help_source content with the help_index_map. 
I solved it this way, do you think it can be a plausible solution to the problem?


alecpl:
Please, explain the problem in more detail. What did you do and what do you expect?

alecpl:
No feedback.

PR #7491, opened 2020-07-16
improve temp_dir config parameter description

thomas-mc-work:


alecpl:
What cache files?

thomas-mc-work:
Here is my source: https://github.com/roundcube/roundcubemail/wiki/Dev-Docs#directory-structure

PR #7496, opened 2020-07-18
fix regression in message list col type detection

### LINKED ISSUE #7495: Subject_col regression
I didn't dig deep, but I found a regression where subject_col value in list widget was wrong (on mail messages list). I think this is caused by some change in 279ae661208842. I fixed one issue in 2de9d11070264, but looking at the code, there might be more issues that might be not that obvious. I'm talking about all `subject_column()` uses in `list.js`.

@johndoh would you like to take a look at this?

johndoh:
fix #7495

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -255,6 +255,9 @@ function rcube_webmail()
         this.enable_command('list', 'checkmail', 'add-contact', 'search', 'reset-search', 'collapse-folder', 'import-messages', true);
 
         if (this.gui_objects.messagelist) {
+          // setup message list cols
+          this.msglist_setup(this.env.layout);
+
           this.env.widescreen_list_template = [
             {className: 'threads', cells: ['threads']},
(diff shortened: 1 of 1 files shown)


program/js/app.js
```diff
@@ -2216,12 +2228,14 @@ function rcube_webmail()
         + (flags.flagged ? ' flagged' : '')
         + (message.selected ? ' selected' : ''),
       row = { cols:[], style:{}, id:'rcmrow'+msg_id, uid:uid },
-      listcols = layout == 'widescreen' ? this.env.listcols_widescreen : this.env.listcols;
+      listsetup = this.msglist_setup(),
```
alecpl, line 2231: That's probably not that big overhead, but do we need to do this on every list record? We call the method in set_message_coltypes() and set_list_options(). Maybe it would be enough to set some env variables and use them here?

johndoh:
I've put the cols and layout back into the `env`. I've checked it works adding/removing cols, shifting col order and switch layout modes.

PR #7497, opened 2020-07-19
elastic: make custom less files optional

johndoh:
Optional `@imports` where added in Less v2 which was released 5 years ago. So by bumping the min version up from 1 to 2 (v3 is already used in jsdeps) then the customisation files `_styles` and `_variables` can be renamed as dist files so they don't conflict with what people might have in their local repos.

the `-x` compression option does not work properly in current version of lessc so I have also update the readme to mention the clean-css plugin and changed the extension of the generated files to `.min.css` so that cssshrink.sh skips the already minified files.


alecpl:
1. I think that we can get rid of the .dist files. Just describe them in README.
2. After I upgraded less in my system, your commands give: "Unable to load plugin clean-css please make sure that it is installed under or at the same level as less". So, this is an additional complexity. Does clean-css minify the file better that cssshrink.sh? If it doesn't, maybe we should just not use that option, and minify with cssshrink.sh.

johndoh:
This is trivial but the current version of cssshrink.sh does not work with Elastic because Elastic CSS is in the styles folder not the root of the skin.

to fix you plugin error try running `npm install -g less-plugin-clean-css`.

minifying test on styles.css:
-x: 118,312B
cssshrink.sh: 118,291B
clean-css: 116,647B

alecpl:
OK. We can use clean-css, but please describe installation of less and clean-css in the README. We also have to install proper version for Travis.

johndoh:
I have added a little more detail to the note in the readme file. do you think it is enough?

But I'm stumped by Travis, may be this change has to wait until the Travis tests can be moved to a more modern distro, it looks like Ubuntu Trust has a very old node-less package.

alecpl:
Instead of installing the node-less package we could just run: `npm install --force -g less less-plugin-clean-css`.

That note should be enough.

johndoh:
Thanks for the tip, that did the trick. I left out the clean-css plugin because the Travis script does not minify the css.

skins/elastic/README.md
```diff
@@ -28,14 +28,15 @@ INSTALLATION
...
-    $ lessc -x styles/styles.less > styles/styles.css
-    $ lessc -x styles/print.less > styles/print.css
-    $ lessc -x styles/embed.less > styles/embed.css
+    $ lessc --clean-css="--s1 --advanced" styles/styles.less > styles/styles.min.css
+    $ lessc --clean-css="--s1 --advanced" styles/print.less > styles/print.min.css
+    $ lessc --clean-css="--s1 --advanced" styles/embed.less > styles/embed.min.css
 ```
-(the -x option minifies the CSS code)
+(`--clean-css="--s1 --advanced"` minifies the css, requires the clean-css Less plugin.
+The can be installed using `npm install less-plugin-clean-css`)
```
alecpl, line 39: "It can be..."

alecpl (changes requested):
Please, remove these .dist files and add a note to README.

skins/elastic/README.md
johndoh: fixed

skins/elastic/README.md
```diff
@@ -29,14 +29,14 @@ INSTALLATION
 
 All styles are written using LESS syntax. Thus it needs to be compiled
 using the `lessc` (>= 2.5.2) command line tool. This comes with the `nodejs-less`
-RPM package which depends on nodejs.
+RPM package or using `npm install less` which depend on nodejs.
 ```
     $ lessc --clean-css="--s1 --advanced" styles/styles.less > styles/styles.min.css
     $ lessc --clean-css="--s1 --advanced" styles/print.less > styles/print.min.css
     $ lessc --clean-css="--s1 --advanced" styles/embed.less > styles/embed.min.css
```
trasherdk, line 36: Shouldn't those double-quotes be single-quotes?

`lessc --clean-css='--s1 --advanced' styles/embed.less > styles/embed.min.css`

To prevent shell expansion.

skins/elastic/README.md
```diff
@@ -28,14 +28,15 @@ INSTALLATION
...
 All styles are written using LESS syntax. Thus it needs to be compiled
-using the `lessc` (>= 1.5.0) command line tool. This comes with the `nodejs-less`
-RPM package which depends on nodejs.
+using the `lessc` (>= 2.5.2) command line tool. This comes with the `nodejs-less`
+RPM package or using `npm install less` which depend on nodejs.
 ```
-    $ lessc -x styles/styles.less > styles/styles.css
-    $ lessc -x styles/print.less > styles/print.css
-    $ lessc -x styles/embed.less > styles/embed.css
+    $ lessc --clean-css="--s1 --advanced" styles/styles.less > styles/styles.min.css
```
thomas-mc-work, line 34: What do the clean-css switches mean? I couldn't find them in their documentation.

johndoh: Look a couple of lines down, its the config for the clean-css Less plugin which does the minifying

PR #7503, opened 2020-07-23
distinctive icon for messages both replied to and forwarded

pajod:
**what was broken?**
the message view does not indicate the difference between
1) a message replied to and
2) a message both replied to and forwarded

**How can we fix it?**
1) chose a separate icon from the fontawesome catalog (as proposed in #5004)
2) display both icons at once

**Why prefer fix 1)?**
because just placing the two icons on top of each other is less readable

**do we need documentation updates?**
no, the only documentation for this feature is the title (mouseover text) and that is correct already.

**regression potential?**
the new icon is slightly less self-explanatory than those already in use.

Keywords: \Answered $Forwarded

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -39,6 +39,7 @@ CHANGELOG Roundcube Webmail
 - Fix so anchor tags without href attribute are not modified (#7413)
 - Fix support for an error as a string in message_before_send hook (#7475)
 - Elastic: Fix redundant scrollbar in plain text editor on mail reply (#7500)
+- Elastic: distinctive icon for messages both replied to and forwarded (#7503)
 
 RELEASE 1.4.7
 -------------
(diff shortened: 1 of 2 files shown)


alecpl:
I don't like the icon. How about we create a custom icon, e.g.
reply-forward
I did it by merging paths from both svg icons and giving `style="opacity:.5"` to one of them. I think that's better than your proposal, but if anyone can create a better option I'll consider it.

gianlucagiacometti:
I like the custom icon @alecpl proposed

alecpl:
Unfortunately with a custom svg file used as content/background we have no control over colors. So, this is not a good solution until someone could convert it to a font-icon.

Fortunately, it is possible to end up with the look I proposed using `:before` and `:after` pseudo-elements. I'm working on this.

alecpl:
Fixed with 56b73525978.

PR #7510, opened 2020-07-26
Tinycp driver

rimendoz86:
TinyCP is a small linux CP that recently added RoundCube to it's installer.  This allows the password driver to send password updates to the TinyCP end point.

### DIFF
--- a/plugins/password/drivers/tinycp.php
+++ b/plugins/password/drivers/tinycp.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * tinycp driver
+ *
+ * Enable the password driver in Roundcube (https://roundcube.net/) for the TinyCP Lightweight Linux Control Panel (https://tinycp.com/).
+ *
+ * @version 1.1
+ * @author Ricky Mendoza (HelloWorld@rickymendoza.dev)
+ * 
+ * Copyright (C) 2020 Ricky Mendoza
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


alecpl:
What is the license of TinyCPConnector? If possible, it would be better to include (about 200 lines) connector's code into the driver file, so the setup is easier.

Please, remove `echo` call from the code.

plugins/password/drivers/tinycp.php
```diff
@@ -0,0 +1,64 @@
...
+        $tinycp_port = rcmail::get_instance()->config->get('tinycp_port');
+        $tinycp_user = rcmail::get_instance()->config->get('tinycp_user');
+        $tinycp_pass = rcmail::get_instance()->config->get('tinycp_pass');
+        $error_message = '';
+        
+        if ($tinycp_host && $tinycp_port && $tinycp_user && $tinycp_pass){
+            try {
+                $tcp = new TinyCPConnector($tinycp_host,$tinycp_port);
+                $tcp->Auth($tinycp_user, $tinycp_pass); 
+                $error_message = $tcp->mail___mailserver___email_pass_change2($username, $newpass);
```
alecpl, line 43: Looking at the connector code, it could return a non-null result also on the success response. I'd propose to not use the method result at all. In case of an error an exception is thrown.

rimendoz86: I did't see any response coming from their email method. Regardless, you're right in the fact that an exception is thrown on error, its not necessary to use the response. I'll remove that and the echo. I'll reach out in the their discord to see if they are okay with me including that code in the driver. If they are okay with it, then yeah it would be a lot easier and shouldn't take long to do.

PR #7517, opened 2020-07-29
Send IMAP ID only after authentication

ledgr:
In case of IMAP proxy, that is doing authentication and routing, receiving ID command may result in `BAD Invalid command` error, as IMAP proxy does not yet know which server should receive the communications. Even if proxies would return correct output for ID command, the "name" field returned would not necessarily be the same as for "name" field value after authenticating.

This applies to Nginx + Dovecot stack.

iPhone Mail, Mac OS X Mail, Microsoft Outlook, Thunderbird, and other major (as 30+ minor) MUAs are sending ID command after authenticating, or even delay it until after NAMESPACE.

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -136,14 +136,7 @@ class rcube_imap extends rcube_storage
 
         if ($this->options['debug']) {
             $this->set_debug(true);
-
-            $this->options['ident'] = array(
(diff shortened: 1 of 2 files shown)


alecpl:
I get the idea, but there's a plugin that sets `ident` property in `storage_connect` hook. We should keep that possibility.

alecpl:
Implemented in 5975742bcb37.

PR #7520, opened 2020-07-29
Update labels.inc

tw-:
Typo


alecpl:
I fixed that on Transifex. It will be merged with the next localizations update.

PR #7529, opened 2020-08-02
MySQL: Don't reissue query after deadlock error

### LINKED ISSUE #7528: Transaction isolation broken with MySQL
On MySQL with InnoDB, situations may arise in concurrent transactions where locks acquired by transactions conflict such that MySQL detects a deadlock and rolls back one or more of the participating transactions until the deadlock situation is resolved.

[MySQL Manual](https://dev.mysql.com/doc/refman/8.0/en/innodb-deadlock-detection.html):
> When deadlock detection is enabled (the default), InnoDB automatically detects transaction deadlocks and rolls back a transaction or transactions to break the deadlock.

At this point, the transactions chosen for rollback my the DBMS  have been rolled back and terminated. The error should be signaled to the application, which can decide whether to restart the _entire_ transaction from scratch or leave it be.

Roundcube, however, for this specific error would re-issue the last statement that raised the deadlock error up to 2 times before handing the error up to the application (in my case, the _application_ is the rcmcarddav plugin).

See `program/lib/Roundcube/db/mysql.php`
```php
    protected function handle_error($query)
    {
        $error = $this->dbh->errorInfo();

        // retry after "Deadlock found when trying to get lock" errors
        $retries = 2;
        while ($error[1] == 1213 && $retries >= 0) {
            usleep(50000);  // wait 50 ms
            $result = $this->dbh->query($query);
            if ($result !== false) {
                return $result;
            }
            $error = $this->dbh->errorInfo();
            $retries--;
        }

        return parent::handle_error($query);
    }
```

Now this is bad, because
1. it continues executing statements in the middle of a transaction that has been rolled back, potentially leaving the data in an inconsistent state
2. the application gets no indication whatsoever that something went wrong, and thus continues to execute one-by-one without being guarded by a transaction

I found this using a test with a simple lost update problem (please don't try to make sense of it, I just used some database field for the sake of the test). It reads a string from a database row, and appends a client-specific suffix to the string. The transactions interleave such that both transactions read the old value from the database and afterwards issue their updates. The transactions are set to use Serializable isolation level, so the result must be as if they executed in sequence (or some may be rolled back at the option of the DBMS).

Each client tries the following transaction:
```sql
SET SESSION TRANSACTION ISOLATION LEVEL SERIALIZABLE;
START TRANSACTION READ WRITE;
SELECT id,filename FROM rckrb_carddav_migrations WHERE `filename` LIKE 'UNITTEST-SYNC%';
-- 58 is the ID of the record as returned by select. XYZ is replaced with a client-specific suffix (PAR or CLD).
UPDATE rckrb_carddav_migrations SET filename='UNITTEST-SYNC-XYZ' WHERE `id` = '58';
COMMIT
```

And here is the MySQL server log:
```
2 Connect   roundcube14@localhost on roundcubemail14krb using Socket 
2 Query INSERT INTO rckrb_carddav_migrations(filename) VALUES ('UNITTEST-SYNC')
2 Query SET SESSION TRANSACTION ISOLATION LEVEL SERIALIZABLE
2 Query START TRANSACTION READ WRITE
3 Connect   roundcube14@localhost on roundcubemail14krb using Socket 
3 Query SET SESSION TRANSACTION ISOLATION LEVEL SERIALIZABLE
3 Query START TRANSACTION READ WRITE
3 Query SELECT id,filename FROM rckrb_carddav_migrations WHERE `filename` LIKE 'UNITTEST-SYNC%'
2 Query SELECT id,filename FROM rckrb_carddav_migrations WHERE `filename` LIKE 'UNITTEST-SYNC%'
2 Query UPDATE rckrb_carddav_migrations SET filename='UNITTEST-SYNC-PAR' WHERE `id` = '58'
3 Query UPDATE rckrb_carddav_migrations SET filename='UNITTEST-SYNC-CLD' WHERE `id` = '58'
3 Query UPDATE rckrb_carddav_migrations SET filename='UNITTEST-SYNC-CLD' WHERE `id` = '58'
2 Query COMMIT
3 Query COMMIT
3 Quit  
2 Query SELECT * FROM rckrb_carddav_migrations WHERE `id` = '58'
2 Query DELETE FROM rckrb_carddav_migrations WHERE `filename` LIKE 'UNITTEST-SYNC%'
2 Quit 
```

Now as you can see, client `3` issues the UPDATE statement a second time. This is because the first update failed with a deadlock error, and roundcube reissues the update. Since the second update is executed without the scope of a transaction, it succeeds. However, the resulting data in the DB is inconsistent from the viewpoint of the application, for which it looks as if:
- Both transactions succeed, value of filename is UNITTEST-SYNC-CLD.

Expected/acceptable behavior from viewpoint of application (with one transaction failing being what typically happens):
- Both transactions fail (rolled back), value of filename stays at UNITTEST-SYNC.
- Transaction of client 2 fails, value of filename is at UNITTEST-SYNC-CLD.
- Transaction of client 3 fails, value of filename is at UNITTEST-SYNC-PAR.
- Both transactions succeed, value of filename is either UNITTEST-SYNC-PAR-CLD or UNITTEST-SYNC-CLD-PAR.

mstilkerich:
Please see issue for a detailed description. (Fixes #7528)

When a deadlock error occurs, MySQL rolls back the _entire_ ongoing
transaction. The application would have to restart the transaction from
the beginning to retry if wanted.

Now roundcube in that case simply retries the statement that raised the
deadlock error, even though everything else in the transaction was
rolled back. This breaks transaction isolation, leaves the database in a
potentially inconsistent state (from the application's data consistency
model point of view), and does not indicate to the application there was
a problem at all to it may even continue executing other potentially
remaining statements of a transaction that has already been rolled back.

-> Remove the retries from roundcube. Retries have to be initiated by
the application if desired, not only retrying the last statement but the
entire transaction from the beginning.

### DIFF
--- a/program/lib/Roundcube/db/mysql.php
+++ b/program/lib/Roundcube/db/mysql.php
@@ -247,29 +247,4 @@ class rcube_db_mysql extends rcube_db
             . " ON DUPLICATE KEY UPDATE $update", $values);
     }
 
-    /**
-     * Handle DB errors, re-issue the query on deadlock errors from InnoDB row-level locking
-     *
-     * @param string Query that triggered the error
-     * @return mixed Result to be stored and returned
-     */
-    protected function handle_error($query)
-    {
-        $error = $this->dbh->errorInfo();
-
-        // retry after "Deadlock found when trying to get lock" errors
-        $retries = 2;
-        while ($error[1] == 1213 && $retries >= 0) {
-            usleep(50000);  // wait 50 ms
-            $result = $this->dbh->query($query);
-            if ($result !== false) {
-                return $result;
-            }
-            $error = $this->dbh->errorInfo();
-            $retries--;
-        }
-
-        return parent::handle_error($query);
-    }
-
 }


alecpl:
I'm not sure how important is that code. Maybe we should skip the code only if inside a transaction.

@thomascube ?

thomascube:
Roundcube itself rarely uses transactions and thus this problem never arised so far. The retries can help if not initially executed in a transaction. Therefore skipping this recovery attempt if a transaction is active, would be the least intrusive change.

mstilkerich:
Just out of curiosity, did you ever observe a deadlock error because of single query transactions or what was the reason you added this code in the first place? I can imagine that lock acquisition may overlap even for two racing single queries, but the database should be able to do the re-execution itself in autocommit mode.

alecpl:
@thomascube Thinking of it more..., aren't we risking that outdated information is inserted to the database if we retry the query after deadlock? Maybe this code should indeed be removed completely. So, I'd also like to hear more about the use-case.

thomascube:
@alecpl I can't tell you anything about the use-case. Too long ago, I have no memory of that.

PR #7534, opened 2020-08-06
Fix limit information error about the max size of the attachment

ghost:
In this way, in the error message, the user can see the correct value of the max size loadable


alecpl:
See #7342.

miaulalala:
Superseeded by https://github.com/roundcube/roundcubemail/pull/7342

PR #7551, opened 2020-08-13
Add variable in app.js to setup the ajax timeout upload file

ghost:
In this way, we can control the ajax timeout on client side when we upload attachments

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -60,6 +60,7 @@ function rcube_webmail()
   this.identifier_expr = /[^0-9a-z_-]/gi;
 
   // environment defaults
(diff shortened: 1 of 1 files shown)


program/js/app.js
```diff
@@ -60,6 +60,7 @@ function rcube_webmail()
   this.identifier_expr = /[^0-9a-z_-]/gi;
 
   // environment defaults
+  this.uploadTimeout = 0 * 1000; // 0 = no timeout | ajax call timeout for loading attachment
```
alecpl, line 63: Please, move it up to the "webmail client settings" section, and remove ` * 1000`.

PR #7560, opened 2020-08-18
Update en_GB.inc

HostOncilla:
Add missing language variable.


alecpl:
First, there's no need to add labels/messages that are identical to en_US. Second, this must be done via Transifex.

Issue #7567, opened 2020-08-23
Closure compiler is no longer available at dl.google.com


X-dark:
Closure compiler download URL referenced here: https://github.com/roundcube/roundcubemail/blob/8e63309203788f4646cc7bf98b858f73fea84f98/bin/jsshrink.sh#L6 is no longer available. It should be downloaded through [Maven](https://mvnrepository.com/artifact/com.google.javascript/closure-compiler) or [GitHub](https://github.com/google/closure-compiler/releases). I am not sure if there is a way to download "latest" as before though.

alecpl:
I think for a while we can live with https://dl.google.com/closure-compiler/compiler-20200719.zip

According to https://github.com/google/closure-compiler/wiki/Binary-Downloads there will be no more binary downloads. In the future we might switch to npm. Or look for other options.

PR #7580, opened 2020-08-28
bin: set error on exit for shell scripts

filippog:
I am building Roundcube via CI jobs, and also ran in #7567 (thank you
for the quick fix!).

The CI job I'm using runs css/js shrink and the build job itself
succeeded, although Roundcube wasn't working.

With this commit all shell scripts will propagate the exit code in case
of errors, e.g. failed downloads.

### DIFF
--- a/bin/cssshrink.sh
+++ b/bin/cssshrink.sh
@@ -1,4 +1,7 @@
 #!/bin/sh
+
+set -e
+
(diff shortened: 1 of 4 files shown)


filippog:
I didn't change `.ci` as I'm not sure how to test it, happy to followup with that too if needed!

PR #7592, opened 2020-09-07
Fixed emails dropdown hiding when the user scrolls it.

i-introvert:
The emails dropdown window from the field "From" on the Compose-page hides when the user scrolls it. And it doesn't matter how he scrolls - by the mouse wheel or clicking on the scrollbar.

emails_dropdown


alecpl:
Thanks. I fixed this in a slightly different way.

PR #7599, opened 2020-09-11
Allow language parameter to be passed to read_localization

### LINKED ISSUE #7598: Allow language parameter to be passed to read_localization
Hi,

for my plugin to work, I had to do this to get the localization texts in a specific language from a skin:

```
$lang = 'en_US'; // Some language from $_GET
$oldLang = $_SESSION['language'];
$_SESSION['language'] = $lang;
$texts = $rcmail->read_localization(__DIR__.'/../../skins/'.$rcmail->config->system_skin.'/localization');
$_SESSION['language'] = $oldLang;
```

This is because there is no way to pass a `$lang` parameter to the function "read_localization":

```
    /**
     * Read localized texts from an additional location (plugins, skins).
     * Then you can use the result as 2nd arg to load_language().
     *
     * @param string $dir Directory to search in
     *
     * @return array Localization texts
     */
    public function read_localization($dir)
    {
        $lang   = $_SESSION['language'];
        $langs  = array_unique(array('en_US', $lang));
        $locdir = slashify($dir);
        $texts  = array();
        ....
```

Would it be possible to just add a `$lang` parameter and use it if it's set (Otherwise fallback to `$_SESSION`)?

codegain:
Allows for a language to be passed to `read_localization`.

Fixes #7598

### DIFF
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -695,13 +695,16 @@ class rcube
      * Read localized texts from an additional location (plugins, skins).
      * Then you can use the result as 2nd arg to load_language().
      *
-     * @param string $dir Directory to search in
+     * @param string      $dir  Directory to search in
+     * @param string|null $lang Language code to read
      *
      * @return array Localization labels/messages
      */
-    public function read_localization($dir)
+    public function read_localization($dir, $lang = null)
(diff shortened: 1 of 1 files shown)


codegain:
Test failures seem to be unrelated

PR #7619, opened 2020-09-21
make auto generated button ids unique between main screen and iframe

johndoh:
when using the Elastic skin on small/phone screens buttons can be copied from the main screen to the content iframe. Eg. list options are copied to the contact view/edit screens.

this change guarantees that auto-generated button ids are unique between the main window and the content iframe so when if buttons are cloned between the two they will still have unique ids.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1643,8 +1643,12 @@ EOF;
         }
 
         if (!$attrib['id']) {
-            $attrib['id'] =  sprintf('rcmbtn%d', $s_button_count++);
+            // ensure auto generated IDs are unique between main window and content frame
+            // Elastic skin duplicates buttons between the two on smaller screens (#7618)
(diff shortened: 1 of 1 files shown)


program/include/rcmail_output_html.php
```diff
@@ -1643,8 +1643,10 @@ public function button($attrib)
         }
 
         if (!$attrib['id']) {
-            $attrib['id'] =  sprintf('rcmbtn%d', $s_button_count++);
+            $prefix       = ($this->framed || $this->env['framed']) ? 'rcmfbtn' : 'rcmbtn';
+            $attrib['id'] = sprintf($prefix . '%d', $s_button_count++);
```
alecpl, line 1647: Not a big issue, but if we're using sprintf(), I'd prefer '%s%d'.

johndoh:
If possible please could this change also be back ported to1.4?

alecpl:
Done.

PR #7629, opened 2020-09-23
Update min PHP version in installer and remove from Travis

johndoh:
following 1cc9d4f update the min php version check in the installer and remove Travis tests on version 5.4


alecpl:
Right. Please, note that I want to keep "unofficial" compatibility with PHP 5.4, so please don't bring any changes in the code that would break that.

PR #7633, opened 2020-09-26
Update to auto_create_user comment

mandza:
There is a lot of misunderstand about auto creating new user, as it would create new user for everyone it the public, when it is just creating email account for localy created users.


alecpl:
For me the added sentence does not clarify the thing at all. I'll try to rephrase the whole description.

mandza:
I would appreciate that. I tried to keep sentence simple but apparently failed :)

alecpl:
Done

PR #7644, opened 2020-09-30
Add third gender option: 'other'

RobinvanderVliet:
This pull request adds a third option for gender, "other", besides the current options "male" and "female", when adding or editing a contact.

Google, Facebook and some other services now also provide a third option when creating an account.


alecpl:
I think it would be good to support that in some standards compatible way, i.e. as vCard v4 implements that.

We should at least make import/export using vCard work using X-GENDER, which is not really a standard, so we can probably put 'other' option there.

I'd prefer vCard v4 approach.

Issue #7647, opened 2020-10-01
Incorrect folders settings


abshoff:
If a special folder is located within some subfolder, then the folders settings view is broken. It adds a virtual folder for an existing folder which itself is not displayed.

![example of a broken folders tree](https://user-images.githubusercontent.com/2471284/94853317-8405a780-042b-11eb-8f4a-ecb1008c9189.png)

alecpl:
Fixed.

PR #7648, opened 2020-10-01
Fix incorrect folders settings 

abshoff:
Fix for #7647

If a special folder was located within some subfolder, then the folders settings view was broken. It added a virtual folder for an existing folder which itself was not displayed. This bugfix puts special folders at the root level and removes any parent folder.

before:

example of a broken folders tree

after:

example of a fixed folders tree

### DIFF
--- a/program/steps/settings/folders.inc
+++ b/program/steps/settings/folders.inc
@@ -292,21 +292,21 @@ function rcmail_folder_subscriptions($attrib)
             $noselect = in_array_nocase('\\Noselect', $attrs);
         }
 
-        $disabled = (($is_protected && $is_subscribed) || $noselect);
+        $is_disabled = (($is_protected && $is_subscribed) || $noselect);
 
         // Below we will disable subscription option for "virtual" folders
         // according to namespaces, but only if they aren't already subscribed.
         // User should be able to unsubscribe from the folder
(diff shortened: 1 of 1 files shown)


alecpl:
What happens when such a special folder has sub-folders?

abshoff:
Those are displayed as subfolders of the special folder, in the original version as well as the proposed fix.

On a mobile phone, for example, it looks like that.

88507EF5-8DB8-456D-92AB-5E244717E5E1

program/steps/settings/folders.inc
```diff
@@ -224,8 +224,9 @@ function rcmail_folder_subscriptions($attrib)
         $folder        = $STORAGE->mod_folder($folder);
         $foldersplit   = explode($delimiter, $folder);
         $name          = rcube_charset::convert(array_pop($foldersplit), 'UTF7-IMAP');
-        $parent_folder = join($delimiter, $foldersplit);
-        $level         = count($foldersplit);
+        $special       = $folder_id == 'INBOX' || isset($special_folders[$folder_id]);
```
alecpl, line 227: $special_folders already contains INBOX. Also, better name would be $is_special.

abshoff: I agree with that and have changed some related variable names.

PR #7661, opened 2020-10-07
implement trim() to password input field

redoo-networks:
We had a long discussion within our team, if we should trim a password and remove leading and trailing spaces within an existing environment.

We come to the conclusion, that it will provide more stability and default users do not use leading and trailing spaces.
We save more time by get rid of Copy & Paste errors in support requests, then it will cost one time to help users fix their password.

Because we want to push this into main repo, we implement a configuration option to enable this. 
Per default this feature is disabled and not used. 
We also implement this into the "passwords" plugin to transparent remove leading/trailing spaces.

Because this is our first contribution to roundcube, please let me know when some formatting is not correct or the feature is not wanted.

Regards,
Stefan

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -1409,3 +1409,7 @@ $config['message_show_email'] = false;
 // 0 - Reply-All always
 // 1 - Reply-List if mailing list is detected
 $config['reply_all_mode'] = 0;
+
+// Enables the trim of leading and trailing spaces in passwords during login and password plugin
+// If enabled and existing users did use spaces in front or end of passwords, they cannot login
+$config['password_trim'] = false;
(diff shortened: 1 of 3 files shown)


alecpl:
I don't like an extra option for this. I guess, a more appropriate way would be:
1. Always trim passwords in password plugin, or forbid entering them with an error to the user.
2. When logging in:
- Try to log in with the supplied password
- If it fails - check if the typed password had any spaces before or after
- Try again with leading and trailing spaces stripped away

philliptaylorpro:
I don't really like this idea. You may be the only company in the world to trim passwords like this that I'm aware, so it seems like perhaps this is one or two people using bad practice and lazy copy/pasting than actually something roundcube should accommodate. Especially since no one else does. Given the password goes upstream to the mailserver I think this is asking for trouble.

PR #7686, opened 2020-10-19
change seperator of pagetitle

vaelu:
since we have this in our version since some time it would be cool if this could be integrated


alecpl:
What's wrong with this separator? I'll definitely not add another config option for such a thing.

vaelu:
okay, just want to change this and thought that it's changeable in the skin but as it wasn't I added this

PR #7688, opened 2020-10-20
Steps -> Actions refactoring

alecpl:
@johndoh @thomascube This is not finished, but I wanted you to know what I'm up to.

The reasons for this refactoring are:
- I always wanted to do this, but I was always afraid of it (possible bugs, hard backporting of fixes)
- a lot of planned changes in the code regarding PHP8 strictness make this a good time for refactoring
- cleaner code, better testability (I plan some improvements on this front, and more tests which is kind of required in PHP8 era of our code), no globals, easier code shareability (that's why all methods are static).

ps. this is not yet addressing PHP8.


alecpl:
I'm going to merge this tomorrow and then work on more tests and PHP8 compatibility.

thomascube:
I'm a bit late to comment but the changes look good! However, if not too late, I'd suggest to branch off for release 1.5. I think we have a good load of changes and new features for 1.5 and a major refactoring like this should go into a new milestone.

alecpl:
I don't really see a reason to branch off. We do what we usually do, i.e. we do not branch off before rc. And we're not even ready for the beta release. There's still a lot to do to make the code working with PHP8 with no warnings. I set the beta release date to myself around the New Year.

trybowski:
Due to this change the OAuth2 seems to be broken. index.php still tries to include file from /steps/.

alecpl:
@trybowski Fixed in b3677c3e0be, but I didn't tested if OAuth is working.

PR #7690, opened 2020-10-21
Do not remove commas when importing a groups/categories list

tpayen:
I probably don't getting the point, but why removing commas in this case if it's the right separator to use for a groups/categories list ? 
By doing this, importing a well separated groups/categories list never work

### DIFF
--- a/program/lib/Roundcube/rcube_csv2vcard.php
+++ b/program/lib/Roundcube/rcube_csv2vcard.php
@@ -635,7 +635,6 @@ class rcube_csv2vcard
 
         if (!empty($contact['groups'])) {
             // categories/groups separator in vCard is ',' not ';'
-            $contact['groups'] = str_replace(',', '', $contact['groups']);
             $contact['groups'] = str_replace(';', ',', $contact['groups']);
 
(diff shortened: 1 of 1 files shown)


johndoh:
That change came from this issue #4710. I believe the intention was to remove the reserved comma character from group names. Ie. other systems might export a group name like "Friends, family and neighbors" which is a single group but Roundcube will detect the comma as a separator.

As you say a CSV file can also be generated using comma as a separator so for me its a 50/50 thing.

You might have seen in git-master there is a intermediate step for CSV imports now to allow manual mapping of the fields. For me adding another option in there for group separator would be over the top but it could be done.

tpayen:
Thanks John for your answer.

I didn't see the issue #4710, it's a defendable idea, and like you said its a 50/50 thing.

I saw your commit for CSV imports manual mapping (this is great) and I agree this would be over the top for what concerns us here (also I'm not sure "average" users will understand this point).

I see two solutions then, either to make it configurable for example by letting the possibility of configuring a other character as separator or I change my csv exporter to use "," instead of ";" as group separator

PR #7703, opened 2020-10-30
When creating or forwarding email, entering email address in "to" field causes "internal server error" popup

dmmiller2:
Fixes syntax error when trying to fetch names from "Collected Recipients" and "Trusted Senders" when trying to address new or forwarded email. Syntax error noted in debug logging output. Extraneous comma removed.


alecpl:
This has been fixed in the recent code refactorings.

PR #7714, opened 2020-11-06
Allow multiple messages to be bounced/resent at once

endnil:
This PR extends the bounce interface to allow multiple messages to be selected and bounced to the same recipient.

It also sneaks in a change to remember (in a user pref) the folder the bounced messages were last saved in.  (I'm happy to separate this out if needed, though.)

Please let me know if any other changes are needed.  Thanks for all your work on Roundcube.

### DIFF
--- a/program/actions/mail/bounce.php
+++ b/program/actions/mail/bounce.php
@@ -28,29 +28,33 @@ class rcmail_action_mail_bounce extends rcmail_action
     public function run($args = [])
     {
         $rcmail     = rcmail::get_instance();
-        $msg_uid    = rcube_utils::get_input_value('_uid', rcube_utils::INPUT_GP);
+        $msg_uids   = rcube_utils::get_input_value('_uids', rcube_utils::INPUT_GP);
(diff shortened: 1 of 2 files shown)


alecpl:
I don't like the idea because it has performance implications, and it probably does not work when you select all messages (using "select all").

endnil:
No worries, thank you for the feedback @alecpl.

Indeed, I see now that it doesn't work with the select all function.

Are the performance implications to do with the increase in response time when sending and saving multiple messages?

I'm happy to spend some more time on this if it will be useful.

alecpl:
> Are the performance implications to do with the increase in response time when sending and saving multiple messages?
This too but imagine a user bouncing a whole folder with thousends of messages, or even less messages but with big size.

Because of that zipdownload plugin has a limit, but doing the same for bounce might not be as reliable.

endnil:
Thanks for explaining. I can see how while this works well for our niche use-case, it's not generally suitable.

Issue #7724, opened 2020-11-14
Password: Support for ARGON2 family of hashes


fluteze:
Please consider adding this feature to the password changing plugin:
```
--- password.php.orig	2020-11-14 12:26:00.748848486 -0700
+++ password.php	2020-11-14 12:24:07.090332419 -0700
@@ -535,6 +535,16 @@
         }
 
         switch ($method) {
+        case 'argon2i':
+            $crypted = password_hash($password, PASSWORD_ARGON2I);
+            $prefix  = '{ARGON2I}';
+            break;
+
+        case 'argon2id':
+            $crypted = password_hash($password, PASSWORD_ARGON2ID);
+            $prefix  = '{ARGON2ID}';
+            break;
+
```

and the configuration file comment section:
```
--- password.inc.php.orig	2020-11-14 12:48:59.621988887 -0700
+++ password.inc.php	2020-11-14 12:49:44.521798416 -0700
@@ -43,7 +43,7 @@
 $config['password_force_new_user'] = true;
 
 // Default password hashing/crypting algorithm.
-// Possible options: des-crypt, ext-des-crypt, md5-crypt, blowfish-crypt,
+// Possible options: argon2i, argon2id, des-crypt, ext-des-crypt, md5-crypt, blowfish-crypt,
 // sha256-crypt, sha512-crypt, md5, sha, smd5, ssha, ssha512, samba, ad, dovecot, clear.
 // For details see password::hash_password() method.
 $config['password_algorithm'] = 'argon2id';
```
This in use in my installation, though someone might like it.

Not sure why the first patch above looks funny and the second is ok.

johndoh:
Its better to submit PRs than try and past patches into tickets.

Also it should probably be in the config notes and may be even checked in the code that PASSWORD_ARGON2I requires PHP >= 7.2 and PASSWORD_ARGON2ID >= 7.3 since Roundcube supports >=5.5.

fluteze:
I like the idea of flexibility. Yes, there does need to be some checking done for PHP versions. I'll work on a patch and submit the relevant PR.

robo2bobo:
Please be aware, that RHEL and CentOS (R.I.P) do not support argon under PHP at all, even in PHP 7.x that comes with RHEL 8 and CentOS 8.

[...]

PR #7725, opened 2020-11-15
add support for php's password_hash() function

### LINKED ISSUE #7723: Password change plugin adding empty passwords
The password change plugin returns empty an password field if the "$config['password_algorithm']"  contains an invalid entry.
This is a result of the switch/case statement having no default action. Here is a patch to fix this issue:
```
`--- password.php.orig	2020-11-14 12:26:00.748848486 -0700
+++ password.php	2020-11-14 12:24:07.090332419 -0700
@@ -768,6 +778,11 @@
         case 'clear':
             $crypted = $password;
 	    break;
+
+	default:
+            return false;
+            break;
+
`

johndoh:
ref #7724

adds support for PHP's `password_hash()` function to the password plugin. I also put the hashing options in alphanumeric order for easier management.

also fixes #7723

### DIFF
--- a/plugins/password/config.inc.php.dist
+++ b/plugins/password/config.inc.php.dist
@@ -44,8 +44,11 @@ $config['password_force_new_user'] = false;
 
 // Default password hashing/crypting algorithm.
 // Possible options: des-crypt, ext-des-crypt, md5-crypt, blowfish-crypt,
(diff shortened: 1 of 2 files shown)


alecpl:
I don't like the use of PHP constants in config and separate option for password-hash algorithm(s). I would rather just add `argon2i` and `argon2id` options to `password_algorithm`. We can have `password_algorithm_options` instead of `password_php_hash_options`.

PR #7733, opened 2020-11-20
Phpdoc type annotations

mstilkerich:


program/lib/Roundcube/rcube_plugin.php
```diff
@@ -256,7 +256,7 @@ public function add_label()
     /**
      * Wrapper for rcube::gettext() adding the plugin ID as domain
      *
-     * @param string $p Message identifier
+     * @param mixed $p Named parameters array or label name
```
alecpl, line 259: `string|array` would be better.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -397,7 +397,7 @@ function delete_group($group_id)
      * @param string $newname  New name to set for this group
      * @param string &$newid   New group identifier (if changed, otherwise don't set)
      *
-     * @return boolean New name on success, false if no data was changed
+     * @return boolean|string New name on success, false if no data was changed
```
alecpl, line 400: This should be `string|false`.

alecpl (changes requested):

mstilkerich:
Thanks for the feedback, I adapted as suggested

PR #7735, opened 2020-11-21
Add error handling for unsupported hash method

fluteze:
When using MySQL database for passwords, an unsupported or typo'ed hash method name was causing user passwords to be deleted with only the contents of the $PREFIX to be stored. i.e. password field in the database would have '{ARGON2ID}' but no following text to contain the password.


plugins/password/localization/en_US.inc
```diff
@@ -25,6 +25,7 @@ $messages['nopassword'] = 'Please input new password.';
 $messages['nocurpassword'] = 'Please input current password.';
 $messages['passwordincorrect'] = 'Current password incorrect.';
 $messages['passwordinconsistency'] = 'Passwords do not match, please try again.';
+$messages['passwdhashnotsupported'] = 'Password hash function unknown or unsupported. Possible typo in hash name.');
```
johndoh, line 28: There is an extra ')' on this line

fluteze: Oops. Removed the extra )

fluteze: This is my first time submitting a PR update via github. Do I need to do something to have the update propagate?

plugins/password/password.php
```diff
@@ -769,6 +773,14 @@ static function hash_password($password, $method = '', $prefixed = true)
 
         case 'clear':
             $crypted = $password;
+            break;
+        default:
+            rcube::raise_error(array(
+                'code' => 600, 'file' => __FILE__, 'line' => __LINE__,
+                'message' => "Password plugin: Hash method not supported."
+                ), true, true);
```
alecpl, line 781: The way you call raise_error() make it stop script execution. And this is good. We don't want to display the error to the end user. Please, remove use of PASSWORD_HASH_NOT_SUPPORTED and the following return statement.

fluteze: Done.

PR #7739, opened 2020-11-23
update deps

johndoh:
This will need another review before merge as some libs have a frequent release schedule. My idea is this could be merged just before the 1.5-beta release.

I have checked for compatibility issues.

A quick note about PHPUnit: version 8 makes some breaking changes and requires PHP >= 7.2, Return type declarations need to be added to some functions, see https://phpunit.de/announcements/phpunit-8.html

Less.js: held at version 3.9 because of speed issue ref https://github.com/less/less.js/issues/3434

### DIFF
--- a/jsdeps.json
+++ b/jsdeps.json
@@ -61,21 +61,21 @@
     {
       "lib": "openpgp",
       "name": "OpenPGP.js",
-      "version": "4.4.6",
+      "version": "4.10.9",
       "url": "https://raw.githubusercontent.com/openpgpjs/openpgpjs/v$v/dist/openpgp.min.js",
       "api_url": "https://api.github.com/repos/openpgpjs/openpgpjs/contents/dist/openpgp.min.js?ref=v$v",
       "dest": "plugins/enigma/openpgp.min.js",
-      "sha1": "e142168db8e666a40fcb4c48ef89c9d774134f8b",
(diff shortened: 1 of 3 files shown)


jsdeps.json
```diff
@@ -101,11 +101,11 @@
     {
       "lib": "less",
       "name": "LessJS",
-      "version": "2.7.3",
+      "version": "3.13.0",
```
alecpl, line 104: From my experience less v3 is much slower than v2, but we'll have to check again.

alecpl:
PHPUnit v7 works with PHP8 for me, so we don't have to bump this one.

jsdeps.json
johndoh: Confirmed it is a lot slower. Tests on the mail screen show around 609ms for v2 and 1056ms for v3. Its only used for dev so may be there is no need to update at this time.

johndoh: The Less.JS speed issues starts with version 3.10 ref https://github.com/less/less.js/issues/3434 so update to 3.9.0 should be fine, 3.9 is very slightly faster than 2.7 for me.

alecpl:
Less.js 3.13.0 fixes some perf issues, could you test it?

johndoh:
Yep 3.13.0 is good, updated the PR.

PR #7745, opened 2020-11-25
Update labels.inc

j-herden:
small typo corrected


alecpl:
We manage localization via https://www.transifex.com/roundcube/roundcube-webmail. I see it's fixed there. Please, don't waste our time.

PR #7751, opened 2020-11-27
Fix array access before checking for array

### LINKED ISSUE #7750: rcmail_output_html: Warning: Illegal string offset 'extwin'
If you use the `rcmail_output_html::redirect` function in a plugin to redirect the user to a different url, it throws an error if `$p` is given as `string` and `$this->env['extwin']` is set:

```
    /**
     * Redirect to a certain url
     *
     * @param mixed $p      Either a string with the action or url parameters as key-value pairs
     * @param int   $delay  Delay in seconds
     * @param bool  $secure Redirect to secure location (see rcmail::url())
     */
    public function redirect($p = array(), $delay = 1, $secure = false)
    {
        if ($this->env['extwin'])
            $p['extwin'] = 1;
        $location = $this->app->url($p, false, false, $secure);
        header('Location: ' . $location);
        exit;
    }
```

This is because if `$this->env['extwin']` is set, `$p` will be used as an array (which it isn't). I think there should be a check like `is_array` before accessing `$p`

codegain:
Fixes #7750


alecpl:
It should rather do `if (!is_array($p)) { $p = ['_action' => $p]; }`

codegain:
@alecpl I use this function in a plugin to redirect to a different website. Is this still possible if the URL is passed into `_action`?

alecpl:
Ah, despite the argument description url() method actually accepts urls:
```
    if (!is_array($p)) {
        if (strpos($p, 'http') === 0) {
            return $p;
        }
```
So, I guess the fix would need to be extended to support this case.

ps. that prefix check is really naive and should be improved.

alecpl:
Fixed with #7823

PR #7770, opened 2020-12-07
Rename composer.json-dist to composer.json

mvorisek:
Allow to require this project by composer for static analysis for plugin development.


mvorisek:
requires too much additional changes

Issue #7777, opened 2020-12-13
(Minor) Return 204 rather than 404 on missing contact photo


johndoh:
When requesting a contact photo for a contact which has no photo stored Roundcube returns either the blank.gif or a `404 Photo not found` error depending on the `_error` param in the URL. In some cases (eg. Elastic skin) `onerror` handling on the `<img>` tag is used to display a placeholder image in the event no contact photo exists.

According to the HTML spec error handling for `<img>`'s is dependent on the content of the server response and not the HTTP status code:

From http://www.whatwg.org/specs/web-apps/current-work/#the-img-element
> Whether the image is fetched successfully or not (e.g. whether the response code was a 2xx code or equivalent) must be ignored when determining the image's type and whether it is a valid image.
>
> Note: This allows servers to return images with error responses, and have them displayed.

Based on this I think a status code of 204 (No Content) could be used instead of 404 or even 200. I think this would eliminate some noise from the results of log analysis where sysadmins may monitor 404 response to detect application/configuration errors or even attacks.

I've tested this in IE11, Chrome and Firefox and the contact and placeholder images are showing correctly.

alecpl:
Makes sense. Create a PR, there's one assertion in `tests/Actions/Contacts/Photo.php` that will have to change.

thomascube:
The magic number 7777 :-)

Thanks a lot for the input and PR!

johndoh:
> The magic number 7777 :-)

Hehe yep :-)

I thought I'd look into if there was any realistc alternative to the 404 response after this [forum post](https://www.roundcubeforum.net/index.php/topic,29200.msg74441.html?PHPSESSID=buodl1qk72576gq81dins5kg29#msg74441).

alecpl:
Done.

Issue #7765, opened 2020-12-04
Photo not found Issue


ADDISON74:
I am using RoundCube 1.4.8 (Elastic theme) with a virtual machine (Debian + Virtualmin). In order to reproduce this issue please follow these steps:

1. Get into RoundCube and compose a couple of short emails, just text without any images. For receiver's email address use the same email address used for accessing RoundCube.

2. Send the emails. Open Inspector into your browser and click on Console tab. Once you receive the messages click on them and watch the Console.

Here is the error messages I got for every click:

```
GET https://www.mydomain.tld/roundcube/?_task=addressbook&_action=photo&_email=contact%40mydomain.tld&_error=1 404 (Photo not found)
```

By deleting the part "_action=photo" from the link and loaded in a separate tab I was able to load the Address Book page without any erros in Console.

Now go to RoundCube Settings and switch from Elastic to Larry theme. Click again on the messages in Inbox and watch the Console. This time there are no errors.

P.S. I accidentally discovered this issue. I have been using the Larry theme for a long time.

alecpl:
Not a bug.

ADDISON74:
Just "not a bug"? Without any feedback? 

I'm sorry I wasted 30 minutes of my time investigating and writing about this issue.

alecpl:
The fact that the request returns 404 does not mean it is an error. In this case it means "there's no contact with a photo and specified email address".

ADDISON74:
I appreciate for the feedback. 

How do you explain this issue appears only in Elastic theme but not in Larry? A small change is needed even it is inoffensive one. I never used in 5 years contacts having photos associated. Even I like the new theme I will continue using Larry. I don't like seeing errors.

PR #7778, opened 2020-12-13
return 204 rather than 404 on missing contact photo

johndoh:
for #7777


johndoh:
I've added a small change in `rcmail_action_mail_show::message_contactphoto()` to remove the hard coded image path.

alecpl:
Quite a lot of plugins use the blank.gif image. I suggest to keep it, even if it's not used by the core.

johndoh:
Done. I thought it might be confusing to have it there when the core uses the constant set in rcmail_output rather than this file.

ADDISON74:
It seems that you must be contributor to this project in order for a reported issue to be considered. The same issue was reported by me here #7765 and it was closed with "not a bug" phrase.

Perhaps a peaceful approach on the part of some may be needed in the future.

PR #7781, opened 2020-12-14
cleanup: Remove PHP warnings that could appear in error.log on failed authentication

jpbion:
Self-explanatory. With PHP 7+, you can use the null coalescing operator '??' to avoid an E_NOTICE error on an undefined value. But since roundcubemail wants to be compatible with PHP back to 5.5, you need to use an isset to avoid the E_NOTICE. These changes to index.php do this at the top level file only, around a failed user authentication.

### DIFF
--- a/index.php
+++ b/index.php
@@ -157,7 +157,7 @@ if ($RCMAIL->task == 'login' && $RCMAIL->action == 'login') {
             $error_code = rcmail::ERROR_INVALID_REQUEST;
         }
         else {
-            $error_code = is_numeric($auth['error']) ? $auth['error'] : $RCMAIL->login_error();
+            $error_code = isset($auth['error']) && is_numeric($auth['error']) ? $auth['error'] : $RCMAIL->login_error();
         }
 
         $error_labels = [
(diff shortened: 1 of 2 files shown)


index.php
```diff
@@ -168,7 +168,7 @@
             rcmail::ERROR_RATE_LIMIT       => 'accountlocked',
         ];
 
-        $error_message = !empty($auth['error']) && !is_numeric($auth['error']) ? $auth['error'] : ($error_labels[$error_code] ?: 'loginfailed');
+        $error_message = !empty($auth['error']) && !is_numeric($auth['error']) ? $auth['error'] : (isset($error_labels[$error_code]) && $error_labels[$error_code] ?: 'loginfailed');
```
alecpl, line 171: This is wrong. BTW, this could be fixed also by making sure the 'error' key always exists.

jpbion: I’m really not seeing where it is wrong. It simply does a check with isset - which as the first part of the && expression, if false it makes the whole expression false without PHP even trying to evaluate the other parts of the && expression. I may be missing something blindingly obvious, of course. I agree it’s better if the error key exists! :)

alecpl: It returns boolean instead of a string.

alecpl:
BTW, we require an error_reporting setting that excludes E_NOTICE. This case will be a problem in PHP8 so we'll fix it anyway, but support for PHP8 is in progress.

alecpl:
dcfd12e98aa5 should fix this.

PR #7785, opened 2020-12-15
deluser.sh should exit with success

tomsommer:


bin/deluser.sh
```diff
@@ -142,3 +142,5 @@ else {
         echo "Successfully deleted user $user->ID\n";
     }
 }
+
+exit(0);
```
alecpl, line 146: `exit(0)` at the end is not needed. I believe.

tomsommer:
Correct. Just clearer intent. 

--
Tom Sommer

alecpl:
We do not do this in any other script, so I propose to remove this part.

PR #7791, opened 2020-12-19
fix a few PHP notices

johndoh:
fix a few PHP notices when all logging is used on PHP7 - I guess these are really PHP8 fixes but I have not had a chance to test there yet.

### DIFF
--- a/program/lib/Roundcube/rcube_imap_cache.php
+++ b/program/lib/Roundcube/rcube_imap_cache.php
@@ -1092,7 +1092,7 @@ class rcube_imap_cache
                 $mbox_data = $this->imap->folder_data($mailbox);
 
                 // Removed messages found
-                $uids = rcube_imap_generic::uncompressMessageSet($mbox_data['VANISHED']);
+                $uids = isset($mbox_data['VANISHED']) ? rcube_imap_generic::uncompressMessageSet($mbox_data['VANISHED']) : null;
                 if (!empty($uids)) {
                     $removed = array_merge($removed, $uids);
                     // Invalidate index
@@ -1189,7 +1189,7 @@ class rcube_imap_cache
         if (isset($msg->body)) {
             $length = strlen($msg->body);
 
-            if ($msg->body_modified || $size + $length > $this->threshold * 1024) {
+            if (!empty($msg->body_modified) || $size + $length > $this->threshold * 1024) {
                 unset($msg->body);
             }
             else {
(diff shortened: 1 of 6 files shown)


program/lib/Roundcube/rcube_imap_cache.php
```diff
@@ -216,7 +216,7 @@ function get_index($mailbox, $sort_field = null, $sort_order = null, $existing =
                 $sort_field = $index['sort_field'];
             }
 
-            if ($sort_field != $index['sort_field']) {
+            if (isset($index['sort_field']) && $index['sort_field'] != $sort_field) {
```
alecpl, line 219: I don't think this is the same. If `$index['sort_field']` isn't set it still can be different than `$sort_field`.

program/lib/Roundcube/rcube_imap_cache.php
```diff
@@ -855,7 +855,7 @@ private function validate($mailbox, $index, &$exists = true)
         // and many rcube_imap changes to connect when needed
 
         // Check UIDVALIDITY
-        if ($index['validity'] != $mbox_data['UIDVALIDITY']) {
+        if (isset($index['validity']) && $index['validity'] != $mbox_data['UIDVALIDITY']) {
```
alecpl, line 858: Not exactly the same.

program/lib/Roundcube/rcube_message.php
```diff
@@ -880,19 +880,19 @@ private function parse_structure($structure, $recursive = false)
                         continue;
                     }
 
-                    if ($mail_part->headers['content-id']) {
+                    if (isset($mail_part->headers['content-id'])) {
```
alecpl, line 883: Better `!empty()` here.

program/lib/Roundcube/rcube_message.php
```diff
@@ -880,19 +880,19 @@ private function parse_structure($structure, $recursive = false)
...
                         $mail_part->content_id = preg_replace(['/^</', '/>$/'], '', $mail_part->headers['content-id']);
                     }
 
-                    if ($mail_part->headers['content-location']) {
+                    if (isset($mail_part->headers['content-location'])) {
```
alecpl, line 887: Better `!empty()` here.

program/lib/Roundcube/rcube_message.php
```diff
@@ -938,7 +938,7 @@ private function parse_structure($structure, $recursive = false)
                     if (isset($inline_object->content_id)) {
                         $a_replaces['cid:'.$inline_object->content_id] = $part_url;
                     }
-                    if ($inline_object->content_location) {
+                    if (isset($inline_object->content_location)) {
```
alecpl, line 941: `!empty()`

alecpl (changes requested):

PR #7798, opened 2020-12-22
Remove invalid search mods from user preferences

### LINKED ISSUE #7789: "UID SEARCH: Unknown argument" on invalid search_mods
Hi,

I got a few users complaining that their search bar is not working anymore (On roundcube 1.4.9). After some digging I found the following:
If there are invalid `search_mods` saved for the user, the `UID SEARCH` command fails.

I looked at one user's saved preferences and found the following (dump of `user->get_prefs()`)

```
Array
(
    [search_mods] => Array
        (
            [*] => Array
                (
                    [Sent Messages] => 1
                )
        [...]
```

This results in the following `UID SEARCH` error when searching:

```
A0005 UID SEARCH HEADER SENT MESSAGES test
A0005 BAD Error in IMAP command UID SEARCH: Unknown argument TEST (0.001 + 0.000 secs).
```

I don't know how `Sent Messages` did end up in the `search_mods` but it's impossible for the user to remove it on their own. After changing the search filter this simply results in something like this:

```
A0005 UID SEARCH OR HEADER SENT MESSAGES test HEADER TO test
A0005 BAD Error in IMAP command UID SEARCH: Unknown argument TEST (0.001 + 0.000 secs).
```

Would it be possible to check the `search_mods` array before doing the search and simply remove invalid values? Or maybe after the user logs in: Clean the `search_mods` of invalid values and save it?

codegain:
Fixes #7789

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -268,6 +268,22 @@ class rcmail_action_mail_index extends rcmail_action
             }
         }
 
+        // Validate the search mods
+        $invalid_search_mod_found = false;
+
+        foreach ($mods as $mbox => $mbox_mods) {
+            foreach ($mbox_mods as $mod => $value) {
+                if (!rcube_utils::check_search_mod($mod)) {
+                    $invalid_search_mod_found = true;
+                    unset($mods[$mbox][$mod]);
+                }
+            }
+        }
+
+        if ($invalid_search_mod_found) {
+            $rcmail->user->save_prefs(array('search_mods' => $mods));
+        }
+
         return $mods;
     }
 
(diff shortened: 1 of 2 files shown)


codegain:
@alecpl The regex I use is based on RFC 2882 (https://tools.ietf.org/html/rfc2822#section-3.6.8) and RFC 5322 (https://tools.ietf.org/html/rfc5322#section-3.6.8). It tought adding this to "rcube_utils" would be better than just using the regex directly in the `search_mods()` function.

alecpl:
I've decided for a different approach in 837078cd23d46.

PR #7800, opened 2020-12-27
get minifying tools from npm

johndoh:
https://github.com/roundcube/roundcubemail/issues/7567#issuecomment-678982082 suggested perhaps using `npm` to download the google closure compiler. The `yuicompressor` util used for CSS minifying is also available on `npm`. Since `npm` is already suggested for installing `less` for the Elastic skin I don't think this adds any new dependency and it means you don't need Java on the build machine any more. Also I think using `npm` probably gives more security than the unverified downloads currently used.

### DIFF
--- a/bin/cssshrink.sh
+++ b/bin/cssshrink.sh
@@ -3,37 +3,17 @@
 set -e
 
 PWD=`dirname "$0"`
-JAR_DIR='/tmp'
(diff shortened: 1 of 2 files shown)


alecpl:
I think you still need java, but it is better anyway.

PR #7807, opened 2020-12-31
add placeholder text when no responses on compose screen

johndoh:
I think the Responses menu on the compose screen looks a little strange when the use has no responses configured:

before

this PR adds a bit of placeholder text for this case:

after

If the user uses the inline Create function to make a new response the placeholder text is removed.


alecpl:
Bug: When I have a response defined, immediately after accessing the mail compose page "Edit responses" item is inactive. It does not activate even after I create another response.

johndoh:
I cannot recreate the bug you mentioned. app.js lines 365 and 5123 should take care of enabling the edit option and they do for me.. Not enabling after you added a new response inline is really strange as that is hard coded to enabled :/ I'm not sure what could cause it. I'll keep trying to recreate.

I've added `aria-disabled="true"`

alecpl:
False alarm. Now it works for me. Sorry for the noise.

program/js/app.js
```diff
@@ -3922,7 +3926,7 @@ function rcube_webmail()
         $('#' + ref.env.composebody).hide();
 
         // disable commands that operate on the compose body
-        ref.enable_command('spellcheck', 'insert-sig', 'toggle-editor', 'insert-response', 'save-response', false);
+        ref.enable_command('spellcheck', 'insert-sig', 'toggle-editor', 'insert-response', 'save-response', 'edit-responses', false);
```
alecpl, line 3929: I think this is a mistake. edit-responses does not operate on compose body and I see no reason to disable it in mailvelope mode.

program/js/app.js
```diff
@@ -5134,12 +5139,17 @@ function rcube_webmail()
             return rcube_event.cancel(e);
           }
         });
+
+      // remove the placeholder item if its there
+      $('a.insertresponse.placeholder').remove();
```
alecpl, line 5144: I'd feel better if you narrow the search to the responseslist object.

alecpl (changes requested):

program/js/app.js
johndoh: save-response is also disabled which does not operate on compose body, I thought disabling edit was consistent

program/js/app.js
johndoh: done

program/js/app.js
johndoh: in mailvelope mode you cannot insert responses or create new ones, having edit enabled seems kind of pointless. perhaps it would be better to disable the whole responses menu in this case.

alecpl: edit-responses is an "external link" so it should be just enabled. Disabling it would be like disabling edit-identities command.

alecpl (changes requested):

program/js/app.js
johndoh: do you mean you want the edit link always enabled no matter what? because its "edit" I thought it should only be enabled if there is a response to edit. If so I can revert that part of the PR back to current behaviour.

PR #7809, opened 2021-01-02
Improve preciseness of some PHPdoc annotations

mstilkerich:
Another few improvements to the type annotations to improve static analysis results.

### DIFF
--- a/program/lib/Roundcube/rcube_addressbook.php
+++ b/program/lib/Roundcube/rcube_addressbook.php
@@ -123,9 +123,9 @@ abstract class rcube_addressbook
      *
      * @param string|string[] $fields   List of fields to search in
      * @param string|string[] $value    Search value
-     * @param int    $mode    Search mode. Sum of self::SEARCH_*.
-     * @param bool   $select  True if results are requested, False if count only
-     * @param bool   $nocount True to skip the count query (select only)
+     * @param int             $mode     Search mode. Sum of self::SEARCH_*.
+     * @param bool            $select   True if results are requested, False if count only
+     * @param bool            $nocount  True to skip the count query (select only)
      * @param string|string[] $required List of fields that cannot be empty
      *
      * @return object rcube_result_set List of contact records and 'count' value
--- a/program/lib/Roundcube/rcube_addressbook.php
+++ b/program/lib/Roundcube/rcube_addressbook.php
@@ -51,52 +51,52 @@ abstract class rcube_addressbook
     /** @var string Name of the primary key field of this addressbook. Used to search for previously retrieved IDs. */
     public $primary_key;
 
-    /** @var bool   True if the addressbook supports contact groups. */
-    public $groups        = false;
+    /** @var bool True if the addressbook supports contact groups. */
+    public $groups = false;
 
     /**
-     * @var bool    True if the addressbook supports exporting contact groups. Requires the implementation of
+     * @var bool True if the addressbook supports exporting contact groups. Requires the implementation of
      *              get_record_groups().
      */
     public $export_groups = true;
 
-    /** @var bool   True if the addressbook is read-only. */
-    public $readonly      = true;
+    /** @var bool True if the addressbook is read-only. */
+    public $readonly = true;
 
     /**
-     * @var bool    True if the addressbook does not support listing all records but needs use of the search function.
+     * @var bool True if the addressbook does not support listing all records but needs use of the search function.
      */
-    public $searchonly    = false;
+    public $searchonly = false;
 
-    /** @var bool   True if the addressbook supports restoring deleted contacts. */
-    public $undelete      = false;
+    /** @var bool True if the addressbook supports restoring deleted contacts. */
+    public $undelete = false;
 
-    /** @var bool   True if the addressbook is ready to be used. See rcmail_action_contacts_index::$CONTACT_COLTYPES */
-    public $ready         = false;
+    /** @var bool True if the addressbook is ready to be used. See rcmail_action_contacts_index::$CONTACT_COLTYPES */
+    public $ready = false;
 
     /**
      * @var null|string|int If set, addressbook-specific identifier of the selected group. All contact listing and
      *                      contact searches will be limited to contacts that belong to this group.
      */
-    public $group_id      = null;
+    public $group_id = null;
 
-    /** @var int    The current page of the listing. Numbering starts at 1. */
-    public $list_page     = 1;
+    /** @var int The current page of the listing. Numbering starts at 1. */
+    public $list_page = 1;
 
-    /** @var int    The maximum number of records shown on a page. */
-    public $page_size     = 10;
+    /** @var int The maximum number of records shown on a page. */
+    public $page_size = 10;
 
     /** @var string Contact field by which to order listed records. */
-    public $sort_col      = 'name';
+    public $sort_col = 'name';
 
     /** @var string Whether sorting of records by $sort_col is done in ascending (ASC) or descending (DESC) order. */
-    public $sort_order    = 'ASC';
+    public $sort_order = 'ASC';
 
     /** @var string[] A list of record fields that contain dates. */
-    public $date_cols     = [];
+    public $date_cols = [];
 
     /** @var array Definition of the contact fields supported by the addressbook. */
-    public $coltypes      = [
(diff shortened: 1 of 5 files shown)


program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -121,12 +121,12 @@ function refresh_search()
...
      *
-     * @param array  $fields   List of fields to search in
-     * @param string $value    Search value
-     * @param int    $mode     Search mode. Sum of self::SEARCH_*.
-     * @param bool   $select   True if results are requested, False if count only
-     * @param bool   $nocount  True to skip the count query (select only)
-     * @param array  $required List of fields that cannot be empty
+     * @param string|string[] $fields   List of fields to search in
+     * @param string|string[] $value    Search value
+     * @param int    $mode    Search mode. Sum of self::SEARCH_*.
```
alecpl, line 126: Please, params descriptions should be aligned with each other.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -121,12 +121,12 @@ function refresh_search()
...
+     * @param bool   $select  True if results are requested, False if count only
+     * @param bool   $nocount True to skip the count query (select only)
+     * @param string|string[] $required List of fields that cannot be empty
```
alecpl, line 129: The code indeed accepts a string, but I would not advertise this possibility here.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -121,12 +121,12 @@ function refresh_search()
     /**
      * Search records
      *
-     * @param array  $fields   List of fields to search in
-     * @param string $value    Search value
-     * @param int    $mode     Search mode. Sum of self::SEARCH_*.
-     * @param bool   $select   True if results are requested, False if count only
-     * @param bool   $nocount  True to skip the count query (select only)
-     * @param array  $required List of fields that cannot be empty
+     * @param string|string[] $fields   List of fields to search in
```
alecpl, line 124: Please, take a look at rcube_contacts param descriptions. We should use similar here. I mean the description should say something about the type difference too.

alecpl (changes requested):

program/lib/Roundcube/rcube_addressbook.php
mstilkerich: Fixed

program/lib/Roundcube/rcube_addressbook.php
mstilkerich: From my viewpoint as developer of an implementation of this interface, I'm looking at this not as advertisement of possibilities but as required expected behavior that I have to implement. So unless roundcube guarantees not to use this interface with string parameters, it must be listed here.
Looking at the roundcube code, it seems to _exclusively_ use this parameter in its string form, I did not find a single use of the array variant:

```
./program/actions/mail/search_contacts.php:            $result = $source->search($afields, $search, $search_mode, true, true, 'email');
./program/actions/mail/autocomplete.php:                if ($result = $abook->search($fields, $search, $mode, true, true, 'email')) {
./program/actions/contacts/mailto.php:                $sources[] = $contacts->search($contacts->primary_key, $cid, 0, true, true, 'email');
./program/actions/contacts/photo.php:                    $result = $abook->search(['email'], $email, 1, true, true, 'photo');
```
Therefore, I would suggest to keep the interface definition as proposed.

program/lib/Roundcube/rcube_addressbook.php
mstilkerich: I adapted the texts from rcube_contacts. As the function is not very comprehensively specified, I also added more extensive description of my understanding that I gathered from studying the roundcube code. I you are open to such documentation, I would also contribute more in this direction for other functions of this interface, which would also help me to get a common understanding / agreement on the interface definition for addressbook implementations.

mstilkerich:
Thanks for your fast feedback, Alec. I really appreciate your responsiveness to pull requests. Please find my comments as replies to yours.

program/lib/Roundcube/rcube_result_set.php
```diff
@@ -21,16 +21,35 @@
...
- * Implenets Iterator and thus be used in foreach() loops.
+ * Implements Iterator and can thus be used in foreach() loops.
  *
  * @package    Framework
  * @subpackage Addressbook
  */
 class rcube_result_set implements Iterator, ArrayAccess
 {
+    /**
+     * @var int $count  The number of total records. Note that when only a subset of records of records is requested,
```
alecpl, line 32: Variable name here is optional, so I propose to not use it. Also, the extra spaces between type and description are not needed.

program/lib/Roundcube/rcube_result_set.php
```diff
@@ -21,16 +21,35 @@
...
     public $first      = 0;
+
+    /**
+     * @var bool $searchonly True if the results are from an addressbook that does not support listing all records but
+     *                       requires the search function to be used.
+     */
     public $searchonly = false;
+
+    /**
+     * @var array $records The data records of he result set. May be a subset of the total records, e.g. for one page.
```
alecpl, line 51: Typo: he -> the.

program/lib/Roundcube/rcube_result_set.php
```diff
@@ -21,16 +21,35 @@
...
  * @package    Framework
  * @subpackage Addressbook
  */
 class rcube_result_set implements Iterator, ArrayAccess
 {
+    /**
+     * @var int $count  The number of total records. Note that when only a subset of records of records is requested,
+     *                  this number may be higher than the number of data records in this result set.
+     */
     public $count      = 0;
```
alecpl, line 35: After adding the comment extra spaces here are not needed and should be removed.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -121,14 +162,36 @@ function refresh_search()
...
      *
-     * @param array  $fields   List of fields to search in
-     * @param string $value    Search value
-     * @param int    $mode     Search mode. Sum of self::SEARCH_*.
-     * @param bool   $select   True if results are requested, False if count only
-     * @param bool   $nocount  True to skip the count query (select only)
-     * @param array  $required List of fields that cannot be empty
+     * Depending on the given parameters the search() function operates in different modes (in the order listed):
+     *
+     * Mode "Direct ID search" - when $fields is either 'ID' or $this->primary_key
```
alecpl, line 167: Nicely written, but the use of word "mode" might be confused with the $mode param below.

alecpl (changes requested):

program/lib/Roundcube/rcube_result_set.php
mstilkerich: Fixed

program/lib/Roundcube/rcube_result_set.php
mstilkerich: Fixed

program/lib/Roundcube/rcube_result_set.php
mstilkerich: I hope I understood correctly that this is about the spaces preceeding the assignment - I removed these at all places where I added comments.

program/lib/Roundcube/rcube_addressbook.php
mstilkerich: Good point. I couldn't think of a suitable alternative term for mode, so I removed terming the different modes of operation (ID search etc.) as mode.

program/lib/Roundcube/rcube_result_set.php
```diff
@@ -21,17 +21,36 @@
...
  *
  * @package    Framework
  * @subpackage Addressbook
  */
 class rcube_result_set implements Iterator, ArrayAccess
 {
-    public $count      = 0;
-    public $first      = 0;
+    /**
+     * @var int The number of total records. Note that when only a subset of records of records is requested,
```
alecpl, line 32: Last thing. Duplicated "of records".

alecpl (changes requested):

program/lib/Roundcube/rcube_result_set.php
mstilkerich: Thanks, fixed.

PR #7821, opened 2021-01-10
ensure rcmail_output object exists

johndoh:
From reading the code I think the newmail_notifier plugin has a similar issue to https://github.com/johndoh/roundcube-contextmenu/issues/126. patch to fix.


alecpl:
I don't think this is a problem, because of `public $task = 'mail|settings';`.

alecpl:
BTW, I think we should move output initialization in cli mode from clisetup.php to rcmail::startup().

johndoh:
confirmed the task check mitigates the issue. sorry for the noise.

PR #7823, opened 2021-01-10
improve detection of external urls for redirect

johndoh:
possible alternate approach for #7751. Use built in PHP URL detection.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -608,8 +608,12 @@ EOF;
      */
     public function redirect($p = [], $delay = 1, $secure = false)
     {
-        if (!empty($this->env['extwin'])) {
-            $p['extwin'] = 1;
+        if (!empty($this->env['extwin']) && !(is_string($p) && preg_match('#^https?://#', $p))) {
+            if (!is_array($p)) {
+                $p = ['_action' => $p];
+            }
+
+            $p['_extwin'] = 1;
         }
 
         $location = $this->app->url($p, false, false, $secure);
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -1032,7 +1032,7 @@ class rcmail extends rcube
     public function url($p, $absolute = false, $full = false, $secure = false)
     {
         if (!is_array($p)) {
-            if (strpos($p, 'http') === 0) {
+            if (preg_match('#^https?://#', $p)) {
                 return $p;
             }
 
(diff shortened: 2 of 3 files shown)


alecpl:
I don't think we have to be so strict. What if the url is somehow invalid, but is a string starting with `http://` (or `https://`). I think all we need is a simple prefix regexp.

Also, maybe it is not a big deal, but if we don't add _extwin=1 to the string url, then maybe it should be at least documented. Finally, do we use this `extwin` argument anywhere? Maybe we don't really need this part.

johndoh:
You think a check like `preg_match('#^http[s]?://#', $str)` would be enough?

`extwin` I'm not sure about. I don't think it needs to be or should be appended to the full URLs because if you are redirecting to a full URL then its going to be outside of Roundcube right? and so it would be meaningless or have unpredictable meaning possibly. If full URL is provided then I don't think it should be altered.

In the redirect function I think it could be changed to `_extwin` as that is more correct for the URL param. The `_` is auto added in the `rcmail::redirect()` function but three is no need to rely on it.  The presence of the param in URLs is certainly used for actions which load in popups but if its really needed in this function I don't know.

alecpl:
> You think a check like `preg_match('#^http[s]?://#', $str)` would be enough?

Yes. The `[s]?` can be just `s?`

> `extwin` I'm not sure about. I don't think it needs to be or should be appended to the full URLs because if you are redirecting to a full URL then its going to be outside of Roundcube right? and so it would be meaningless or have unpredictable meaning possibly. If full URL is provided then I don't think it should be altered.

I was thinking about investigating if the extwin param is really needed here, but maybe it's not worth the effort and we can just ignore it if the input is a string url, indeed.

alecpl:
I suggest to get rid of the utility function at all.

johndoh:
utility function removed

program/include/rcmail_output_html.php
```diff
@@ -608,11 +608,23 @@ public function reset($all = false)
      */
     public function redirect($p = [], $delay = 1, $secure = false)
     {
-        if (!empty($this->env['extwin'])) {
-            $p['extwin'] = 1;
-        }
+        $parser = function($input) use ($secure) {
```
alecpl, line 611: There's no reason for a lambda function here.

johndoh: For me it makes the code cleaner. Otherwise you end up checking if the param is its not an array then if its a url then if it is an array (to add the external option). With the function you can just stop processing if its  url and there is one less array check.

PR #7439, opened 2020-06-16
Add httpapi password change driver to connect to generic HTTP/HTTPS APIs

davidc:

PR #7824, opened 2021-01-10
Added password driver for Mail-in-a-Box

ashtokalo:
The driver allows to change password for email through API of Mail-in-a-Box service (https://mailinabox.email)

### DIFF
--- a/plugins/password/config.inc.php.dist
+++ b/plugins/password/config.inc.php.dist
@@ -471,3 +471,12 @@ $config['password_kpasswd_cmd'] = '/usr/bin/kpasswd';
 // ---------------------
 // put token number from Modoboa server
 $config['password_modoboa_api_token'] = '';
+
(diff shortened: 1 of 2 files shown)


alecpl:
1. All new code should use Guzzle HTTP Client instead of curl.
2. Is there no way to change password using the current user credentials instead of the admin credentials?
3. For drivers like this maybe a better option would be to use the universal driver as proposed in #7439.

ashtokalo:
>     1. All new code should use Guzzle HTTP Client instead of curl.
Done
 
>     2. Is there no way to change password using the current user credentials instead of the admin credentials?
[Unfortunately, no](https://mailinabox.email/api-docs.html#operation/setMailUserPassword) :(

>     3. For drivers like this maybe a better option would be to use the universal driver as proposed in #7439.
It's up to you. But I don't believe to universal solutions. Finally they become too complex and too buggy.

PR #7825, opened 2021-01-12
IMAP "semi-asynchronous" backwards search in batches

lennartack:
IMAP searches can take a very long time if you have a large inbox. In my inbox it is common that the HTTP timeout occurs before the search has finished. This feature is intended to resolve this problem and would greatly enhance user experience, in my opinion.

If you turn on "backwards" search, the IMAP search command will use an RFC 3501 message sequence range to search the _n_ most recent messages in the folder. Once the web client receives a reply, it will send a search request for the next _n_ emails in the folder. I have hardcoded _n_ to 500 for body and full text searches and 5000 for other searches.

**Open issues:**
1. Everything is sorted DESC according to the internal IMAP server arrival date and the user cannot change this. This should still be made clear in the UI.
2. There is no implementation for multi folder search yet. Perhaps this should be disabled and made clear in the UI.
3. ~~Backwards search with threads listing is still buggy. The threads implementation looks complicated and I have run out of time to study it.~~
4. The feature is only available with the Elastic skin.
5. The length _n_ of the sequence range is set client-side, which is a bit awkward. It should probably be a server-side configuration parameter instead.
6. Maybe it should also be implemented for contacts search.

### DIFF
--- a/program/actions/mail/search.php
+++ b/program/actions/mail/search.php
@@ -132,6 +132,9 @@ class rcmail_action_mail_search extends rcmail_action_mail_index
         
 
         if ($search_str) {
+            if (!empty($range)) {
+                $rcmail->storage->search_disable_sort();
+            }
             $_SESSION['search'] = $rcmail->storage->get_search_set();
             $_SESSION['last_text_search'] = $str;
         }
@@ -179,7 +182,7 @@ class rcmail_action_mail_search extends rcmail_action_mail_index
         else if (!empty($result) && !empty($result->incomplete)) {
             $rcmail->output->command('continue_search', $search_request);
         }
-        else if (empty($range) || $last_batch && ($count == 0 || $range[0] == 0)) {
+        else if (empty($range) || isset($last_batch) && ($count == 0 || $range[0] == 0)) {
             $rcmail->output->show_message('searchnomatch', 'notice');
             $rcmail->output->set_env('multifolder_listing', isset($result) ? !empty($result->multi) : false);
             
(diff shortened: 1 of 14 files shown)


alecpl:
I didn't review yet, but I think I'd prefer this to be implemented as a plugin. This probably makes most sense on a slow imap server. There are setups that have no problem in searching even very big folders. The open issues make it not ready and questionable.

lennartack:
> I didn't review yet, but I think I'd prefer this to be implemented as a plugin. This probably makes most sense on a slow imap server. There are setups that have no problem in searching even very big folders. The open issues make it not ready and questionable.

With a plugin implementation, would it be possible to run the backwards search with the normal search form in the UI? That is all that matters to me.

If you provide some feedback on the code and the issues I will continue this project once I have time.

bsdice:
> IMAP searches can take a very long time if you have a large inbox

I just stumbled on this ticket. Not sure if you use dovecot or something else, if dovecot, try the dovecot-fts-xapian fulltext-search plugin. On Arch, e.g. from https://archlinux.org/packages/community/x86_64/dovecot-fts-xapian/ with configuration similar to https://github.com/grosjo/fts-xapian/commit/1e6d3b28389b2d1adac0dcf9772ecb71629e8d8b

Anecdotally I have around 160.000 messages in IMAP Archive/ folders going back to 1995, in total ~24 GB for the Maildir. Using this plugin for dovecot I could swear fulltext search will return results faster than a Google search. I also use attachment decoders to make PDF, XLS etc. searchable. Previously I relied on Thunderbird, and Outlook before that, to find stuff in my Inbox, but this solution provides superior search results faster compared to those two fat clients. The server is usually lightly loaded and not even particularly beefy, a Xeon E3 1275 v6 with 64 GB ECC RAM from 5 years ago and an Intel P3520 PCIe SSD the same age.

PR #7834, opened 2021-01-16
Add hook to modify the hint in the upload form

gianlucagiacometti:
I find this hook useful to add a more detailed description in case of attachment manipulation (i.e. replacing attachment with a download link) or company's restrictions on attachments.
Would you mind adding it?
Thank you anyway

### DIFF
--- a/program/include/rcmail_action.php
+++ b/program/include/rcmail_action.php
@@ -505,6 +505,9 @@ abstract class rcmail_action
 
         $hint = html::div('hint', $rcmail->gettext(['name' => 'maxuploadsize', 'vars' => ['size' => $max_filesize]]));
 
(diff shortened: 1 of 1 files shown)


alecpl:
I really don't like a hook that does just that. I guess, you can modify the hint in javascript or with `template_object_composeattachmentform` hook.

gianlucagiacometti:
Thank you for your response. I will investigate the template_object_composeattachmentform hook.
Kind regards

gianlucagiacometti:
Thank you very much for the hint! This function, connected to the hook, solved the problem
```
    public function update_hint($args) {
        if ($args['mode'] == 'hint') {
            $args['content'] = $args['content'] . "added content";
        }
        return $args;
    }
```

PR #7844, opened 2021-01-23
retry login if IMAP Servers timeout.

hefee:
When connection fails in a recoverable way, we should retry to connect
to the server again. This should help in cases, when the connection to the IMAP server is not 100% stable.

In #7402 it was suggested to implement this retry feature as plugin. In order to create such a plugin, it needs the information about the last login attempt. The last attempt is added to the `data['conn']`.

We also face the issue, that we only call `storage_connect` before the try to connect. That's why we need a way to stop instantly when no new attempt should be done.

(followup of #7402, #5393)

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -157,6 +157,10 @@ class rcube_imap extends rcube_storage
 
             $data = $this->plugins->exec_hook('storage_connect', array_merge($this->options, $data));
 
+            if ($attempt > 1 && !$data['retry']) {
+                $break;
+            }
+
             if (!empty($data['pass'])) {
                 $pass = $data['pass'];
             }
--- a/plugins/reconnect/composer.json
+++ b/plugins/reconnect/composer.json
@@ -1,0 +1,13 @@
+{
+    "name": "reconnect",
+    "type": "roundcube-plugin",
+    "description": "Reconnects to server for several attempts.",
+    "license": "GPLv3+",
+    "version": "0.1",
+    "authors": [
+        {
+            "name": "Sandro Knauß",
+            "email": "hefee@debian.org"
+        }
+    ]
+}
--- a/plugins/reconnect/config.inc.php.dist
+++ b/plugins/reconnect/config.inc.php.dist
@@ -1,0 +1,4 @@
+<?php
+
+// Maximum attempts to connect the IMAP server
+$config['reconnect_imap_max_attempts'] = 5;
--- a/plugins/reconnect/readme.md
+++ b/plugins/reconnect/readme.md
@@ -1,0 +1,13 @@
+# RoundCube Reconnect Plugin
+
+RoundCube reconnect Plugin is a small plugin that will try to reconnect to an
+IMAP server, if there is no explicit error code replied. If there is a know
+failure like wrong password, no additional attempts are triggered. This should
+help in cases, when the connection to the IMAP server is not 100% stable.
+
+## Configuration
+
+You can specify the maximum attempts to connect the IMAP server.
+
+  // Maximum attempts to connect the IMAP server
+  $config['reconnect_imap_max_attempts'] = 5;
--- a/plugins/reconnect/reconnect.php
+++ b/plugins/reconnect/reconnect.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * RoundCube Reconnect Plugin
+ *
+ * @version 0.1
+ * @author Sandro Knauß <hefee@debian.org>
+ * @license GPLv3+
+ */
+class reconnect extends rcube_plugin
+{
+    private $max_attempts;
+
+    function init()
+    {
+        $rcmail = rcmail::get_instance();
+
+        $this->load_config();
+
+        $this->imap_max_attempts = $rcmail->config->get('reconnect_imap_max_attempts', 5);
+
+        $this->add_hook('storage_connect', array($this, 'storage_connect'));
+    }
+
+    function storage_connect($args)
+    {
+        $args['retry'] = ($args['attempt'] <= $this->imap_max_attempts);
+
+        if ($args['attempt'] == 1) {
+            return $args;
+        }
+
+        $storage = rcmail::get_instance()->get_storage();
+        switch ($storage->get_error_code()) {
+        case rcube_imap_generic::ERROR_NO:
+        case rcube_imap_generic::ERROR_BAD:
+        case rcube_imap_generic::ERROR_BYE:
+            $args['retry'] = false;
+            break;
+        }
+
+        if ($args['retry']) {
+            // if we do a new attempt, sleep 50 to 150ms before retry.
+            usleep(rand(50*1000, 150*1000));
+        }
+
+        return $args;
+    }
+
+}
+
+?>


plugins/retrylogin/readme.md
```diff
@@ -0,0 +1,10 @@
+# RoundCube RetryLogin Plugin
+
+RoundCube RetryLogin Plugin is a small plugin that will try to reconnect to an IMAP server, if there is no explicit error code replied. If there is a know failure like wrong password, no additional attempts are triggered. This should help in cases, when the connection to the IMAP server is not 100% stable.
+
+## Configuration
+
+You can specify the maximum attempts to connect the IMAP server.
+
+  // Maximum attempts to connect the IMAP server
+  $config['retrylogin_max_attempts'] = 5;
```
alecpl, line 10: Wouldn't be better to specify this as a time interval?

plugins/retrylogin/composer.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "retrylogin",
```
alecpl, line 2: I'm not sure I like the plugin name. Maybe `imap_reconnect` or `imap_retry` would be better.

plugins/retrylogin/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
alecpl, line 1: We don't need the LICENSE file here.

program/lib/Roundcube/rcube_imap.php
```diff
@@ -152,11 +152,16 @@ public function connect($host, $user, $pass, $port = 143, $use_ssl = null)
                 'user'           => $user,
                 'attempt'        => ++$attempt,
                 'socket_options' => [],
+                'conn'           => $this->conn,
```
alecpl, line 155: $this->conn is a public property so you don't really need this.

plugins/retrylogin/retrylogin.php
```diff
@@ -0,0 +1,52 @@
...
+
+    $this->add_hook('storage_connect', array($this, 'storage_connect'));
+  }
+
+  function storage_connect($args)
+  {
+    $args['retry'] = ($args['attempt'] <= $this->max_attempts);
+
+    if ($args['attempt'] == 1) {
+        rcube::write_log('imap','Connecting to IMAP server attempt:' . $args['attempt']);
```
alecpl, line 29: We write to imap log only in debug mode. And in this mode we already have "Connecting to..." in the log. So, I would just remove this.

plugins/retrylogin/retrylogin.php
```diff
@@ -0,0 +1,52 @@
...
+        case rcube_imap_generic::ERROR_BAD:
+        case rcube_imap_generic::ERROR_BYE:
+            $args['retry'] = false;
+            break;
+    }
+
+    if ($args['retry']) {
+        // if we do a new attempt, sleep 50 to 150ms before retry.
+        usleep(rand(50*1000, 150*1000));
+        rcube::write_log('imap','Connecting to IMAP server attempt:' . $args['attempt']);
```
alecpl, line 44: Remove this write_log.

program/lib/Roundcube/rcube_imap.php
```diff
@@ -152,11 +152,16 @@ public function connect($host, $user, $pass, $port = 143, $use_ssl = null)
...
                 'retry'          => false
             ];
 
             $data = $this->plugins->exec_hook('storage_connect', array_merge($this->options, $data));
 
+            if ($attempt > 1 && !data['retry']) {
```
alecpl, line 161: You don't need this `if` statement. We check `retry` flag in the while statement below. You just set it appropriately from the plugin.

alecpl (changes requested):

plugins/retrylogin/retrylogin.php
hefee: I use roundcube 1.3.15 in production and there I do not see any "Connecting to..." line in imap log, nor can I see that this line is somewhere created in `program/lib/Roundcube/rcube_imap.php`. But things may change deeper in the stack.

I cannot really see in the communication with the server and I think it is good to know that the retry plugin is doing further attempts to login:

```
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> Connecting to IMAP server attempt:1
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] S: * OK [CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+ STARTTLS LOGINDISABLED] Dovecot (Debian) ready.
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] C: A0001 STARTTLS
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] S: A0001 OK Begin TLS negotiation now.
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] C: A0002 CAPABILITY
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] S: * CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+ AUTH=PLAIN AUTH=LOGIN
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] S: A0002 OK Pre-login capabilities listed, post-login capabilities have more.
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] C: A0003 ID ("name" "Roundcube" "version" "1.3.15" "php" "7.3.19-1~deb10u1" "os" "Linux" "command" "/?_task=login")
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] S: * ID ("name" "Dovecot")
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] S: A0003 OK ID completed.
[23-Jan-2021 03:11:01 +0100]: <goet9ij9> [5055] C: A0004 AUTHENTICATE PLAIN ****** [37]
[23-Jan-2021 03:11:03 +0100]: <goet9ij9> [5055] S: A0004 NO [AUTHENTICATIONFAILED] Authentication failed.
[23-Jan-2021 03:11:03 +0100]: <goet9ij9> Connecting to IMAP server attempt:2
[...]
[23-Jan-2021 03:11:09 +0100]: <goet9ij9> Connecting to IMAP server attempt:3
```

program/lib/Roundcube/rcube_imap.php
hefee: But how do the plugin get access to the current instance of rcube_imap?

program/lib/Roundcube/rcube_imap.php
hefee: here is an example without this if statement:
*attempt=1, retry= True
  -> do a `$this->conn->connect`
  -> while loop decide to go on
*attempt=2 checks the last attempt and decides not to continue retry=False
  -> still do a  `$this->conn->connect`
  -> while loop stops

as you see without this if statement another reconnect attempt is executed, but this is not what we want to do. The problem we face is that we call the storage_connect hook only before we connect and not after it. This if statement makes it possible to stop immediately and not do an unexpected reconnect attempt.

I know it is a behavior change, so we could also use another key in the array to tell to stop immediately.

plugins/retrylogin/readme.md
hefee: You mean define X attempts per Y mins or you want to define the sleep between the attempts?

program/lib/Roundcube/rcube_imap.php
alecpl: rcube::get_storage()

plugins/retrylogin/retrylogin.php
alecpl: https://github.com/roundcube/roundcubemail/blob/1eebfd723bc6706d76c280dbeaf708d4ba8f010d/program/lib/Roundcube/rcube_imap_generic.php#L1043 If you really think it is useful in non-debug mode, make it a config option in the plugin.

program/lib/Roundcube/rcube_imap.php
alecpl: I see, but I still don't like it. I'll consider this. BTW, there's missing `$` character.

plugins/retrylogin/readme.md
alecpl: I meant to define maximum time from the request start to the moment you give up. I think it might be more useful that a number of attempts, but maybe it does not really matter.

plugins/retrylogin/retrylogin.php
hefee: Okay I think this is enough.

program/lib/Roundcube/rcube_imap.php
hefee: It needs to be `rcmail::get_instance()->get_storage()` at least this this line I can find in other plugins.

plugins/retrylogin/readme.md
hefee: The current bugs #7402, #5393 both around of flaky connection to IMAP server, where attempts seems the better solution. Maybe the word timeout in title is misleading. It also handles the case, when the IMAP server closes with ""Empty startup greeting". A timeout setting would mean that the plugin needs to store the the timestamp of each first attempt to calculate the timeout, this makes the plugin more complex and we need to store this additional data.

program/lib/Roundcube/rcube_imap.php
hefee: Fixed the `$` character.

I can fully understand that it does not looks not very clean. If you have any other idea...

plugins/imap_reconnect/imap_reconnect.php
```diff
@@ -0,0 +1,51 @@
...
+  function storage_connect($args)
+  {
+    $args['retry'] = ($args['attempt'] <= $this->max_attempts);
+
+    if ($args['attempt'] == 1) {
+        return $args;
+    }
+
+    $storage = rcmail::get_instance()->get_storage();
+    switch ($storage->conn->errornum) {
```
alecpl, line 33: You could also use more official api: $storage->get_error_code().

plugins/imap_reconnect/config.inc.php.dist
```diff
@@ -0,0 +1,6 @@
+<?php
+
+// Maximum attempts to connect the IMAP server
+$config['imap_reconnect_max_attempts'] = 5;
+
+?>
```
alecpl, line 6: No closing tags, please.

plugins/imap_reconnect/composer.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "imap reconnect",
```
alecpl, line 2: 4 spaces indent, please.

plugins/imap_reconnect/imap_reconnect.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * RoundCube Imap Reconnect Plugin
+ *
+ * @version 0.1
+ * @author Sandro Knauß <hefee@debian.org>
+ * @license GPLv3+
+ */
+class imapreconnect extends rcube_plugin
+{
+  private $max_attempts;
```
alecpl, line 11: 4 spaces indent, please.

alecpl (changes requested):

plugins/imap_reconnect/config.inc.php.dist
hefee: done.

plugins/imap_reconnect/imap_reconnect.php
hefee: done.

alecpl:
Now I have a question. Why would IMAP connection be unstable and not others. Shouldn't we implement the same for at least SMTP? What about managesieve?

Maybe the plugin should be named just 'reconnect', so we can extend it with code for other connection types in future, if needed.

hefee:
> Now I have a question. Why would IMAP connection be unstable and not others. Shouldn't we implement the same for at least SMTP? What about managesieve?

Well I havn't seen the root of issue for quite a while, but when we debugged the issue, we didn't found out why the IMAP connection were closed by the IMAP server only that ~1% failing and retry solved the issue. I think it may happen also for SMTP and managesieve, but as the amount of connections to SMTP and managesieve are several of magnitudes lower than IMAP, it is properly harder to see the issue.

> Maybe the plugin should be named just 'reconnect', so we can extend it with code for other connection types in future, if needed.

ACK. I understood you correctly, that I do not need to invenst time to add these connection types?

alecpl (approved):

PR #7859, opened 2021-01-27
Try handle body part with body.peek when binary.peek cause an error on imap server (dovecot)

lucasmarin:
**RC Version:** 1.4.10
**Dovecot Version:** 2.3.11.3
**Capability:** [CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE SORT SORT=DISPLAY THREAD=REFERENCES THREAD=REFS THREAD=ORDEREDSUBJECT MULTIAPPEND URL-PARTIAL CATENATE UNSELECT CHILDREN NAMESPACE UIDPLUS LIST-EXTENDED I18NLEVEL=1 CONDSTORE QRESYNC ESEARCH ESORT SEARCHRES WITHIN CONTEXT=SEARCH LIST-STATUS BINARY MOVE SNIPPET=FUZZY PREVIEW=FUZZY STATUS=SIZE SAVEDATE LITERAL+ NOTIFY SPECIAL-USE QUOTA]

Message examples: https://skybox.skymail.net.br/s/N5NLnZoeM9boaJm (Protected by password, please send a message to me to get it) 

These messages are being successfully read in Gmail, Thunderbird and Outlook.

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -2413,6 +2413,17 @@ class rcube_imap extends rcube_storage
             $formatted = $formatted && $o_part->ctype_primary == 'text';
             $body = $this->conn->handlePartBody($this->folder, $uid, true,
                 $part ? $part : 'TEXT', $o_part->encoding, $print, $fp, $formatted, $max_bytes);
+            if (!$body && $this->conn->errornum === rcube_imap_generic::ERROR_FETCH_BINARY) {
+                $this->connect(
+                    $this->options['host'],
+                    $this->options['user'],
+                    $this->options['password'],
+                    $this->options['port'],
+                    $this->options['ssl']
+                );
+                $body = $this->conn->handlePartBody($this->folder, $uid, true,
+                    $part ? $part : 'TEXT', $o_part->encoding, $print, $fp, $formatted, $max_bytes, false);
+            }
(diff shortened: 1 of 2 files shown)


alecpl:
I don't like it. What version of dovecot are you using?

lucasmarin:
Hi @alecpl. I updated the PR with more info.

If you want to downlod the message examples, please send me an e-mail on lucasmarin@gmail.com.

Thx

alecpl:
I would rather ask dovecot developers to do something with it, because it's an evident dovecot bug.

lucasmarin:
I think we can ask dovecot to do something about it too... and also deal with dovecot erros retrying read these e-mails with another way (body.peek). 

Maybe this is not the better way to do it, but the fact another clients read these e-mail doesn't make you think about why RC don't?

alecpl:
The clients you mentioned just don't use binary fetches, they fetch the whole message and parse it by themselves. That's the difference. For performance reasons Roundcube uses other standard ways and depends on them being not buggy.

alecpl:
Reconnecting on error is that kind of a server bug workaround I don't like in general.

alecpl:
BTW, you can disable use of BINARY either via Roundcube config (`imap_disabled_caps`) or via Dovecot config.

Issue #6411, opened 2018-08-27
SMTP XCLIENT support


lgblgblgb:
Scenario: I have a fairly complex postfix configuration for SMTP mail submission server, involving many postfix table lookups (including net based ones, SQL, pcre, internal RBLs for *clients* - yes), policy servers and so on to try to mitigate the impact of stolen passwords exploited then by spammers to push spam out. Roundcube is configured to use this SMTP server as well. Surely, this requested feature can be useful for other scenarios as well, I guess.

However, the problem, that postfix cannot do the same checks as with bare SMTP connections, since the client is always my roundcube server's IP address. And so on. Since postfix supports the XCLIENT ESMTP extension, it's possible to "inject" the information to tell postfix what it should think about the origin of the mail submission SMTP connection (surely, usage of XCLIENT must be configured to allow a very trusted, source, ie internal behind-the-firewall roundcube server, not for the whole world ...). Without this, I would need to duplicate all the mechanisms into roundcube what postfix can check for SMTP clients from the net. If roundcube can extend its smtp session handling to optionally support XCLIENT, it would be able to pass the origin of the HTTP connection instead as the source IP (and probably other info as well) for the postfix SMTP server. I am not sure if XCLIENT is supported by other MTAs as well, but at least postfix surely knows about it.

Here is a little document also with a sample SMTP session to demonstrate the feature:

http://www.postfix.org/XCLIENT_README.html

(just to note, postfix also know XFORWARD, which is a different topic though)

lgblgblgb:
By the way, I had a look on the roundcube plugin documentation if it's possible to "inject" extra commands in the SMTP session with a roundcube plugin to implement this (without a core change, as a plugin), but since it's my first try to look on the source, I am not sure if it's possible. I would avoid to hack into the basic smtp client such a feature otherwise ...

alecpl:
It is not possible with the plugin API. You have to add some code around line 141 of rcube_smtp.php file, i.e. after connect() method use. You could use $this->conn->command() to send the XCLIENT command. See lines 130-140 to see how to use Net_SMTP object and handle errors. First you probably should use getServiceExtensions('XCLIENT') to check for the extension support.
Pull requests are welcome.

lgblgblgb:
[...]

PR #7891, opened 2021-02-18
SMTP: Add support for sending XCLIENT LOGIN

kwiatek6324:
PR created from issue: https://github.com/roundcube/roundcubemail/issues/6411

It will send XCLIENT LOGIN=webmailuser@domain.com when configured.
In some situation there is no possibility use  smtp_user and smtp_pass,
but we want to process outgoing message as authenticated.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -292,6 +292,9 @@ $config['smtp_auth_cid'] = null;
 // Optional SMTP authentication password to be used for smtp_auth_cid
 $config['smtp_auth_pw'] = null;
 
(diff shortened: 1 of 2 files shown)


alecpl:
The ticket was about `ADDR` attribute, so it looks like you're trying to solve a different issue.

According to Postfix docs "Some Postfix implementations do not implement the PORT or LOGIN attributes.". So, maybe we should check if the attribute is supported.

kwiatek6324:
@alecpl ok, will prepare new version.

PR #7893, opened 2021-02-19
Add support for XCLIENT ADDR and LOGIN extension

kwiatek6324:
PR created from issue: #6411

PR Checks if XCLIENT extension is active.

Roundcube will send XCLIENT ADDR=x.y.z.z when configured and XCLIENT ADDR extension is available
In special configuration we want smtp server to use originating IP for policy checking instead of IP from web server.

Roundcube will send XCLIENT LOGIN=webmailuser@domain.com when configured and XCLIENT LOGIN extension is available
In some situation (f.e. SSO) there is no possibility use smtp_user and smtp_pass, but we want to process outgoing message as authenticated.

### DIFF
--- a/program/lib/Roundcube/rcube_smtp.php
+++ b/program/lib/Roundcube/rcube_smtp.php
@@ -164,6 +164,14 @@ class rcube_smtp
             $this->conn->setTimeout($timeout);
         }
 
+        if (!$this->_process_xclient()) {
+            list($code,) = $this->conn->getResponse();
+            $this->error = ['label' => 'smtpconnerror', 'vars' => ['code' => $code]];
+            $this->conn  = null;
+
+            return false;
+        }
+
         $smtp_user = str_replace('%u', $rcube->get_user_name(), $CONFIG['smtp_user']);
         $smtp_pass = str_replace('%p', $rcube->get_user_password(), $CONFIG['smtp_pass']);
         $smtp_auth_type = $CONFIG['smtp_auth_type'] ?: null;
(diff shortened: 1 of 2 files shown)


program/lib/Roundcube/rcube_smtp.php
```diff
@@ -164,6 +164,14 @@ public function connect($host = null, $port = null, $user = null, $pass = null)
...
         }
 
+        $exts = $this->conn->getServiceExtensions();
+        if (isset($exts['XCLIENT']) && ($rcube->config->get('smtp_xclient_login')==1 || $rcube->config->get('smtp_xclient_addr')==1) ) { 
+            $cmd="";
+            if ($rcube->config->get('smtp_xclient_login')==1 && preg_match("/LOGIN/",$exts['XCLIENT'])) { $cmd.=" LOGIN=".$rcube->get_user_name(); } 
+            if ($rcube->config->get('smtp_xclient_addr')==1 && preg_match("/ADDR/",$exts['XCLIENT'])) { $cmd.=" ADDR=".rcube_utils::remote_addr(); } 
+            if ($cmd!="") { $this->conn->command("XCLIENT".$cmd,array(220)); } 
+            }
+
```
alecpl, line 174: 1. Your code style is really bad.
2. The config options are bool, so `==1` is not needed.
3. The regexp for ADDR will match also DESTADDR. I would propose to use explode() and in_array_nocase() instead of regexps. And yes, we should use case insensitive comparison.
4. If the IP address is an IPv6 address, it should be prefixed with `IPV6:`. If for some reason remote_addr() returns no address `[UNAVAILABLE]` label should be used.
5. Move the code to a separate method.
6. Limit number of config->get() calls.

alecpl (changes requested):

program/lib/Roundcube/rcube_smtp.php
kwiatek6324: @alecpl Corrected version.

alecpl:
The error handling is probably redundant, and your code style is still not acceptable.

alecpl (changes requested):
Fix code style, please.

PR #7921, opened 2021-03-08
Add managesieve hook to modify extensions

tomsommer:
Add plugin hook to managesieve to manipulate extensions returned by server

### DIFF
--- a/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
+++ b/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php
@@ -142,7 +142,7 @@ class rcube_sieve_engine
             $this->script = [];
         }
         else {
-            $this->exts = $this->sieve->get_extensions();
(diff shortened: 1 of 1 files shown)


tomsommer:
Also helps out #4615

alecpl:
What is the use-case? The mentionet ticket is very old, and there is a config option to disable specific extensions. You can also set 'disabled' item on `managesieve_connect` hook.

tomsommer:
The use-case is to remove the 'redirect' extension from the UI.

alecpl:
So, it is already possible.

tomsommer:
It's not, managesieve_disabled_extensions does not affect the 'redirect' extension:

        $select_action->add($this->plugin->gettext('messageredirect'), 'redirect');

alecpl:
Redirect is not an extension. Your way will not work either.

tomsommer:
Hrm, correct. I will think of another way

Issue #4981, opened 2008-08-08
On reply, truncate subject prefix in foreign languages


rcubetrac:
_Reported by abaum on 8 Aug 2008 08:18 UTC as Trac ticket #1485258_

There are a lot of misconfigured mail clients (e.g. MS Outlook) out there who use wrong, non-english subject prefixes (e.g. the german edition of Outlook uses "AW:" instead of "Re:"). Now, if you reply on such a mail, the subject will become "Re: AW: Subject" instead of "Re: Subject". Other mail clients can detect this.
We could make an additional entry in the translation files and truncate the wrong prefix.

_Keywords: subject prefix translation_
_Migrated-From: http://trac.roundcube.net/ticket/1485258_

frontierpsycho:
I encountered this bug, it's quite annoying. Are there any plans to work on it? 

If not, if I find time I'll look into it.

alecpl:
Configuration options implemented.

A future scope might be to make these options available to the end-user via Preferences > Composing messages > Advanced options. Pull requests would be appreciated. @tpayen maybe, you?

PR #7929, opened 2021-03-09
Add config options for subject prefixes

tpayen:
Some french mail clients use "Tr:" instead of "Fwd:" as forward indicator in subject. 
I don't know if there is such exceptions in others languages in which case this maybe worth the use of a configuration parameter instead of an hardcoded regex ?

### DIFF
--- a/program/include/rcmail_sendmail.php
+++ b/program/include/rcmail_sendmail.php
@@ -1192,22 +1192,14 @@ class rcmail_sendmail
     {
         $subject = trim($subject);
 
-        // replace Re:, Re[x]:, Re-x (#1490497)
-        $response_prefixes = rcmail::get_instance()->config->get('subject_response_prefixes', ['Re:']);
-        $data = array_map(function($prefix) { 
-            $prefix = strtolower(str_replace(':', '', $prefix)); 
-            return $prefix.':|'.$prefix.'\[\d\]:|'.$prefix.'-\d:'; 
-        }, $response_prefixes);
-        $prefix = '/^('.implode('|', $data).')\s*/i';
-        do {
-            $subject = preg_replace($prefix, '', $subject, -1, $count);
-        }
-        while ($count);
+        //  Add config options for subject prefixes (#7929) 
+        $subject = rcube_utils::remove_subject_prefix($subject, 'reply');
 
         // replace (was: ...) (#1489375)
         $subject = preg_replace('/\s*\([wW]as:[^\)]+\)\s*$/', '', $subject);
+        $subject = rcmail::get_instance()->config->get('response_prefix', 'Re:') . ' ' . $subject;
 
-        return rcmail::get_instance()->config->get('default_response_prefix', 'Re:') . ' ' . $subject;
+        return trim($subject);
     }
 
     /**
@@ -1241,13 +1233,9 @@ class rcmail_sendmail
         }
         // create a forward-subject
         else if ($this->data['mode'] == self::MODE_FORWARD) {
-            $forward_prefixes = rcmail::get_instance()->config->get('subject_forward_prefixes', ['Fwd:', 'Fw:']);
-            if (preg_match('/^('.implode('|', $forward_prefixes).')/i', $this->options['message']->subject)) {
-                $subject = $this->options['message']->subject;
-            }
-            else {
-                $subject = $this->rcmail->config->get('default_forward_prefix', 'Fwd:') . ' ' . $this->options['message']->subject;
-            }
+            //  Add config options for subject prefixes (#7929) 
+            $subject = rcube_utils::remove_subject_prefix($this->options['message']->subject, 'forward');
+            $subject = trim($this->rcmail->config->get('forward_prefix', 'Fwd:') . ' ' . $subject);
         }
         // creeate a draft-subject
         else if ($this->data['mode'] == self::MODE_DRAFT || $this->data['mode'] == self::MODE_EDIT) {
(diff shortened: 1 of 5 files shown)


alecpl:
We have an old ticket for this: #4981. I think it should be a config option with a default that contains most common cases.

tpayen:
Hi Alec,
I've added a new commit to support config options for subject prefixes.
4 more options are possible : 
 - default_forward_prefix (default = 'Fwd:') : the default prefix when forwarding an email
 - default_response_prefix (default = 'Re:') : the default prefix when responding to an email
 - subject_response_prefixes (default = ['Re:']) : the prefixes list to clean for responses
 - subject_forward_prefixes (default = ['Fwd:', 'Fw:']) : the prefixes list to clean for forwarded messages

program/include/rcmail_sendmail.php
```diff
@@ -1202,7 +1207,7 @@ public static function reply_subject($subject)
         // replace (was: ...) (#1489375)
         $subject = preg_replace('/\s*\([wW]as:[^\)]+\)\s*$/', '', $subject);
 
-        return 'Re: ' . $subject;
+        return rcmail::get_instance()->config->get('default_response_prefix', 'Re:') . ' ' . $subject;
```
alecpl, line 1210: I would remove `default_` from these option names. Also, what if it is set to an empty string? You should trim the return value.

program/lib/Roundcube/rcube_imap_generic.php
```diff
@@ -2742,7 +2742,10 @@ public static function sortHeaders($messages, $field, $order = 'ASC')
                     $value = str_replace('"', '', $value);
 
                     if ($field == 'subject') {
-                        $value = preg_replace('/^(Re:\s*|Fwd:\s*|Fw:\s*)+/i', '', $value);
+                        $forward_prefixes = rcmail::get_instance()->config->get('subject_forward_prefixes', ['Fwd:', 'Fw:']);
+                        $response_prefixes = rcmail::get_instance()->config->get('subject_response_prefixes', ['Re:']);
+                        $data = array_map(function($prefix) { return $prefix . '\s*'; }, array_merge($forward_prefixes, $response_prefixes));
+                        $value = preg_replace('/^('.implode('|', $data).')+/i', '', $value);
```
alecpl, line 2748: So, we have similar code in a few places. I would suggest to unify and de-duplicate this.

I.e. create `rcube_utils::remove_subject_prefix($subject, $mode)` where mode would be 'reply', 'forward' or 'both'.

alecpl (changes requested):

tpayen:
Thanks for these relevant remarks. I've added a new commit to add the `rcube_utils::remove_subject_prefix` function and make everything a little cleaner

program/include/rcmail_sendmail.php
```diff
@@ -1192,17 +1192,14 @@ public static function reply_subject($subject)
...
-        $prefix = '/^(re:|re\[\d\]:|re-\d:)\s*/i';
-        do {
-            $subject = preg_replace($prefix, '', $subject, -1, $count);
-        }
-        while ($count);
+        //  Add config options for subject prefixes (#7929) 
+        $subject = rcube_utils::remove_subject_prefix($subject, 'reply');
 
         // replace (was: ...) (#1489375)
         $subject = preg_replace('/\s*\([wW]as:[^\)]+\)\s*$/', '', $subject);
```
alecpl, line 1199: How about moving this into remove_subject_prefix() (in 'reply' mode)?

program/lib/Roundcube/rcube_utils.php
```diff
@@ -1580,4 +1580,41 @@ public static function temp_filename($file_name, $unique = true, $create = true)
...
+
+    /**
+     * Clean the subject from reply and forward prefix
+     * 
+     * @param string $subject Subject to clean
+     * @param string $mode Mode of cleaning : reply, forward or both
+     * 
+     * @return string Cleaned subject
+     */
+    public static function remove_subject_prefix($subject, $mode = 'both')
```
alecpl, line 1592: How do you feel about adding some tests in tests/Framework/Utils.php?

program/lib/Roundcube/rcube_utils.php
```diff
@@ -1580,4 +1580,41 @@ public static function temp_filename($file_name, $unique = true, $create = true)
...
+    {
+        // Clean subject prefix for reply, forward or both
+        if ($mode == 'both') {
+            $reply_prefixes = rcmail::get_instance()->config->get('subject_reply_prefixes', ['Re:']);
```
alecpl, line 1596: Let's do `$config = rcmail::get_instance()->config` first and then use it to make the code shorter.

program/lib/Roundcube/rcube_utils.php
```diff
@@ -1580,4 +1580,41 @@ public static function temp_filename($file_name, $unique = true, $create = true)
...
+            $prefixes = rcmail::get_instance()->config->get('subject_reply_prefixes', ['Re:']);
+        }
+        else if ($mode == 'forward') {
+            $prefixes = rcmail::get_instance()->config->get('subject_forward_prefixes', ['Fwd:', 'Fw:']);
+        }
+
+        // replace Re:, Re[x]:, Re-x (#1490497)
+        $pieces = array_map(function($prefix) {
+            $prefix = strtolower(str_replace(':', '', $prefix)); 
+            return $prefix.':|'.$prefix.'\[\d\]:|'.$prefix.'-\d:';
```
alecpl, line 1610: Add a space between dots, please. Or everything using double-quote'd syntax.

alecpl (changes requested):

program/lib/Roundcube/rcube_utils.php
tpayen: I'll try but I don't promise anything  ;-)
I'll do an independent commit for this

PR #7956, opened 2021-03-18
Fix phpdoc for rcube::decrypt()

mstilkerich:
openssl_decrypt() may return false on error (e.g. if the cipher cannot
be decrypted with the given key), and therefore this function can also
return false.

Note: Alternatively, it would be possible to return an empty string like the function does for other errors. I don't know what's intended for this interface and went without code change.

### DIFF
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -924,9 +924,9 @@ class rcube
      *
      * @param string $cipher Encrypted text
      * @param string $key    Encryption key to retrieve from the configuration, defaults to 'des_key'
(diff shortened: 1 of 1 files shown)


alecpl:
We should probably change all `return ''` there to `return false`.

mstilkerich:
If you are not afraid that something will break :-) Did as requested.

I also adapted the return type of encrypt(), which also returns false on error. For encrypt, it returns an empty string if given something else than a string to encrypt (certainly an error) or given an empty string to encrypt (not sure if you consider this an error). Maybe this should be changed to return false as well? If so, only if string is non-string type, or also for an empty string?

Issue #7522, opened 2020-07-30
Could not forward message flagged as spam on 1.4.7


samuelmf:
I coulnd't forward messages flagged as ***SPAM***  by spam assasin i get a SMT Error 550.
But on my old roundcube version on the server i'm able to forward messages flagged as spam.

It's possible to disable outgoing message check if its on spam, on the roundcube configuration or any way to fix.

By the moment the only way to bypass the outgoing messages check is using the option Resend (bounce).

alecpl:
Sounds like an SMTP issue. I'm not sure I understand it. Could you enable the smtp_debug option and provide the log?

samuelmf:
> Sounds like an SMTP issue. I'm not sure I understand it. Could you enable the smtp_debug option and provide the log?

Hi, thanks for reply, how can enable that option? on the config file?

samuelmf:
I  have set this lines on the config file
`// system error reporting, sum of: 1 = log; 4 = show
$config['debug_level'] = 1;

// Log SQL queries
$config['sql_debug'] = true;

// Log IMAP conversation
$config['imap_debug'] = true;

// Log LDAP conversation
$config['ldap_debug'] = true;

// Log SMTP conversation
$config['smtp_debug'] = true;`

But i'm  unnable to view smtp file under logs folder

samuelmf:
Content of the smtp.log
```
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Connecting to mail.peihap.gob.pe:587...
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Recv: 220-opower.opowerserver.net ESMTP Exim 4.93 #2 Wed, 29 Jul 2020 21:55:58 -0500 
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Recv: 220-We do not authorize the use of this system to transport unsolicited, 
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Recv: 220 and/or bulk e-mail.
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Send: EHLO peihap.gob.pe
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Recv: 250-opower.opowerserver.net Hello peihap.gob.pe [66.85.141.178]
[30-Jul-2020 02:55:58 +0000]: <25ac829c> Recv: 250-SIZE 52428800
[...]

Issue #7494, opened 2020-07-18
Unable to reply to HTML Formatted emails received from Outlook Mail


tashecooke:
Working through the Browser console for errors I found a 403 error for the server response to a Roundcube POST Request.
In looking through Network > Preview /Response for further detail, I have found it to be a Firewall Response( Wordfence ) to a XSS in POST body  - here is the WordPress back end firewall message :

**Wordfence  Traffic Blocking** 

Christchurch, New Zealand
/roundcube/?_task=mail&_unlock…	7/17/2020 7:22:22 PM	122.59.155.125	...125-adsl.sparkbb.co.nz	403	
Type: Blocked
Activity Detail
 Christchurch, New Zealand left https://nzwebxpress.com/roundcube/?_task=mail&_action=compose&_id=5835812945f11519e47581 and was blocked by firewall for XSS: Cross Site Scripting in POST body: _message=%3Cp%3E%26nbsp%3B%3C%2Fp%3E%0D%0A%3Cdiv%20id%3D%22_rc_sig%22%3E%26nbsp%3B%3C%2Fdiv%3E%0D%0A%3Cp%3E%2… at https://nzwebxpress.com/roundcube/?_task=mail&_unlock=loading1594970543024&_framed=1&_lang=en_GB
7/17/2020 7:22:22 PM (56 minutes ago)  
IP: 122.59.155.125 Hostname: 122-59-155-125-adsl.sparkbb.co.nz
Human/Bot: Human
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36

What I am keen to understand is why this happens only to Outlook emails that I am replying to - as previously pointed out.

I would appreciate your insight before I exclude these parameters in Wordfence Filtering .

_Originally posted by @tashecooke in https://github.com/roundcube/roundcubemail/issues/7488#issuecomment-659963467_

alecpl:
We need to find out which part of the request precisselly is causing the problem. I can't do this for you. 

Use "Source code" button in the HTML editor toolbar to modify the HTML content. I propose to cut the code one chunk at a time and click the Save button every time. With this you should be able to narrow it down to the specific line(s).

tashecooke:
I have attached the Form Data for the POST that was  flagged as 403 by Wordfence as indicated above ,
[...]

Issue #7488, opened 2020-07-16
Unable to reply to HTML Formatted emails received from Outlook Mail Client


tashecooke:
I am running the latest version of Roundcube 1.4.7. on a Virtualmin Server install. 

I have now determined that when a user receives an email in Roundcube that is a HTML formatted email sent from a Outlook Mail Client and tries to reply - the sending message hangs for a few minutes and then errors with a "Requested Timed Out"msg. When this mail is converted to plain text you can reply without issues.

This does not happen for any HTML formatted mail received from Android Mail , Web Mail or GMail.

I have tried all three  User Interface Skins - all three have the same issue.

alecpl:
This might be a HTML5 parser issue. Could you provide a sample? What PHP version?

tashecooke:
Thanks for your response.
I am running PHP-FPM 7.4.8 Server.
For the sample , do wish to see the email Frame Source code?

alecpl:
I wish to see the full email message source.

tashecooke:
<!DOCTYPE html>
<html lang="en">

<head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8">
<title>Roundcube Webmail :: From Outlook Mail Client</title>
	<meta name="viewport" content="width=device-width, initial-scale=1.0, shrink-to-fit=no, maximum-scale=1.0">
<meta name="theme-color" content="#f4f4f4">
<meta name="msapplication-navbutton-color" content="#f4f4f4">

	<link rel="shortcut icon" href="skins/elastic/images/favicon.ico?s=1594702112">

	<link rel="stylesheet" href="skins/elastic/deps/bootstrap.min.css?s=1594702112">
			<link rel="stylesheet" href="skins/elastic/styles/styles.css?s=1594702112">
		
	<link rel="stylesheet" type="text/css" href="plugins/jqueryui/themes/elastic/jquery-ui.css?s=1594702112">
<script src="program/js/jquery.min.js?s=1594702112"></script>
<script src="program/js/common.min.js?s=1594702112"></script>
<script src="program/js/app.min.js?s=1594702112"></script>
<script>
/*
        @licstart  The following is the entire license notice for the 
        JavaScript code in this page.

        Copyright (C) The Roundcube Dev Team

        The JavaScript code in this page is free software: you can redistribute
        it and/or modify it under the terms of the GNU General Public License
        as published by the Free Software Foundation, either version 3 of
        the License, or (at your option) any later version.

[...]

PR #7959, opened 2021-03-20
handle http 500 error on message send

johndoh:
ref: #7494, #7488, #7522

in cases where the user tries to send a message but the request cannot be processed by Roundcude the UI remains locked on "sending message" until timeout.

if the request was successfully processed by Roundcube the UI is always unlocked and either the success message is shown or the error that Roundcube got from the server so if the UI is still locked after the iframe is loaded then something unexpected must have happened and in this case we can show a connection error to the user.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -4029,7 +4029,7 @@ function rcube_webmail()
             hidden = $("[name='_pgpmime']", form),
             msgid = ref.set_busy(true, draft || saveonly ? 'savingmessage' : 'sendingmessage');
 
-          form.target = ref.get_save_target();
+          form.target = ref.get_save_target(msgid);
           form._draft.value = draft ? '1' : '';
           form.action = ref.add_url(form.action, '_unlock', msgid);
           form.action = ref.add_url(form.action, '_framed', 1);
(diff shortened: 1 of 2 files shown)


program/js/app.js
```diff
@@ -5211,11 +5211,29 @@ function rcube_webmail()
...
     this.dummy_iframe('savetarget', 'about:blank')
-      .on('load error', function() { $(this).remove(); });
+      .on('load error', function() {
+        // catch invalid/error response from server and unlock the UI (#7494, #7488, #7522)
+        // a short delay is used to allow any valid response to be processed
+        setTimeout(function() {
+          if (ref.iframe_error_func)
+            ref.iframe_error_func();
+        }, 100);
+
```
alecpl, line 5234: I think we could make all of this much simpler. Can't we just do something like
```
if (e.type == 'error') { 
    ref.iframe_loaded(true);
    ref.display_message('connerror', 'error');
}
```

johndoh: I can't make the event `error` fire. I understand it should fire "if the iframe fails to load" but if the web server returns a 500 error then its still the `load` event which is fired, at least for me. There is some discussion here https://bugs.chromium.org/p/chromium/issues/detail?id=365457

I put the func into the `this.iframe_error_func` var for 2 reasons. 1) the function can be removed on successful load - even with the timer I thought it best to use a belt and braces approach. 2) while this is a non-breaking change, nothing changes unless the `unlock` is defined, i though using the var in this way made it more extensible.

alecpl: `$(this).contents().find('script').length` is 0 on error. Maybe we could use this. We can make it even more specific and assign `id` attribute to the script element, so we're 100% sure it's the Roundcube generated page content.

johndoh: I think I understand and I like your suggestion, it seems much more reliable that using a timer and checking the `busy` flag.

PR #7983, opened 2021-03-30
Redesigned dark-mode toggle

thomascube:
Make the taskbar button to switch dark/light mode an actual toggle element as it is widely seen in the web.

This is how the toggle looks like in different screen sizes:
dark-mode-toggle-dark
dark-mode-toggle-tablet
dark-mode-toggle-popover

### DIFF
--- a/skins/elastic/styles/dark.less
+++ b/skins/elastic/styles/dark.less
@@ -263,6 +263,24 @@ html.dark-mode {
         background-color: lighten(@color-dark-list-selected-background, 5%);
     }
 
+    .darkmode-toggle-label {
+        background-color: @color-taskmenu-button-selected-background;
+
+        &:before {
+            opacity: 1.0;
+        }
+    }
+
+    .darkmode-toggle-span {
+        background-color: @color-dark-font;
+    }
+
+    #layout-menu.popover {
+        .darkmode-toggle-label {
+            background-color: @color-taskmenu-button-selected-background;
+        }
+    }
+
     // ---------------------------------
     // Lists
     // ---------------------------------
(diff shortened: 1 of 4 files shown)


alecpl:
It looks strange without a label where all buttons have it.

johndoh:
For me I don't think the toggle needs to be on taskbar because I don't think its a setting you need to change very often. Except it might be useful on the mail preview/view screen if dark mode has messed up the message display. I think the primary dark mode toggle could be in Preferences > User Interface. Just my 2 cents.

thomascube:
@alecpl code size should not be an argument and the current implementation as a taskbar element is far from optimal because it has a totally different function than the other elements here.

@johndoh I basically agree that the taskbar isn't the right place but we lack a better position with appropriate visibility. Hiding it in the preferences might be a bit too discrete. Most websites offering such a toggle have it places in an always visible place as one might also switch spontaneously depending on lighting conditions.

thomascube:
> It looks strange without a label where all buttons have it.

It's not a button like the others, therefore it shall look differently.

thomascube:
What about a different placement top right?

roundcube-dark-toggle-tr

On smaller screens the toggle would just be hidden in the 3-dots menu.

alecpl:
Maybe better, although on screen width from 769px to ~850px it may take a precious place from the toolbar buttons (depending on the plugins enabled). That screen width is probably not very common, though.

> It's not a button like the others, therefore it shall look differently.

It's not that different. The fact that it does not display a dialog (About) or move to another page (any other button) does not mean it's out of place. The action is immediately visible so users should not be confused with it in any way. I don't see an issue in the lack of the selected state (it indicates the state with icon/label), that's why it's in the About/Logout section.

alecpl:
@thomascube I don't want to sound like a dictator, and I really don't like to throw away your time spend on this, but... I look at this once a week since you created the PR and I'm still not convinced it's any better than my approach.

PR #7992, opened 2021-04-03
allow Junk icon label to be localized

sdondley:
Currently, it's not possible to override the default "Junk" label for the Junk mail icon in the toolbar. With this change, the user can add in something like to `$labels['junk'] = 'Spam';` to a localization file to change the label.


johndoh:
dupe #7282

sdondley:
OK, looks like the proper way to address this issue is to modify the appropriate localization file in roundcube's core (e.g. `/usr/sare/roundcube/program/localization/en_US/labels.inc`).

PR #8002, opened 2021-04-14
Update Some Labels and Messages in Persian Language

seyyedhasan6172:
Some messages did not exist in messages.inc and some labels did not exist labels.inc for the Persian language in fa_IR folder.
Also, some labels have a better translation.
I added some that did not exist and edit some translations.


alecpl:
This has to be done via https://www.transifex.com/roundcube/roundcube-webmail/language/fa_IR/ otherwise it will be overwritten next time we update localizations.

PR #8023, opened 2021-04-23
Add exec_hook to new_user_dialog

JensSpanier:
Executes `identity_update` hook before updating identity.

My goal is to update the provided name in hMailServer. For this I'm listening for `identity_update` hook. But it doesn't get called. So I've added a manual exec_hook to plugin `new_user_dialog`.

I was inspired by plugin `new_user_identity`.

I'm not quite sure if handling `abort` this way is correct.

### DIFF
--- a/plugins/new_user_dialog/new_user_dialog.php
+++ b/plugins/new_user_dialog/new_user_dialog.php
@@ -172,10 +172,10 @@ rcube_webmail.prototype.new_user_dialog_close = function() { newuserdialog.dialo
         }
         else {
             // execute hook
-            $plugin = $rcmail->plugins->exec_hook('identity_update', array(
+            $plugin = $rcmail->plugins->exec_hook('identity_update', [
                 'id' => $identity['identity_id'],
                 'record' => $save_data
-            ));
+            ]);
 
             if (!$plugin['abort']) {
                 // save data
--- a/plugins/new_user_dialog/new_user_dialog.php
+++ b/plugins/new_user_dialog/new_user_dialog.php
@@ -176,13 +176,24 @@ rcube_webmail.prototype.new_user_dialog_close = function() { newuserdialog.dialo
                 'id' => $identity['identity_id'],
                 'record' => $save_data
             ));
-            if ($plugin['abort']) return;
-            // save data
-            $rcmail->user->update_identity($plugin['id'], $plugin['record']);
-            $rcmail->user->save_prefs(['newuserdialog' => null]);
-            // hide dialog
-            $rcmail->output->command('new_user_dialog_close');
-            $rcmail->output->show_message('successfullysaved', 'confirmation');
+
+            if (!$plugin['abort']) {
+                // save data
+                $updated = $rcmail->user->update_identity($plugin['id'], $plugin['record']);
+                $rcmail->user->save_prefs(['newuserdialog' => null]);
+            } else {
+                $updated = $plugin['result'];
+            }
+
+            if ($updated) {
+                // hide dialog
+                $rcmail->output->command('new_user_dialog_close');
+                $rcmail->output->show_message('successfullysaved', 'confirmation');
+            } else {
+                //show error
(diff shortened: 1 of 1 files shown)


alecpl:
This is incomplete. Please, take a look at how the hook is handled in program/actions/settings/identity_save.php. Specifically, 'abort', 'message', 'result' and 'record' properties.

Virinum:
Sorry, but I don't know what you want me to do. I had a look at `identity_save.php`.

**Email conversion:** Wasn't done before either
**Abort:** Is handled
**Result:** I don't know what you mean
**Record:** I don't know what you mean
**Message:** Wasn't done before either

Like I said, I looked at the plugin `new_user_identity` and there it is handled nearly the same way:
https://github.com/roundcube/roundcubemail/blob/b4b50b87b35e5a50bd94666057d72ce716082dd7/plugins/new_user_identity/new_user_identity.php#L102-L114

alecpl:
This is how this hook is handled in the other place (the relevant parts that are wrong or missing here):
```php
    if (!$plugin['abort']) {
        $updated = $rcmail->user->update_identity($iid, $save_data);
    }
    else {
        $updated = $plugin['result'];
    }

    if ($updated) {
        $rcmail->output->show_message('successfullysaved', 'confirmation');
    }
    else {
        $error = !empty($plugin['message']) ? $plugin['message'] : 'errorsaving';
        $rcmail->output->show_message($error, 'error', null, false);
    }
```

Virinum:
It's no problem to include this. But it will add a success message, which it didn't had before. Is that wanted?

And am I right that it's also missing in `new_user_identity` plugin?

alecpl:
> It's no problem to include this. But it will add a success message, which it didn't had before. Is that wanted?

The success message is already there.
```
 $rcmail->output->show_message('successfullysaved', 'confirmation');
```
> And am I right that it's also missing in `new_user_identity` plugin?

Yes, but there it is not possible (or much harder) to make use of all hook arguments.

Virinum:
> The success message is already there.

You're right. I've overlooked this.

Hope it's now complete. Learned a lot about hooks. Thanks for your support and all your work!

plugins/new_user_dialog/new_user_dialog.php
```diff
@@ -171,12 +171,29 @@ function save_data()
...
         }
         else {
-            // save data
-            $rcmail->user->update_identity($identity['identity_id'], $save_data);
-            $rcmail->user->save_prefs(['newuserdialog' => null]);
-            // hide dialog
-            $rcmail->output->command('new_user_dialog_close');
-            $rcmail->output->show_message('successfullysaved', 'confirmation');
+            // execute hook
+            $plugin = $rcmail->plugins->exec_hook('identity_update', array(
```
alecpl, line 175: Please, use short array syntax.

plugins/new_user_dialog/new_user_dialog.php
```diff
@@ -171,12 +171,29 @@ function save_data()
...
+                'id' => $identity['identity_id'],
+                'record' => $save_data
+            ));
+
+            if (!$plugin['abort']) {
+                // save data
+                $updated = $rcmail->user->update_identity($plugin['id'], $plugin['record']);
+                $rcmail->user->save_prefs(['newuserdialog' => null]);
```
alecpl, line 183: I would move this save_prefs() call below, under `if ($updated)`.

plugins/new_user_dialog/new_user_dialog.php
```diff
@@ -171,12 +171,29 @@ function save_data()
...
+            } else {
+                $updated = $plugin['result'];
+            }
+
+            if ($updated) {
+                // hide dialog
+                $rcmail->output->command('new_user_dialog_close');
+                $rcmail->output->show_message('successfullysaved', 'confirmation');
+            } else {
+                //show error
```
alecpl, line 193: add a space after `//`

alecpl (changes requested):

PR #8025, opened 2021-04-23
hungarian translation fixes

smatyas:
I've fixed a couple of typos for the hungarian lagnuage.


alecpl:
This have to be done via https://www.transifex.com/roundcube/roundcube-webmail/language/hu_HU/

smatyas:
Yeah, I read the readme first, then registered on transiflex, and then created this MR. If you could accept my join request or ping someone in that team to accept my request, then I could try to add my changes on that platform too.

PR #8026, opened 2021-04-24
implement XOAUTH2 for managesieve

### LINKED ISSUE #7925: Feature request -- oauth/managesieve
Having gotten oauth to work via Keycloak, I've now discovered that when logged in that way it is no longer possible to manage vacation rules and other filters with the managesieve plugin.  Not a huge deal but would definitely be nice if that integration could survive into an oauth environment.

ibotty:
fixes #7925

This still needs a new release of Net_Sieve.

### DIFF
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -93,6 +93,7 @@ class rcmail_oauth
         if ($this->is_enabled()) {
             $this->rcmail->plugins->register_hook('storage_init', [$this, 'storage_init']);
             $this->rcmail->plugins->register_hook('smtp_connect', [$this, 'smtp_connect']);
+            $this->rcmail->plugins->register_hook('managesieve_connect', [$this, 'managesieve_connect']);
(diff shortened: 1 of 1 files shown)


program/actions/login/oauth.php
```diff
@@ -40,6 +40,7 @@ public function run($args = [])
             if ($auth && isset($auth['username'], $auth['authorization'], $auth['token'])) {
                 // enforce XOAUTH2 auth type
                 $rcmail->config->set('imap_auth_type', 'XOAUTH2');
+                $rcmail->config->set('managesieve_auth_type', 'XOAUTH2');
```
alecpl, line 43: This line is most likely not needed.

alecpl (changes requested):

ibotty:
I thought so too.  I thought it would fix an edge case, maybe like the case for smtp is doing.

ibotty:
It would be great if you could take another look.

Issue #8069, opened 2021-05-21
MDN sends Return-Receipt-To header, usually used for DSN


funnisquares:
I've noticed this recently, although Return-Receipt-To is not RFC compliant, it is still widely used, and that is even recognized in [RFC3464](tools.ietf.org/html/rfc3464). Is there a reason for this? This behavior also does not seem to be RFC compliant, since I've seen no mention of this header in [RFC8098](https://datatracker.ietf.org/doc/html/rfc8098).

alecpl:
Hmmm... I guess we could remove it. Thunderbird for example does not use this header. @thomascube ?

funnisquares:
If this is actually an issue, meaning that I am not mistaken in my assertions, this fixes it. In this patch I've made it so DSN sends the non-standard "Return-Receipt-To" header, since it is non-standard I'd understand If it's not desirable for it to be included.
```
Fixes MDN sending both DSN and MDN headers
--- a/program/include/rcmail_sendmail.php
+++ b/program/include/rcmail_sendmail.php
@@ -191,9 +191,11 @@ class rcmail_sendmail
         }

         if ($mdn_enabled) {
-            $headers['Return-Receipt-To']           = $from_string;
             $headers['Disposition-Notification-To'] = $from_string;
         }
+		if ($dsn_enabled){
+            $headers['Return-Receipt-To']           = $from_string;
+		}

         if (!empty($_POST['_priority'])) {
             $priority     = intval($_POST['_priority']);
```

alecpl:
It looks like some old Microsoft extension, but indeed sending it in MDN mode sounds wrong. I'll tag this ticket as a bug, and I think the proper fix will be to remove the header completely.

thomascube:
Can't remember where it came from, probably just looked at other email clients back then. I however agree to remove it.

alecpl:
Fixed via ed17ce58b11.

PR #8071, opened 2021-05-21
Change headers in DSN and MDN requests

dcfreire:
MDN sending "Return-Receipt-To" not only is not RFC compliant and can cause confusion in SMPT servers that expect this header for DSN requests. The first commit removes the "Return-Receipt-To" header, the second commit adds this header for DSN requests (The latter is not RFC compliant, but can make roundcube more compatible to exsisting SMTP solutions).

https://github.com/roundcube/roundcubemail/issues/8069


alecpl:
Fixed via ed17ce58b11.

Issue #5700, opened 2017-03-20
Add more fonts and text sizes to HTML formatting


ygoe:
I'd like to use the Segoe UI font in e-mails. This is the standard UI font in all Windows systems for the last ten years. As I've seen, helvetica/sans-serif are specified as fallback in the style attributes anyway, so this is a common and safe formatting. I do the same in my custom stationary for Thunderbird.

I'd also like to have finer control over the font size. It's a big leap from 8pt to 10pt. There could be 9pt as well to format signature text a bit smaller, but not too small to read. And as we're here, 7, 11 and 13pt might fit nicely, too. This is not a numeric spinbox with free number entry, it's only a pick-one list.

All in all this should be just some more entries to existing drop-down lists.

alecpl:
I don't like these ideas. E-mail agents should use only "safe" fonts. I don't think Segoe UI is one of them.
As for the font-size, I don't think we have to be so precise in font selection. Some mail clients do not even provide such size selection, they use predefined set of formats (mostly for headings). Maybe we should also think about simplifying the editor toolbar even more in the future. The more HTML features you use the more chances you have to break on a receiving client.

ygoe:
I don't see how Andale Mono is a safe font. What does that even mean? Most of the fonts listed are only available on Windows, and Segoe UI is certainly available there, too. If you want it safe, don't specify fonts at all or provide them as web font (which MUAs don't support).

Can I enter the raw HTML for my signature? If you want signatures with anything like a certain style you need fine control over the formatting. I do test what works and what doesn't in other MUAs like Outlook or web mail services. And what I want works just fine in most places and doesn't break in others. Roundcube is clearly the over-simplifying limiting factor for me now.

alecpl:
Of course you can use custom HTML code. See "Source code" button in the editor toolbar.

alecpl:
Maybe we should review our current fonts set (especially for when we implement a new skin with mobile devices support) with help of http://www.cssfontstack.com/. As you see Segoe UI has only 75% coverage on Windows and 0% on Mac (and Linux/Android I guess).

ygoe:
Thanks, I didn't know the source code button. The editor seems to filter the source code anyway so that doesn't help much. The editor really seems to be the limiting factor here. And since the font in the signature is replaced with the main mail font (if nothing is specified) that doesn't help either.

[...]

PR #8078, opened 2021-05-27
Adding missing fonts compared to TinyMCE

tpayen:
TinyMCE support 'Comic Sans MS' and 'Symbol' fonts but they are not available in settings


alecpl:
This breaks some tests. And in relation to #5700 I would rather consider slimming down the list. We do not have to be on par with TinyMCE.

Issue #7381, opened 2020-05-10
Roundcube does not invalidate session when credentials change during active session


GermanCoding:
Quick setup overview:

- I'm using a database for authentication
- Dovecot is configured to authenticate all IMAP logins against that database
- Roundcube connects to the Dovecot IMAP server and passes through credentials
- There is an external authentication program that may change credentials, by modifying the database

Steps to reproduce:

1. Login to Roundcube, using valid credentials.
2. On the external authentication program, change the credentials of the user that just logged in.
    (This causes dovecot to also drop any currently established IMAP connections for that user)
3. As soon as the user tries to interact with anything within Roundcube, Roundcube tries to rebuild the IMAP session, but fails because dovecot will now reject the previously good credentials.
4. Roundcube now shows two errors messages ("connection to storage server failed" and "Server Error: AUTHENTICATE PLAIN: Authentication failed") but does **not** invalidate the user session - they can still roam around the site, which causes roundcube to hammer dovecot with invalid logins - which seems to  bypass even the `login_rate_limit`, potentially causing dozens (or even hundreds) of failed logins if the user refreshes their browser window a couple of times.
5. Roundcube happily reports all these "failed" logins, eventually causing my Fail2Ban to kick in and ban the (pretty much) innocent user

What Roundcube should instead do, is to invalidate the currently active user session as soon as the IMAP server reports "Authentication failed" and send them back to the login page.

Right now, the user needs to manually press "Logout" on their broken session in order to get Roundcube working again.

I've attached a screenshot that shows the described state. Again, this state persists and the user will not get logged out automatically.

![roundcube_failure](https://user-images.githubusercontent.com/4279661/81499249-62642200-92ca-11ea-8c25-8dbf3d806faf.png)

As for logging, roundcube logs the failed logins like this (frequently):

`[10-May-2020 14:28:21 +0200]: <m04im4ah> IMAP Error: Login failed for [user] against [server] from [IP] . AUTHENTICATE PLAIN: Authentication failed. in /var/www/vhosts/mail/httpdocs/program/lib/Roundcube/rcube_imap.php on line 200 (POST /?_task=mail&_action=refresh)`

PS: I'm running Roundcube Webmail 1.4.4

johndoh:
dupe #6474

GermanCoding:
As I guess this is a `wontfix` then, I've implemented a workaround that works for me. I'm posting the code snippet here in case others find it useful:

[...]

Issue #6474, opened 2018-10-03
fail2ban: Stop refreshing when login fails


vpelletier:
I have a fail2ban setup to watch roundcube log for authentication failures (in addition to built-in login rate limiter), and users can change their IMAP password without having to go through roundcube. When they do and they have roundcube open in a tab, they promptly get detected as a bruteforce attempt and banned.

I think auto-refresh should stop and display an error (maybe send user to login form ? not sure how it would play if user is composing an email right at that time) on the first authentication error.

alecpl:
I agree we should display an error. As the imap authentication failure may be temporal and we do not know it's caused by a new password, I would not logout the user.

alecpl:
Done.

vpelletier:
I am not sure the issue if completely resolved: while displaying the error is an improvement, the main issue for me is that users trigger fail2ban "without doing anything", just by having a tab opened somewhere when they change their password.

Unless I'm misreading the patch, or if there is another patch, it seems roundcube will keep refreshing even if there is an error. This may require distinguishing between "credentials received and rejected" and other errors ("browser-server network error", "server internal error" like "couldn't reach auth DB", ...), which hopefully should be possible (I don't know IMAP, but hopefully they do have separate error codes, similar to HTTP 403).

Now that the error became visible, I think the user can more easily understand they are supposed to do something to get a fresh mailbox view.

alecpl:
We could catch authentication errors, but even these can be caused by different factors. We don't know if it's wrong credentials or some temporary failure on authentication DB. Even if server specifically says "wrong credentials, password changed" there's no standard for this. Correct me if I'm wrong. And I don't think we should stop refreshing on every auth error.

thomas-mc-work:
> We could catch authentication errors, but even these can be caused by different factors. We don't know if it's wrong credentials or some temporary failure on authentication DB. Even if server specifically says "wrong credentials, password changed" there's no standard for this.

From my understanding we are talking about two different situations. First: authentication fails for any technical reasons (as you've stated, this could be e.g. a temporary failure on authentication DB). Second: proper authentication failures due to invalid credentials.

My question is here: Can these two situation be distinguished properly by the IMAP status code? I only know HTTP. There the situation is very clear: You receive a `401` if and only if the credentials are invalid. Any other server side errors return a `5xx` status code, client side errors a `4xx`.

[...]

PR #8089, opened 2021-06-06
invalidate session when credentials change during active session

thomas-mc-work:
This is an approach to solve #7381. I made it optional, so it's an opt in and won't bother users unexpectedly.

This is just a first approach. I think I need some assistance to get it working properly.

Current issues with that solution:

1. Is that error code precise enough to detect only authentication errors?
2. The user some doesn't get logged out immediately, but a bit delayed. I don't know how to trigger the frontend to process the action immediately.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -252,6 +252,10 @@ $config['messages_cache_ttl'] = '10d';
 // Note: On MySQL this should be less than (max_allowed_packet - 30%)
 $config['messages_cache_threshold'] = 50;
 
+// Trigger a logout when the IMAP authentication fails in an active session. This can happen, when the user changes it's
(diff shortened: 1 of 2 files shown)


alecpl:
I don't like it, as I explained in #6474.

thomas-mc-work:
I've replied in your mentioned issue. I'd appreciate your thoughts about that. Thank you!

thomas-mc-work:
Seems impossible in a completely sane way:

> There's no way to distinguish these two cases on the IMAP protocol level.

The two cases are:

1. authentication fails for any technical reasons
2. proper authentication failures due to invalid credentials

Finally: It's up to the admin to decide whether case 1 is really an issue. Sadly the patch has to be applied then manually.

PR #8096, opened 2021-06-10
Fix parsing aud jwt claim

### LINKED ISSUE #8088: JWT parser doesn't support aud claim array
The JWT parser currently verifies the `aud` claim using a direct comparison with the client ID:
https://github.com/roundcube/roundcubemail/blob/f7ba562bed77a3fbdf70308e345f34c8f03f1cff/program/include/rcmail_oauth.php#L151-L153

However, according to [RFC7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3), the `aud` claim is an array of case-sensitive strings, which _may_ be shortened to just a case-sensitive string (without the wrapping array) if there is only one.

The current code fails in the default case, since it doesn't take arrays into account. Instead, the code should check whether the client ID is one of the strings in the array.

invliD:
Fixes #8088.

### DIFF
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -148,7 +148,7 @@ class rcmail_oauth
         if (isset($body['azp']) && $body['azp'] !== $this->options['client_id']) {
             throw new RuntimeException('Failed to validate JWT: invalid azp value');
         }
-        else if (isset($body['aud']) && $body['aud'] !== $this->options['client_id']) {
+        else if (isset($body['aud']) && (is_array($body['aud']) ? !in_array($this->options['client_id'], $body['aud']) : $body['aud'] !== $this->options['client_id'])) {
             throw new RuntimeException('Failed to validate JWT: invalid aud value');
         }
         else if (!isset($body['azp']) && !isset($body['aud'])) {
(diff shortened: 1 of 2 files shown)


program/include/rcmail_oauth.php
```diff
@@ -148,7 +148,7 @@ public function jwt_decode($jwt)
         if (isset($body['azp']) && $body['azp'] !== $this->options['client_id']) {
             throw new RuntimeException('Failed to validate JWT: invalid azp value');
         }
-        else if (isset($body['aud']) && $body['aud'] !== $this->options['client_id']) {
+        else if (isset($body['aud']) && (is_array($body['aud']) ? !in_array($this->options['client_id'], $body['aud']) : $body['aud'] !== $this->options['client_id'])) {
```
alecpl, line 151: This line is quite long, I would probably do: `(isset($body['aud']) && !in_array($this->options['client_id'], (array) $body['aud']))`.

tests/Rcmail/Oauth.php
```diff
@@ -28,6 +28,34 @@ function test_jwt_decode()
...
     }
 
+    /**
+     * Test jwt_decode() method with an array aud
+     */
+    function test_jwt_decode_array()
+    {
+        $jwt = 'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImF1ZCI6WyJzb21lLWNsaWVudCJdfQ.signature';
+
+        $oauth = new rcmail_oauth(array(
```
alecpl, line 38: Short array syntax, please. Here and below.

alecpl (changes requested):
And you added tests, great.

PR #8102, opened 2021-06-21
[Feature] Localization: add greeting and forgot-password labels

ABendaravicius:
## Description
Using Rouncubemail as our webmail client we have found missing localization for a couple of labels that we believe could prove useful to many other users as well.

## Changes
Added **2 new** labels, that our team would continue to translate into multiple languages:
- `greeting` label: meant to be a less formal way of saying **Hello** in any available language.
- `forgotpassword` label: quite self-explanatory. As it is currently not an available label, would provide localization option for the **Forgot password** link, regardless of the means it has been implemented into the webmail.

### DIFF
--- a/program/localization/en_GB/labels.inc
+++ b/program/localization/en_GB/labels.inc
@@ -15,12 +15,14 @@
 */
 
 
+$labels['greeting'] = 'Hi';
 $labels['welcome'] = 'Welcome to $product';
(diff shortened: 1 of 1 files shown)


alecpl:
We're not accepting extra labels. Plugins and skins can have their own localization.

PR #8107, opened 2021-06-30
Fixe shift + drag & drop popover on chromium

Rotomeca:
On chromium based navigator, shift + drag & drop a mail doesn't show the popup, this correction correct this.


alecpl:
Thanks. I fixed this with non-ES6 syntax in 92624f16dc4b.

PR #8111, opened 2021-07-03
Always use TLS when requested and supported; fixes (#4590)

c---:
This fixes the long-standing issue where STARTTLS would only be issued if user
authorization was done. (#4590)

NOTE: This requires [Net_SMTP](https://github.com/pear/Net_SMTP) version 1.10.0 or newer


alecpl:
The XCLIENT support is new, so BC break is probably small, I guess, but it might be better to put this code after XCLIENT command is send.

c---:
I moved the STARTTLS to after XCLIENT. Seems to work fine in my tests though I don't have a server that supports XCLIENT.

PR #8115, opened 2021-07-05
Remove image Data URIs extraction into message attachments

facies:
Hello,

I want to suggest to remove this feature : https://github.com/roundcube/roundcubemail/blob/78af81ed87da63048176cb8d963e2b2d49aa4080/program/include/rcmail_sendmail.php#L347

As Data URIs are now well supported by a lot of email clients I think this feature can be removed. It is not useful anymore and it adds a lot of trouble for companies that are using base64 images in email signature. The conversations are quickly flooded with attached files from the signature.

Thank you for looking!


alecpl:
The PR is broken, you can't just do master -> release-1.4 request. And changes as these should go to master only.

Issue #8108, opened 2021-07-02
Very slow loading of data:image content into HTML editor


uli-heller:
I'm using version 1.4.11 of roundcube.

I have an email which I cannot forward or answer to.
Trying to do so produces something like this:

![image](https://user-images.githubusercontent.com/1677414/124223704-8739db00-db04-11eb-957d-79d26a114856.png)

I can provide more details upon request, but I don't want to upload the complete email here.

alecpl:
Check browser console for errors.

uli-heller:
I had to change a setting within roundcube to get the console for the "reply" (no new window):
![image](https://user-images.githubusercontent.com/1677414/124228237-3332f480-db0c-11eb-8898-8b4939f45b80.png)

Here is a screenshot of the message I'd like to reply to:
![image](https://user-images.githubusercontent.com/1677414/124228347-61183900-db0c-11eb-9688-f0db4ceeb01d.png)

And here the screenshot after clearing the console and pressing reply:
![image](https://user-images.githubusercontent.com/1677414/124228483-8dcc5080-db0c-11eb-96ce-9cbe124a27f5.png)

Please note: "To"(An) now contains a value - probably because I changed the roundcube settings?

alecpl:
Strange that there's no error, because the page looks exactly like when some js code wasn't executed, usually stopped by an error.

uli-heller:
Yes, but I didn't made it up. Now that you mention it, I sometimes "see" a screen like the one above for other mails, too. But it disappears pretty soon, probably due to javascript actions. I'll try again using firefox...

uli-heller:
Here are my observations:

- it looks pretty similar in firefox at first
- after a couple of seconds (10-30), the reply mail window changes and looks fine
- I tried again with chrome
- after 2 minutes, the window is still garbled up
- you cannot work within chrome any more, everything is extremely slow
- even closing the window takes time
- closing all roundcube windows/tabs makes chrome snappy again

So: Maybe some kind of javascript loop?

alecpl:
I see that there's quite a big inline image in the html content. Maybe this is the problem. Could you confirm it works with html email that have no such base64 images? You could also try 1.5-rc where we updated TinyMCE version.

uli-heller:
Hey @alecpl , I'll try do verify this. Will probably take a couple of hour since I'm @work at the moment...

uli-heller:
Here is what I did using firefox (FF) and chrome (C):

[...]

PR #8117, opened 2021-07-05
Don't extract image data URIs into message attachments

facies:
Hello,

As Data URIs are now well supported by a lot of email clients I think this feature can be removed. It is not useful anymore and it adds a lot of trouble for companies that are using base64 images in email signature. The conversations are quickly flooded with attached files from the signature.

Thank you for looking!


alecpl:
I hesitate to change this behavior without investigating as many as possible email clients, whether they support or not data URIs.

A quite good reason to not do this is #8108.

PR #8129, opened 2021-07-11
fix #8128

### LINKED ISSUE #8128: Firefox infinate loading display on mail screen
Revently I have noticed strange behavior when using Roundcube on my dev server. On the mail screen Firefox reports the page is loading forever (the X stop icon never changes to the Reload icon and the "Read [url]" message in the bottom left persists).
The browser console reports all resouces are loaded correctly. No JS errors, warnings:
```
Request to access cookie or storage on “[url]/program/resources/dummy.pdf” was blocked because we are blocking all third-party storage access requests and content blocking is enabled.
```

This issue does not effect the usability of the Roundcube interface.

The causes appears to be line 10134:
https://github.com/roundcube/roundcubemail/blob/2f643141b4e16dd70f0d1dda283c2106e63a3009/program/js/app.js#L10126-L10137

It seems like Firefox is getting confused by an element being added and removed from the DOM so quickly as adding a short delay before attempting to remove the dummy PDF object solves the issue, Firefox shows loading complete.

Tested with against 2 different webservers via HTTP and HTTPS connections.

Roundcube: git-master
Plugins: none
Skins: Elastic, Larry, Classic
Firefox 89.0.2
Windows 10

johndoh:
ref #8128

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -10131,7 +10131,10 @@ function rcube_webmail()
         })
         .on('load error', function(e) {
           ref.env.browser_capabilities.pdf = e.type == 'load' ? 1 : 0;
-          $(this).remove();
+
+          // add a short delay before attempting to remove element (#8128)
+          var obj = this;
+          window.setTimeout(function() { $(obj).remove(); }, 10);
         })
         .appendTo(document.body);
       }, 10);


program/js/app.js
```diff
@@ -10131,7 +10131,10 @@ function rcube_webmail()
         })
         .on('load error', function(e) {
           ref.env.browser_capabilities.pdf = e.type == 'load' ? 1 : 0;
-          $(this).remove();
+
+          // add a short delay before attempting to remove element (#)
```
alecpl, line 10135: Please, add the ticket reference.

PR #8132, opened 2021-07-13
Ensure potential phishing addresses are valid

justin-sleep:
Currently, Roundcube will mark emails with a "From" header name containing an "@" as potentially fraudulent. I've noticed this catches GitLab comment notification emails, which are marked as from the person and their username e.g. "Justin Sleep (@justin-sleep)".

I've modified this to use the `check_email` helper function instead.

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -1382,7 +1382,7 @@ class rcmail_action_mail_index extends rcmail_action
             $valid  = rcube_utils::check_email($mailto, false);
 
             // phishing email prevention (#1488981), e.g. "valid@email.addr <phishing@email.addr>"
-            if (!$show_email && $valid && $name && $name != $mailto && strpos($name, '@')) {
+            if (!$show_email && $valid && $name && $name != $mailto && rcube_utils::check_email($name, false)) {
                 $name = '';
                 self::$SUSPICIOUS_EMAIL = true;
             }


alecpl:
I would be careful with that. What if the name is not really a valid email address, but looks like it?

justin-sleep:
Hmm, I see what you're saying. My gut reaction is that if the attacker intentionally mangles the display name to evade detection, there isn't an effective way to prevent that. For example, the name "ceo＠google.com" is not currently caught, since it uses the full-width variant of @.

alecpl:
> Hmm, I see what you're saying. My gut reaction is that if the attacker intentionally mangles the display name to evade detection, there isn't an effective way to prevent that. For example, the name "ceo＠google.com" is not currently caught, since it uses the full-width variant of @.

Good catch. We should fix that.

alecpl:
I just received an email with `From: "Bugzilla@Mozilla" <bugzilla-daemon@mozilla.org>`. It makes the warning to appear. 

I'm wondering now whether the warning makes sense in this case. In such cases we do not show the name to the user, he sees only the email, and the email address itself is not fraudulent at all. So, it looks to me that in this case not displaying the name is enough and the warning should be displayed for cases when the email address is suspicious.

alecpl:
See 12ea080af8c5 and 02c7ddf9c1f. I'll close this pull request.

PR #8140, opened 2021-07-20
Update localization

jonmoesli:


alecpl:
This have to be done via https://www.transifex.com/roundcube/roundcube-webmail/

PR #8145, opened 2021-07-26
Fix dovecot passwdfile change

fribergr:
Currently there is a few problems with the dovecot_passwdfile password-driver.

- Trying to use two variables that are not initialized (`$handle` and `$newhash` - introduced in commit `ca523bd7facf29aaabf08718252314fca05c83a5`)
- If the `password::hash_password`-function returns false (which could happen when using `dovecot` as `password_algorithm`), the code would still proceed to change the password to an empty string even though hashing failed. It now raises an error instead.
- Escaping the hashed password result is not needed and actually breaks if option 'password_dovecotpw_with_method' is set to true resulting in an escaped entry in the passwd-file, i.e. `\{SSHA256\}...`

Let me know if something needs to be changed.

Thanks for developing an awesome mail client!


alecpl:
I think it will be better to log the error and stop execution from inside the hash_password() method. This way it will be working for any driver that uses it.

PR #8150, opened 2021-07-29
Enigma: Add the choice to select by default all or no identity when create a new key.

mickael067:
The change permit to make the choice to select by default all or no identity when create a new key.

Before the change, all identity are selected by default.

I have a lot of identity and deselect every identity to build one key on only one identity take time.

You are free to update the configuration variable name if you want.


alecpl:
Instead of a config option we should re-design the UI. If there's more that one (or three) email address(es) we could display "[x] All my addresses" only, when unchecked a list of addresses (with checkboxes unchecked) would appear. Or something like that.

mickael067:
A redesign of the UI can be a good additional solution, but need that someone work on it and wait time.
And the UI option doesn't remove the need to have  a configuration option to select if all selected or all not selected by default.
I propose to have this configuration option (already working) until someone update the UI to permit all  select or all un-select.
For information, I have more than 100 identity into the list.

PR #8173, opened 2021-08-20
replace Endroid\QrCode with BaconQrCode

johndoh:
Endroid\QrCode requires min PHP 7.3 but now so does master so we can update


alecpl:
There's one problem with this update. https://github.com/endroid/qr-code/issues/341

johndoh:
Ouch! I did not notice that. Perhaps we could switch to [bacon/bacon-qr-code](https://packagist.org/packages/bacon/bacon-qr-code) endroid requires it already any way and its much lighter weight. A very quick test suggests it should be a relatively easy switch.
```php
$renderer = new BaconQrCode\Renderer\ImageRenderer(
    new BaconQrCode\Renderer\RendererStyle\RendererStyle(300, 1),
    new BaconQrCode\Renderer\Image\ImagickImageBackEnd()
);
$writer = new BaconQrCode\Writer($renderer);
return $writer->writeString($data);
```

PR #8247, opened 2021-10-21
Replace crypt() with newer password_hash() wrapper

Github-Citizen:
PHP recommends use of password_hash() instead of crypt() as it automatically generates a stronger salt.


alecpl:
I'm closing this because we have password_hash() based options already in git-master, and 6cfca8dc9f5a.

PR #8270, opened 2021-10-28
Plugin to auto log out users with a POST request

ghost:
This plugin enables users  to auto log out with a POST request sent by submitting a form. This is in fact the counterpart of the 'autologon' plugin which is already part of Roundcube.

When Roundcube is integrated with other applications, a user that is logged in automatically by using the 'autologon' plugin cannot be logged out automatically, unless a plugin like this is also present in Roundcube. This plugin is really needed.


plugins/autologout/composer.json
```diff
@@ -0,0 +1,23 @@
...
+    "description": "Plugin to auto log out users with a POST request sent by submitting a form.",
+    "license": "GPLv3+",
+    "version": "1.0",
+    "authors": [
+        {
+            "name": "Cover Tower LLC",
+            "email": "contact@covertower.com"
+        }
+    ],
+    "repositories": [
```
alecpl, line 13: "repositories" is not needed here. I think we should remove it also from the other core plugins.

plugins/autologout/composer.json
```diff
@@ -0,0 +1,23 @@
...
+        {
+            "type": "composer",
+            "url": "https://plugins.roundcube.net"
+        }
+    ],
+    "require": {
+        "php": ">=5.3.0",
```
alecpl, line 20: use 7.3.0 here for consistency with other plugins and core, even though the plugin itself does not require it.

plugins/autologout/composer.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "roundcube/autologout",
+    "type": "roundcube-plugin",
+    "description": "Plugin to auto log out users with a POST request sent by submitting a form.",
```
alecpl, line 4: "sent by submitting a form" -> "sent from an external site" would be better.

plugins/autologout/autologout.php
```diff
@@ -0,0 +1,57 @@
...
+ * <input type="hidden" name="_action" value="logout" />
+ * <input type="hidden" name="_task" value="logout" />
+ * <input type="hidden" name="_autologout" value="1" />
+ * <input id="loSubmitButton" type="submit" value="Logout" />
+ * </form>
+ */
+
+class autologout extends rcube_plugin
+{
+  public $task = 'logout';
```
alecpl, line 24: Please, indentation with 4 spaces in the whole file.

plugins/autologout/autologout.php
```diff
@@ -0,0 +1,57 @@
...
+
+  function init()
+  {
+    $this->add_hook('startup', array($this, 'startup'));
```
alecpl, line 28: Short array syntax, please.

plugins/autologout/autologout.php
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  function startup($args)
+  {
+    $rcmail = rcmail::get_instance();
+
+    // Change task and action to logout
+    if (!empty($_SESSION['user_id']) && !empty($_POST['_autologout']) && $this->known_client())
+    {
```
alecpl, line 37: `{` on `if` statement should be in the same line.

alecpl (changes requested):

CoverTowerSuperAdmin:
I've made the changes. In the comment at the top of `autologout.php` I also explained that the POST request should be made only by submitting a form (which can be hidden and sent automatically using JQuery) and not by using CURL or other methods. Roundcube's security features prevent such POST requests and users can waste a lot of time trying to send the log out POST request with different methods other than using a form like the one given as example in the comment.

alecpl:
That's right. It has to be posted using the browser that has the Roundcube user session cookies.

plugins/autologout/autologout.php
```diff
@@ -0,0 +1,60 @@
...
+        return $args;
+    }
+
+    function known_client()
+    {
+        /**
+         *  If you want to restrict the use of this plugin to specific
+         *  remote clients, you can verify the remote client's IP like this:
+         *
+         *  if (in_array(rcube_utils::remote_addr(), ['123.123.123.123', '124.124.124.124'])) { return true; }
```
alecpl, line 55: Well, you should mention that in this case the `return true;` at the end needs to be switched to `return false`.

I'd just suggest doing
```
return in_array(rcube_utils::remote_addr(), ['123.123.123.123', '124.124.124.124']);
```
instead of an `if` statement.

CoverTowerSuperAdmin: I've replaced the if statement.

PR #8291, opened 2021-11-05
Mailcow driver for password plugin

Hexaris:


navossoc:
Since 1.6 is still in RC, can we make some changes here?
```php
$host  = $rcmail->config->get('password_mailcow_api_host');
// ...
$response = $client->post('https://' . $host . '/api/v1/edit/mailbox', [
    'headers' => $headers,
    'json'    => $cowdata
]);

```

This hardcoded host line is really bad.
Since mailcow runs on docker most of the users would probably set the config as:

```php
$config['password_driver'] = 'mailcow';
$config['password_mailcow_api_host'] = 'nginx';
$config['password_mailcow_api_token'] = 'my-super-token';
```

In your code, it would become:
'https://nginx/api/v1/edit/mailbox'

This is "OK", but nginx is a invalid server name for the certificate, so I will need to access it using my public server_name...

```php
$config['password_mailcow_api_host'] = 'mydomain.com';
```

So it would become:
'https://mydomain.com/api/v1/edit/mailbox'

This works fine, but it don't follow the same convention as `imap_host` and `smtp_host`.
Like we do with IMAP/SIEVE `tls://dovecot` and SMTP `tls://postfix` ([reference](https://mailcow.github.io/mailcow-dockerized-docs/third_party/third_party-roundcube/#installing-roundcube))

IMHO, this is better:
```php
$response = $client->post($host . '/api/v1/edit/mailbox', [
```

So I can change the schema as I want...
```php
$config['password_mailcow_api_host'] = 'http://nginx:8080';
```

I'm using a reverse proxy in front of nginx on mailcow, so this matches my setup perfectly.

If you really want go further, maybe something like `smtp_conn_options` so we can skip the certificate validation, but I think it's a little bit overkill.

[]'s

alecpl:
You could do `$config['password_http_client'] = ['verify' => false];`, I guess. But I modified the code to accept url in f8301a44b4a36605.

PR #8292, opened 2021-11-06
Update composer.json file to show php-intl is required

isleshocky77:
As of roundcube 1.5, `php-intl` is a required php extension. `composer` should show it and warn as such.

### DIFF
--- a/composer.json-dist
+++ b/composer.json-dist
@@ -10,6 +10,7 @@
     ],
     "require": {
         "php": ">=7.3.0",
+        "ext-intl": "*",
         "pear/pear-core-minimal": "~1.10.1",
         "pear/auth_sasl": "~1.1.0",
         "pear/mail_mime": "~1.10.0",


alecpl:
We require a lot of extensions, so we should list them all or none. I prefer none, because you can build the package on a different system.

isleshocky77:
@alecpl 

> We require a lot of extensions, so we should list them all or none. I prefer none, because you can build the package on a different system.

I personally prefer to see all of them listed. It's easier to tell `composer` to  `--ignore-platform-reqs` then it is to debug why something isn't working just because you're missing a dependency you didn't know about.

I'm fine updating this ticket off of the required and recommended extensions I just tracked down in the [INSTALL](https://github.com/roundcube/roundcubemail/blob/master/INSTALL) document. Would that be something you would accept as a PR? Or are you against including the information?

PR #8313, opened 2021-11-14
random_bytes -> random_string

twekkel:
For PHP version < 7 the custom roundcube implementation for random_bytes made sense, but now not anymore IMHO. Actually only once it returns bytes and in all other cases a string.

Suggest to clean up the code and use the native random_bytes from PHP 7+ and rename random_bytes to random_string.


alecpl:
I would not remove the old method because it will be a BC break for plugins.

twekkel:
Thanks for the review. Didn't think of that... will withdraw.

Issue #8324, opened 2021-11-21
chpass-wrapper.py helper script for password plugin appears to be broken with python 3.9


sdondley:
Can someone more knowledgeable with python please take a look at this:

https://stackoverflow.com/a/70051629/1641112

I'd be happy to supply a commit to patch this but I read about 2 chapters of a python book 10 years ago and haven't touched it since. If someone can confirm this a good fix, I'll do a pull request. Thanks.

alecpl:
Fixed.

The `text` argument does not work with Python < 3.7, so we have to use an alias `universal_newlines`. I tested the change on python 2.7 and 3.6.

PR #8325, opened 2021-11-21
fix(chpasswd): as reported on StackOverflow

faust64:
See https://stackoverflow.com/questions/70051230/php-error-when-trying-to-use-password-reset-plugin-in-roundcube

Newer python versions may fail to change user passwords.
Fix from StevieD -- https://stackoverflow.com/users/1641112/stevied


alecpl:
See #8324.

PR #8327, opened 2021-11-21
Faster random string generation

twekkel:
Calling  (pseudo-random) integers with random_int for longer keys, is kind of slow... because of the loop for each character.

Requesting multiple random bytes in one go, is quicker (~10x for 32 bytes). Random_bytes is "true random", not pseudo.

The conversion from base64 to base62 can cause a too short string in <1% of the cases, hence the loop. Benchmark code attached...
[r.php.gz](https://github.com/roundcube/roundcubemail/files/7576461/r.php.gz)

### DIFF
--- a/program/lib/Roundcube/rcube_utils.php
+++ b/program/lib/Roundcube/rcube_utils.php
@@ -1417,15 +1417,12 @@ class rcube_utils
             return random_bytes($length);
         }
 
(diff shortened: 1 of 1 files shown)


alecpl:
I don't really like it:
1. We're not using this method in a performance sensitive context, so it really does not matter.
2. Your way is less secure and utilizes more bytes than needed.

Issue #8309, opened 2021-11-11
"Dark Mode" makes scrollbars too narrow


trueshanti:
Switching from bright- to darkmode lets horizontal scrollbars get slimmer (with) .. this inflicts problems targeting scrollbars with pointerdevices , especially when using screensetting for handicapped people (like HUGE cursor). even using touchscreen is harder

if the sliders wouldnt loose width and also the up/down-arrow of the scrollbar wouldnt vanish in darkmode - it would all be fine

thanks for this great software

rc version is 1.5.0
i can reproduce this in latest firefox and chromium

trueshanti:
this happens up to the viewpoint that disables the icon-titles on the leftside funtionbar (tested by zooming in) then the browser switches back to unthemed scrollbars - which is good again and not so ugly anyway ;-)

alecpl:
I like these scrollbars and I'm not sure what to do about it. Maybe we should have some accessibility setting, but I'd hesitate to add it only for this single purpose.

trueshanti:
Hi, so you can see the difference ? its  approx 2 standard-pixels , thats a huge  difference for a single interface-element. If one sees it as "accessible"-related, so might it be .. 

But I am not required to use accessible-interfaces, and even I find it already making the interface more difficult to use .. 

So people with optical handycaps will have even more issues (as reported)

https://imgur.com/a/1Am6kT3

alecpl:
In master we use slim scrollbars in both dark and light modes. You can change the scrollbar width and rebuilt the skin now. We'll not change the default.

manfredu:
Am 2022-02-04 08:53, schrieb Aleksander Machniak:

> In master we use slim scrollbars in both dark and light modes. You can 
> change the scrollbar width and rebuilt the skin now. We'll not change 
> the default.

It is very difficult to use the small scrollbars in dark mode for 
scrolling. When clicking on the bar, in most cases the function to 
increase the column width gets activated first.

So the only option left for scrolling is the scroll wheel of the mouse, 
but clicking on the bar and dragging it, is nearly impossible.

trueshanti:
exactly the issue ..
@alecpl : i would like to know where i could put changes in css/less to get rid of exactly this .. like setting the same values in darkmode as they are in lightmode  - can you please point me to ?

i am no css-guru :-o

alecpl:
Create `skins/elastic/styles/_variables.less` file and there you can put this:
```
[...]

PR #8352, opened 2021-12-04
Elastic: move scrollbar settings to variables

johndoh:
possible solution for #8309 rather than adding any new settings the `scrollbar-width` CSS value could be moved to a Less variable `@scrollbar-dark-width` and then it can be overridden/customized like any other for specific requirements.

while I was there I added scrollbar styles into light mode so they can be customized in the same way. Tested in Firefox and Edge.

### DIFF
--- a/skins/elastic/styles/widgets/common.less
+++ b/skins/elastic/styles/widgets/common.less
@@ -239,6 +239,20 @@ fieldset.image-attachment {
     bottom: @layout-header-height;
 }
 
+textarea,
+select,
+.popover-body,
+.popupmenu,
+.ui-dialog-content,
+.frame-content,
+.formcontent,
+.table-responsive,
+.table-responsive-sm,
+.scroller {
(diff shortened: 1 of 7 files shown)


skins/elastic/styles/widgets/common.less
```diff
@@ -239,6 +239,42 @@ fieldset.image-attachment {
     bottom: @layout-header-height;
 }
 
+html {
+    scrollbar-color: @color-scrollbar-thumb @color-scrollbar-track;
+    scrollbar-width: @scrollbar-width;
```
alecpl, line 244: This shold be better placed in global.less.

johndoh: I put this block here because `.ios-scroll` and `.webkit-scroller` styles where already here.

johndoh: Is this more like it?

PR #8402, opened 2022-01-17
Some minor improvements for password plugin

matwey:
* Allow LDAP URI to communicate via UNIX socket (when available)
* Allow SASL bind for LDAP search queue.

### DIFF
--- a/plugins/password/config.inc.php.dist
+++ b/plugins/password/config.inc.php.dist
@@ -182,6 +182,29 @@ $config['password_ldap_starttls'] = false;
 // Default: '3'
 $config['password_ldap_version'] = '3';
 
+// LDAP search bind type
+// This option works only with 'ldap_simple' driver.
+// 'sasl' bind type is particularly useful when local UNIX socket is used to connect
+// the server, or client TLS certificate validation is enabled. In this cases we
+// can avoid using explicit DN.
+// Example: 'sasl'
+// Default: 'bind'
+$config['password_ldap_bind_method'] = 'bind';
+
+// LDAP SASL mech
+// SASL mech when 'password_ldap_bind_method' is 'sasl'.
+// This option works only with 'ldap_simple' driver.
+// Example: 'EXTERNAL'
+// Default: null
+$config['password_ldap_mech'] = null;
+
+// LDAP SASL realm
+// SASL realm when 'password_ldap_bind_method' is 'sasl'.
+// This option works only with 'ldap_simple' driver.
+// Example: 'myrealm'
+// Default: null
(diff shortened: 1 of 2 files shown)


alecpl:
Why sasl bind for search and not also for record modification? It would be good to describe new options in config.inc.php.dist. And it looks to me that 'password_ldap_uri' does not work in the 'ldap' driver. This needs to be fixed or noted.

matwey:
Thank you for the feedback!

- I think it is a good idea to enable SASL also for password update.
- `config.inc.php.dist` update is also on the way.
- Indeed, `ldap` driver uses deprecated library which doesn't support LDAP URI.

plugins/password/config.inc.php.dist
```diff
@@ -171,6 +171,13 @@ $config['password_ldap_host'] = 'localhost';
 // Default: '389'
 $config['password_ldap_port'] = '389';
 
+// You can also configure LDAP connection using URI.
+// This options works only with 'ldap_simple' driver and overrides
+// 'password_ldap_host' and 'password_ldap_port'.
+// Example: 'ldapi:///'
+// Default: null
+$config['password_ldap_uri'] = null;
```
alecpl, line 179: Many other options e.g. `imap_host` support URI syntax, therefore I think it would make sense to support URIs in `password_ldap_host` instead.

plugins/password/drivers/ldap_simple.php
```diff
@@ -222,10 +243,28 @@ function search_userdn($rcmail, $ds)
 
         $this->_debug("C: Bind " . ($search_user ? $search_user : '[anonymous]'));
 
-        // Bind
-        if (!ldap_bind($ds, $search_user, $search_pass)) {
-            $this->_debug("S: ".ldap_error($ds));
-            return false;
+        switch ($rcmail->config->get('password_ldap_search_bind_method')) {
```
alecpl, line 246: Do we really need a separate method/mech/realm for search? Is it a real use-case to have a different method for both binds?

alecpl (changes requested):

matwey:
Sorry for this long delay. I am working on changes.

plugins/password/config.inc.php.dist
matwey: `password_ldap_host` is used instead of `password_ldap_uri`

plugins/password/drivers/ldap_simple.php
matwey: `password_ldap_search_`/`password_ldap_admin_` renamed to `password_ldap_`

PR #8419, opened 2022-01-25
Add utils script to initialize or update database

thomascube:
This can be used by the Docker images where we don't know whether the configured database is already initialized.


alecpl:
It's strange to have initdb.sh, updatedb.sh and the third one. Instead I'd probably add an optional argument to initdb.sh to execute update if the database is already initialized.

PR #8431, opened 2022-01-31
Fix session change time not updating for certain timezones

### LINKED ISSUE #8303: session times out
Hi,
I am not sure if it is a bug, a configuration or related to the used openSUSE package, but I have the bahaviour since the update to roundcubemail 1.5

Before (1.4.x) I had a tab (FF) open with roundcubemail and it stayed open and in the tab I could see if a new mail arrived.

Now it loggs out after some minutes and I have to log in twice. The first try ends with a yellow warning bottom right "invalid entry. No data will be saved". Then the second try is ok.

(PHP 7.4.6)

carlreinke:
Use of `time_diff` was inverted, so if `time_diff` was negative, the session would almost certainly expire before `changed` would be updated.

Fixes  #8303.

### DIFF
--- a/program/lib/Roundcube/session/db.php
+++ b/program/lib/Roundcube/session/db.php
@@ -174,7 +174,7 @@ class rcube_session_db extends rcube_session
     public function update($key, $newvars, $oldvars)
     {
         $now = $this->db->now();
-        $ts  = microtime(true);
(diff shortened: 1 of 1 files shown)


alecpl:
Thanks. I pushed my version of the fix in 2c0b870bd693d.

Issue #8444, opened 2022-02-10
Encoding problem for RTF content in TNEF (winmail.dat)


jkacena:
Hello,
the encoding of message.rtf from winmail.dat seems to be broken.
It is the issue with Outlook generated messages written in the Czech language and it started to appear when we switched from 1.4.10 to 1.5.2. 

I believe it is related to this changelog entry in 1.5.0:
- Support displaying RTF content (including encapsulated HTML) from a TNEF attachment 

The e-mail itself has two mime parts. The first is  text/plain; charset="iso-8859-2" and the second is winmail.dat attachment.
I think the reason why 1.4.10 is ok is because it shows the e-mail body from the first mime (plain text only) and here the charset decodig is done correctly. While 1.5.2 shows rtf body and here the decoding is not ok.

1.5.2 with broken czech:
![obrazek](https://user-images.githubusercontent.com/82454538/153404321-42fbc72f-caf4-4060-add2-16bd8b6cd968.png)

1.4.10 is ok:
![obrazek](https://user-images.githubusercontent.com/82454538/153404451-be8e0a05-94a3-47e2-8824-95cabf69c2d0.png)

I'm not an expert on TNEF encapsulation, but when I extracted the winmail.dat, the first line of message.rtf looks like this:
{\rtf1\ansi\ansicpg1250\fromhtml1 \fbidis \deff0{\fonttbl
where cpg1250 is probably code page 1250 (the Czech encoding used in windows environments)

Note: I run both RC versions in parallel on the same server, with same PHP (7.3 with mbstring and iconv). The imap server is the same and I'm looking at the exact same message on the imap

The e-mail message itself:
[email.txt](https://github.com/roundcube/roundcubemail/files/8040784/email.txt)
.

alecpl:
You can disable TNEF processing with `tnef_decode=false`.

alecpl:
Fixed.

PR #8446, opened 2022-02-10
Update rcube_tnef_decoder.php

jkacena:
my quick fix for #8444


alecpl:
Proper fix in 6303d3177e0384.

PR #8464, opened 2022-03-02
Don't declare default statement for postgres columns in initial schem…

interduo:
…a - it's not needed as db engine do the same if it's not set


interduo:
https://www.postgresql.org/docs/14/ddl-default.html

`If no default value is declared explicitly, the default value is the null value. This usually makes sense because a null value can be considered to represent unknown data.`

PR #8477, opened 2022-03-12
Allow to set HTML Content-Transfer-Encoding based on UseMIMEfor 8bit

bojovyletoun:
[Original name "Allow to set HTML Content-Transfer-Encoding  by  "Use MIME for 8bit config setting]

I modified Content-Transfer-Encoding for HTML parts.   Previously it was hardcoded quoted-printable. Now it  is determined in same manner as for Text part.

**Is there any caveat in it?** I have Settings-Composing-Advanced:Use MIME for 8bit obviously off so it is effectively 8bit.

Allows to set Content-Transfer-Encoding for HTML part in similiar way as for Text part. (Based on Settings-Writing-Use MIME for 8bit)
If it is intentional (bugs in user agents...), please explain. According to my observations nearly everyone uses it and I think I should not be problematic, but I am not sure totally.

### DIFF
--- a/program/include/rcmail_sendmail.php
+++ b/program/include/rcmail_sendmail.php
@@ -279,7 +279,7 @@ class rcmail_sendmail
 
         // encoding settings for mail composing
         $message->setParam('text_encoding', $transfer_encoding);
-        $message->setParam('html_encoding', 'quoted-printable');
+        $message->setParam('html_encoding', $transfer_encoding);
         $message->setParam('head_encoding', $head_encoding);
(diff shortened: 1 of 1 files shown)


alecpl:
A slightly modified version pushed.

Windkracht8:
Hi, I've had to revert this in my installation because Mail_mime does not wrap lines when html_encoding is not quoted-printable or base64. My Exim server rejects emails where the lines are too long, they should not be over 998 octets.
8bit is not necessarily an issue, it just means Mail_mime won't wrap the lines and this needs to be done somewhere else.

PR #8479, opened 2022-03-17
Add ssha256 password algo

### LINKED ISSUE #8459: Support for SSHA256
[Mailcow](https://github.com/mailcow/mailcow-dockerized) (a mailserver-suite based on postfix, dovecot, rspamd) recommends in the [documentation](https://mailcow.github.io/mailcow-dockerized-docs/third_party/third_party-roundcube/#enable-change-password-function-in-roundcube) to patch [`password.php`](https://github.com/roundcube/roundcubemail/blob/master/plugins/password/password.php) manually to make `ssha256` work. Of course, when Roundcube is updated, this can break. I was wondering if it would make sense to have this directly in Roundcube so that manual patching is no longer necessary.

I am not familiar with encryption standards in detail, but ssha256 seems to be one standard among many.

ghost:
Suggestion to resolve https://github.com/roundcube/roundcubemail/issues/8459

This is my first pull request, i am thankful for hints or improvements

* tested with your workflows
* tested on my own mailcow installation

### DIFF
--- a/plugins/password/password.php
+++ b/plugins/password/password.php
@@ -636,6 +636,23 @@ class password extends rcube_plugin
             $prefix  = '{SSHA}';
             break;
 
+        // base64 encoded ssha256 for mailcow
+        case 'ssha256':
+            $salt = rcube_utils::random_bytes(8);
+
+            if (function_exists('hash')) {
+                $crypted = base64_encode( hash('sha256', $password . $salt, TRUE ) . $salt );
+                $prefix  = '{SSHA256}';
+            }
+            else {
+                rcube::raise_error([
+                        'code' => 600, 'file' => __FILE__, 'line' => __LINE__,
+                        'message' => "Password plugin: Your PHP installation does not have the hash() nor sha1() function"
+                   ], true, true
+               );
+            }
+            break;
+
         case 'ssha512':
             $salt = rcube_utils::random_bytes(8);
 
--- a/plugins/password/password.php
+++ b/plugins/password/password.php
@@ -641,16 +641,19 @@ class password extends rcube_plugin
             $salt = rcube_utils::random_bytes(8);
 
             if (function_exists('hash')) {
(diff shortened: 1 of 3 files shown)


plugins/password/password.php
```diff
@@ -636,6 +636,23 @@ public static function hash_password($password, $method = '', $prefixed = true)
...
+            $salt = rcube_utils::random_bytes(8);
+
+            if (function_exists('hash')) {
+                $crypted = base64_encode( hash('sha256', $password . $salt, TRUE ) . $salt );
+                $prefix  = '{SSHA256}';
+            }
+            else {
+                rcube::raise_error([
+                        'code' => 600, 'file' => __FILE__, 'line' => __LINE__,
+                        'message' => "Password plugin: Your PHP installation does not have the hash() nor sha1() function"
```
alecpl, line 650: Remove "nor sha1()" from the message, please.

plugins/password/password.php
```diff
@@ -636,6 +636,23 @@ public static function hash_password($password, $method = '', $prefixed = true)
             $prefix  = '{SSHA}';
             break;
 
+        // base64 encoded ssha256 for mailcow
+        case 'ssha256':
+            $salt = rcube_utils::random_bytes(8);
+
+            if (function_exists('hash')) {
+                $crypted = base64_encode( hash('sha256', $password . $salt, TRUE ) . $salt );
```
alecpl, line 644: Redundant spaces around brackets. `true` in lower case. Do you have an idea why salt handling is slightly different for ssha512 below?

alecpl (changes requested):
I'd also mention ssha256 in config.inc.php.dist file, where we have a list of available algorithms.

plugins/password/password.php
TheFirstGoodmaN: :facepalm: is removed now

PR #8554, opened 2022-06-04
"about:blank" page triggers "load error"

endrik-exe:
"about:blank" page triggers "load error" when the iframe is appended to the document body


alecpl:
Thanks. I've found one more occurrence in the code. Fixed.

PR #8562, opened 2022-06-12
Add missing german translations in de_DE

stevemeier:
Added seven missing translations (noticed that "Folder successfully emptied." was not translated, so I checked all).


alecpl:
I updated the localization at https://www.transifex.com/roundcube/roundcube-webmail It will be updated in git next time we import localizations from Transifex.

PR #8571, opened 2022-06-14
Managesieve plugin: sieverules compatibility

marczi:
This PR adds on-the-fly conversion for sieve scripts that contain disable rules created by the abandoned sieverules plugin.
That plugin uses PHP serialization to add the disabled rule as a comment to the script. This modification can convert that PHP object structure to the format that the Managesieve plugin understands


alecpl:
Something's went wrong. I suppose a rebase is needed or something.

marczi:
Dear @alecpl I've fixed this PR as well, please rewrite it. Thanks in advance

alecpl:
It would be nice to have a test for this feature.

marczi:
Dear @alecpl thanks for the suggestion, I've added a test script. It helped to fix a warning of checking an undefined array key

alecpl:
The feature has been removed for security reasons. We don't want `unserialize()` on a user provided input.

PR #8578, opened 2022-06-15
Allow plugins to modify body after formating

Rotomeca:


program/lib/Roundcube/rcube_message.php
```diff
@@ -300,6 +300,14 @@ public function get_part_body($mime_id, $formatted = false, $max_bytes = 0, $mod
         if (!$mode && $body && $formatted) {
             $body = self::format_part_body($body, $part, $charset);
         }
+     
+        // allow plugins to modify body after formating
+        $plugin = $this->app->plugins->exec_hook('message_part_body_after',
```
alecpl, line 305: I think it makes sense only if `!$mode`.
Also, the hook is not executed for the case when `$part->body` is set, i.e. lines 257-286 above.

I think these make it not that useful.

PR #5335, opened 2016-06-22
HAproxy protocol header support for IMAP connections

logan893:
Introduce optional support to inject PROXY protocol headers after
opening IMAP TCP streams.

Version 1 (text based) and version 2 (binary) protocol header types are
supported. Supports both IPv4 and IPv6 style headers.

http://www.haproxy.org/download/1.6/doc/proxy-protocol.txt

PR #8625, opened 2022-07-22
HAproxy protocol header support for IMAP connections

cmollekopf:
Introduce optional support to inject PROXY protocol headers after
opening IMAP TCP streams.

Version 1 (text based) and version 2 (binary) protocol header types are
supported. Supports both IPv4 and IPv6 style headers.

http://www.haproxy.org/download/1.6/doc/proxy-protocol.txt

This rebases https://github.com/roundcube/roundcubemail/pull/5335 by @logan893 on top of master with slight adjustments.

Tested with nginx.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -158,6 +158,8 @@ $config['imap_auth_type'] = null;
 //
 // proxy_protocol is used to inject HAproxy style headers in the TCP stream
 // See http://www.haproxy.org/download/1.6/doc/proxy-protocol.txt
+// WARNING: Please note this is currently incompatible with implicit ssl,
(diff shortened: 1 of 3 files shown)


nward:
Hi - I have just come across this PR.

I have achieved the same use case by adding `x-originating-ip` to the IMAP ident. In my situation, this is preferable, as the IMAP endpoint that Roundcube connects to is the same as the one users can hit, so the haproxy header shows the Roundcube server IP (or IMAP client), and the IMAP ident shows the Roundcube end user IP.

I have implemented this as a plugin, but would be happy to submit it here to be included as an optional flag if that would be useful for the project?

PR #8666, opened 2022-08-09
Initialize HTTP client using get_http_client() for OAuth requests

jorgelzpz:
Hi, I noticed OAuth outgoing requests are not using the `get_http_client()` helper method, which means any options defined in `$config['http_client']` are not honored (e.g. custom proxy).

This PR switches from initializing a `GuzzleHttp\Client` from scratch to using the aforementioned method when dealing with OAuth requests.

### DIFF
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -17,7 +17,6 @@
  +-----------------------------------------------------------------------+
 */
 
-use GuzzleHttp\Client;
 use GuzzleHttp\MessageFormatter;
 use GuzzleHttp\Exception\RequestException;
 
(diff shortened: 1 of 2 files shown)


program/include/rcmail_oauth.php
```diff
@@ -219,8 +218,7 @@ public function request_access_token($auth_code, $state = null)
                 }
 
                 // send token request to get a real access token for the given auth code
-                $client = new Client([
-                    'timeout' => 10.0,
```
alecpl, line 223: @thomascube set this timeout for a reason, so let's not remove it. I propose to make the timeout a rcmail_oauth option, with this default value.

jorgelzpz: Fair enough, I have added a new `oauth_timeout` setting and the code now uses it.

Issue #8310, opened 2021-11-12
Remove configuration options for imap/smtp/managesieve/ldap port 


alecpl:
It is not stated clearly in the configuration file comments, but all "host" options support `host:port` syntax. Therefore I suggest we remove the associated "port" options.

For IMAP:
- remove `default_port` option
- if `default_host` does not include a port, use 993 if ssl:// prefix is used, 143 otherwise.
- we could also rename `default_host`, to `imap_host`.

For SMTP:
- remove `smtp_port` option
- if `smtp_server` does not include a port, use 465 if ssl:// prefix is used, 587 otherwise.
- we could also rename `smtp_server`, to `smtp_host`.

For LDAP:
- remove `port` option from `ldap_public` array.
- remove `use_tls` option from `ldap_public` array.
- if `hosts` does not include a port, use 636 if ldaps:// prefix is used, 389 otherwise.

For managesieve:
- remove `managesieve_port` option.
- remove `managesieve_usetls` option
- if `managesieve_host` does not include a port, determine it automatically with a fallback to 4190.

alecpl:
@thomascube any objections?

thomascube:
@alecpl Agreed. We just have to document this breaking change or automatically migrate old configs.

alecpl:
Done.

Issue #7127, opened 2019-12-22
Cannot connect to Dovecot Manage Sieve when SSL/TLS is enabled


Frederick888:
I'm using Roundcube with Dovecot. I just enabled Manage Sieve in Roundcube and Dovecot and it worked flawlessly without SSL/TLS. However, Roundcube can no longer connect to Manage Sieve service after SSL/TLS was turned on.

#### Software Versions
Ubuntu: 18.04.3
PHP: 7.3.13-1+ubuntu18.04.1+deb.sury.org+1
dovecot / dovecot-sieve / dovecot-managesieved: 2.2.33.2-1ubuntu4.5
Roundcube: 1.4.1

#### Roundcube logs
```
==> logs/errors.log <==
[22-Dec-2019 08:50:33 +0000]: <la88taa7> PHP Error: Failed to read from socket (GET /?_task=settings&_action=plugin.managesieve)
[22-Dec-2019 08:50:33 +0000]: <la88taa7> PHP Error: Unable to connect to managesieve on imap.mydomain.com:4190 in /var/www/mail/plugins/managesieve/lib/Roundcube/rcube_sieve_engine.php on line 223 (GET /?_task=settings&_action=plugin.managesieve)
[22-Dec-2019 08:50:33 +0000]: <la88taa7> PHP Error: Not currently in AUTHORISATION state (GET /?_task=settings&_action=plugin.managesieve)
[22-Dec-2019 08:50:33 +0000]: <la88taa7> PHP Error: Failed to read from socket (GET /?_task=settings&_action=plugin.managesieve)
```

#### Dovecot Manage Sieve Configuration
```
# /etc/dovecot/conf.d/20-managesieve.conf
service managesieve-login {
  inet_listener sieve {
    port = 4190
    ssl = yes
  }
  inet_listener sieve_deprecated {
    port = 2000
    ssl = yes
  }
  service_count = 1
  process_min_avail = 0
  vsz_limit = 64M
}
service managesieve {
  process_limit = 1024
}
protocol sieve {
}
```

#### Roundcube managesieve configuration
```php
// /var/www/mail/config/config.inc.php
$config['managesieve_host'] = 'tls://imap.%d';
$config['managesieve_usetls'] = true;
$config['managesieve_debug'] = true;
```

#### OpenSSL Test Results
First I tested with `openssl s_client -connect imap.mydomain.com:4190 </dev/null` in the same machine and then I also tested my PHP OpenSSL configuration with the following script:
```php
[...]

PR #8670, opened 2022-08-11
Restore support for ManageSieve over implicit SSL

smokris:
Prior to #8310 (893216cb297268d222ae49099e6654a304b72e3f), it was possible to connect to ManageSieve over implicit SSL using a config like this:
```php
$config['managesieve_host'] = 'ssl://mail-0.local:4190';
```
(See https://github.com/roundcube/roundcubemail/issues/7127#issuecomment-568284068 for example.)

Roundcube's ManageSieve plugin would leave the host's `ssl://` prefix intact, and pass it through to `Net_Sieve`, to `Net_Socket`, and finally to PHP's `stream_socket_client()` `$address` argument.  But as of #8310 (893216cb297268d222ae49099e6654a304b72e3f), it now strips off the `ssl://` URI scheme, so `stream_socket_client()` doesn't know that it should implicitly perform the SSL handshake.

This PR restores support for implicit SSL connections by preserving the `ssl://` URI scheme, and mentions support for it in the example config.  I've tested this locally using Dovecot Pigeonhole.

### DIFF
--- a/plugins/managesieve/config.inc.php.dist
+++ b/plugins/managesieve/config.inc.php.dist
@@ -8,7 +8,8 @@
 // For example %n = mail.domain.tld, %d = domain.tld
 // If port is omitted it will be determined automatically using getservbyname()
 // function, with 4190 as a fallback.
-// Note: Add tls:// prefix to enable TLS
+// Note: Add tls:// prefix to enable explicit STARTTLS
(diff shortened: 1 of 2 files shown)


smokris:
Thanks, @alecpl!

PR #8684, opened 2022-08-20
Add password driver for the gandi.net mail API

Kartoffelbauer:
I recently switched to Roundcube and was surprised to see how well designed this app is.
However, when setting up the password plugin, I noticed that there is no password driver for my provider.
So I did the only reasonable thing and coded one myself.

The driver uses the gandi.net mail API (https://api.gandi.net/docs/email/) to update the user password. It also includes a strength driver that conforms to the gandi.net rules for changing passwords, and of course some translations and documentation. It has been tested with PHP 7.4 on Roundcube 1.6.

This took al lot of work and I would really appreciate it if you would look over it and consider a merge.

### DIFF
--- a/plugins/password/README
+++ b/plugins/password/README
@@ -53,6 +53,7 @@
  2.1.27. dovecot_passwdfile
  2.1.28. Mailcow
  2.1.29. LDAP Samba Active Directory (ldap_samba_ad)
+ 2.1.30  gandi (gandi.net mail-API)
  2.2.    Password Strength Drivers
  2.2.1.  Zxcvbn
  2.2.2.  Have I been pwned? (pwned)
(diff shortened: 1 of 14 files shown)


alecpl:
Don't add localizations, they need to be done via Transifex. They will got removed when we sync the localization.

Kartoffelbauer:
Happy to hear that (-:
Do I have to register my text for translation somewhere or does this happen automatically?

alecpl:
Transifex pulls it automatically from en_US.inc (from master branch, i.e. after it's merged).

alecpl:
I hesitate to merge this because gandi.net is not open-source nor can be installed on premises. Am I wrong? Also, its password policy is very non-generic, i.e. we need an extra localization label to handle that.

Kartoffelbauer:
Yes, gandi.net is a domain provider and although their API is very well documented, it is not open-source. And yes, there is no (clean) way to add password policy other than through localization labels.

While I'd be very happy to see this merged, I also see the issue with closed source code, so you'll have to decide.

PR #8690, opened 2022-08-26
darkmode blockquote patch

pswilde:
Currently in dark mode, a blockquote in an email message content appears as a dark background with dark grey text. This makes reading blockquotes in message content nearly impossible without switching to light mode.
I have added a new color: `@color-dark-blockquote-background: #f8f9f9;` which matches the lightmode blockquote background and applied this color to blockquotes in the dark.less file.
This should make blockquotes appear readable within the rest of the email content.

### DIFF
--- a/skins/elastic/styles/colors.less
+++ b/skins/elastic/styles/colors.less
@@ -222,15 +222,16 @@
 
 
 // Dark mode colors
-@color-dark-main:           darken(@color-main, 30%);
-@color-dark-background:     #21292c;
-@color-dark-font:           #c5d1d3;
-@color-dark-border:         #4d6066;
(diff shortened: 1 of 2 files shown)


alecpl:
I'm unable to reproduce the issue. The text inside blockquotes is not dark-grey. For me it's like this:
quoting

pswilde:
Thanks for your reply. To be honest I can't quite remember the situation which made this happen for me, it was a while ago now. In truth, it may be the version bundled with my email service was not up to date and this may already have been fixed in a later version.
Regardless, it is no longer causing me a problem for whatever reason so I'll close this one down. Thanks.

PR #8701, opened 2022-09-07
Added prioritized charsets for Slovenian language

rokj:


alecpl:
This method is deprecated and not used anymore, but I guess we can merge that.

PR #8733, opened 2022-09-29
Wrap the mail date in a <time> tag

williamdes:
Some screen readers and other tools could benefit from a machine readable date.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time
- https://www.htmlgoodies.com/html5/its-time-for-an-html5-time-tag/
- https://css-tricks.com/time-element/

It's my first contribution here, so I may not be well polished about ways to code here

Note: adding a `title` attribute with some custom date format would allow a nice tooltip on the date for the user to see another representation of the date. I was initially thinking that browsers would do that but in fact no. So anyway I contribute this and hope you will like it. It's just more standard HTML for a date or time to be wrapped in such a tag

### DIFF
--- a/program/actions/mail/show.php
+++ b/program/actions/mail/show.php
@@ -473,8 +473,11 @@ class rcmail_action_mail_show extends rcmail_action_mail_index
             $header_value = null;
 
             if ($hkey == 'date') {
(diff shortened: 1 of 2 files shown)


williamdes:
Nobody even did interact with this..

PR #8742, opened 2022-10-11
fixing PHP Warning:  Undefined array key 3 in /var/www/html/program/l…

gs4711:
Fixing out-of-bound exceptions:

   …ib/Roundcube/rcube_imap.php on line 2067

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -2064,7 +2064,7 @@ class rcube_imap extends rcube_storage
                     $struct->ctype_secondary = strtolower($part[$i]);
 
                     // read content type parameters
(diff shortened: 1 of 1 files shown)


program/lib/Roundcube/rcube_imap.php
```diff
@@ -2064,7 +2064,7 @@ protected function structure_part($part, $count = 0, $parent = '', $mime_headers
                     $struct->ctype_secondary = strtolower($part[$i]);
 
                     // read content type parameters
-                    if (is_array($part[$i+1])) {
+                    if (($i+1 < count($part)) && is_array($part[$i+1])) {
```
alecpl, line 2067: I'd just do isset() && is_array().

alecpl:
Already fixed.

PR #8743, opened 2022-10-11
Rcube result index

gs4711:
fixing PHP Warning:  Undefined array key \"ALL\" in /var/www/html/program/lib/Roundcube/rcube_result_index.php on line 427

### DIFF
--- a/program/lib/Roundcube/rcube_result_index.php
+++ b/program/lib/Roundcube/rcube_result_index.php
@@ -424,7 +424,12 @@ class rcube_result_index
         $params['ORDER']   = $this->order;
 
         if ($param !== null) {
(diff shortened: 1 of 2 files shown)


program/lib/Roundcube/rcube_result_index.php
```diff
@@ -424,7 +424,12 @@ public function get_parameters($param = null)
         $params['ORDER']   = $this->order;
 
         if ($param !== null) {
-            return $params[$param] ?? null;
+            if (array_key_exists($param, $params)) {
+                return $params[$param] ?? null;
```
alecpl, line 428: This change does not make sense.

PR #8749, opened 2022-10-16
Pass port 0 as integer to Redis::connect() with unix socket

mstilkerich:
The second parameter to Redis::connect() is an int parameter. Passing null for it will trigger a PHP deprecation notice with PHP 8.1.

The Redis::connect() documentation shows that a 0 is used for the port parameter in the examples with unix sockets where parameters beyond the second parameter are set.

### DIFF
--- a/program/lib/Roundcube/cache/redis.php
+++ b/program/lib/Roundcube/cache/redis.php
@@ -110,7 +110,7 @@ class rcube_cache_redis extends rcube_cache
 
             if (substr($redis_host, 0, 7) === 'unix://') {
                 $host = substr($port, 2);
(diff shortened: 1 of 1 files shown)


mstilkerich:
With PHP 8.1 and using a unix socket to connect to redis, the roundcube error.log is cluttered with deprecation notices like:

```
[16-Oct-2022 11:54:30 Europe/Berlin] PHP Deprecated:  Redis::connect(): Passing null to parameter #2 ($port) of type int is deprecated in /var/www/roundcube/program/lib/Roundcube/cache/redis.php on line 122
```

PR #8750, opened 2022-10-17
update zh_CN Label

hihuangwei:
update zh_CN Label


alecpl:
I merged this, but localization should be done via [Transifex.](https://explore.transifex.com/roundcube/roundcube-webmail/)

Issue #8143, opened 2021-07-23
Display Brand Indicator Message Identification (BIMI) images


candrews:
> Brand Indicator Message Identification (BIMI) is an industry-wide standards effort to use brand logos as indicators to help email recipients recognize and avoid fraudulent messages.

https://blog.mxtoolbox.com/2019/07/03/bimi-record-what-is-it-how-does-it-add-to-dmarc/

It would be nice if roundcube displayed these images like this mockup shows:
![gmail with bimi](https://mxtoolbox.files.wordpress.com/2019/07/bimi-inbox-after.png?w=625)

In short, to get the bimi image, check for a TXT DNS record at `default._bimi.`<domain>. The contents of that record will look like this:
`v=BIMI1;l=https://www.integralblue.com/logo.svg`
The `l` value is the URL of the image, which is always an SVG.

alecpl:
There's a draft (https://datatracker.ietf.org/doc/draft-blank-ietf-bimi/) but I wouldn't call that a standard (yet).

Anyway, sounds like a good idea for a plugin.

m-ueberall:
For the record, [rspamd release 3.2 added a preliminary version of its BIMI plugin](https://github.com/rspamd/rspamd/blob/79417a5f81e6d990364d3a541928b278f30492d2/ChangeLog#L11). Links of potential interest in case a plugin is going to make use of the associated headers and service/helper:
* [[Feature] Add BIMI support #3935](https://github.com/rspamd/rspamd/issues/3935)
* [BIMI helper service](https://github.com/rspamd/bimi-helper)

da2x:
Google, Yahoo!, FastMail, and Pobox now [supports BIMI](https://bimigroup.org/bimi-infographic/); with other providers considering it.

InputOutputZ:
Here is a PR for BIMI Support, refer to #8759, a tiny contribution from me.

candrews:
I've submitted an MR that adds a bimi plugin: https://github.com/roundcube/roundcubemail/pull/8860

The plugin supports caching and does svg filtering. I'm eager to receive feedback and :crossed_fingers: get it merged :smile:

knofte:
Hey champs, any update to this? :)

Neustradamus:
If someone can update the @candrews PR with latest BIMI draft version?
- https://github.com/roundcube/roundcubemail/pull/8860
- https://datatracker.ietf.org/doc/html/draft-brand-indicators-for-message-identification

[...]

PR #8759, opened 2022-10-24
Adding BIMI Support

InputOutputZ:
Adding BIMI Support

### DIFF
--- a/program/actions/mail/show.php
+++ b/program/actions/mail/show.php
@@ -392,6 +392,25 @@ class rcmail_action_mail_show extends rcmail_action_mail_index
             $attrib['onerror'] = "this.onerror = null; this.src = '$placeholder';";
         }
 
+        // Check BIMI if is present, and submit in place of error src, otherwise default placeholder.
+        $domain = explode("@",self::$MESSAGE->sender['mailto']);
+        if(sizeof($domain) >= 2){
+            $domain = $domain[sizeof($domain)-1];
+            $result = dns_get_record("default._bimi.".$domain, DNS_TXT);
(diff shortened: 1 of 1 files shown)


alecpl:
I think that a better place for this would be in program/actions/contacts/photo.php. Also, the content will have to be secured using `rcmail_attachment_handler::svg_filter()`. Finally, I'd probably store the svg file in the temp folder so every user could just use it skipping the DNS request.

InputOutputZ:
Aha, I agree. I'm not sure how to use properly the plugins framework, therefore when you have time, please have a look at it. I will try to pick things perhaps and use it in future contributions. 

With thanks.

alecpl:
Actually you could use `identicon` plugin as an example. It does exactly what you need.

candrews:
I've submitted a PR that adds a bimi plugin: https://github.com/roundcube/roundcubemail/pull/8860

The plugin supports caching and does svg filtering as discussed here in this PR. I'm eager to receive feedback and :crossed_fingers: get it merged :smile:

knofte:
How's this going?
Would be great to get bimi support (without VMC) in base :)

Neustradamus:
If someone can update the your @candrews PR with latest BIMI draft version?
- https://github.com/roundcube/roundcubemail/pull/8860
- https://datatracker.ietf.org/doc/html/draft-brand-indicators-for-message-identification

Thanks in advance.

Neustradamus:
Some people ask news about BIMI:
- https://github.com/roundcube/roundcubemail/issues/8143

PR #8766, opened 2022-10-28
Purge messages from junk folder when delete_junk enabled

shaun-a-johnson:
If delete_junk is on, end user expectation is that both single message delete as well as 'empty folder' operations should immediately delete messages from junk and bypass trash.


shaun-a-johnson:
I am uncertain how to resolve the failed check - I walked through manually and everything appears to be working as expected with the 'Mark as Junk' plugin with this patch -  was i perhaps supposed to fork from something other than master?

Issue #8738, opened 2022-10-05
Roundcube does not work anymore when using a reverse proxy with a subpath


Diman0:
After upgrading Roundcube 1.5.3 to the latest version 1.6.0, it no longer works under a reverse proxy, under a subpath.

Log-in is redirecting to `/?_task=login` instead of `/subpath/?_task=login`.
The old behaviour was `./?_task=login`

This was working before. See also [#186](https://github.com/roundcube/roundcubemail-docker/issues/186).

J0riz:
This issue is related to #8170 

This is caused by a new `$_SERVER['REQUEST_URI']` check introduced in RoundCube 1.6 in `program/include/rcmail.php` on line 1155

I did for now just disable the check what doesn't sound like a nice solution but makes it work again if you use a reverse proxy:

```
        if ($absolute || $full) {
            // add base path to this Roundcube installation
            $prefix = $base_path ?: '/';

            // prepend protocol://hostname:port
            if ($full) {
                $prefix = rcube_utils::resolve_url($prefix);
            }

            $prefix = rtrim($prefix, '/') . '/';
        }
        else {
            // Disable check for REQUEST_URI
            //if (isset($_SERVER['REQUEST_URI'])) {
            //    $prefix = preg_replace('/[?&].*$/', '', $_SERVER['REQUEST_URI']) ?: './';
            //}
            //else {
                $prefix = './';
            //}
        }
```

I hope someone can come up with a more nice solution.

thomascube:
@alecpl This seems like a major regression and should be fixed in a 1.6.x release. There are a number of issues related to this reported to the docker repo where Roundcube mostly runs behind a reverse proxy.

alecpl:
Feel free to work on this if you know a solution that would cover this issue and #8170 case.

thomascube:
@alecpl Please review the proposed solution in PR #8770

PR #8770, opened 2022-10-31
Add config option for request uri field (#8738)

thomascube:
This can be used to read a custom header sent by a reverse proxy to resolve the absolute path to Roundcube.

Resolves issue #8738

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -817,6 +817,13 @@ $config['no_save_sent_messages'] = false;
 // Note: Use assets_path to not prevent the browser from caching assets
 $config['use_secure_urls'] = false;
 
+// Specify the $_SERVER field that contains the full path of the original HTTP request.
+// This might be changed when Roundcube runs behind a reverse proxy using a subpath.
+// The reverse proxy config can specify a custom header (e.g. X-Forwarded-Path) containing
+// the path under which Roundcube is exposed to the outside world (e.g. /rcube/).
+// This header vaue is then availbale in PHP with $_SERVER['HTTP_X_FORWARDED_PATH'].
+$config['request_uri_field'] = 'REQUEST_URI';
+
 // Allows to define separate server/path for image/js/css files
 // Warning: If the domain is different cross-domain access to some
 // resources need to be allowed
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -1119,9 +1119,9 @@ class rcmail extends rcube
         }
 
         $base_path = '';
-        $server_var = $this->get_requesr_uri_field();
+        $server_var = $this->get_request_uri_field();
         if ($server_var && !empty($_SERVER[$server_var])) {
-            $base_path = preg_replace('/[?&].*$/', '', $_SERVER[$server_var]) ?: './';
+            $base_path = preg_replace('/[?&].*$/', '', $_SERVER[$server_var]);
         }
         else if (!empty($_SERVER['REDIRECT_SCRIPT_URL'])) {
             $base_path = $_SERVER['REDIRECT_SCRIPT_URL'];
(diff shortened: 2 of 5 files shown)


alecpl:
When we deal with proxy headers we should probably check `proxy_whitelist` (see rcube_utils for a reference usage).

Also, wouldn't it be enough to just set the path in the config for simplicity? I guess use of a header covers mixed or multi-host setups, but are they common?

thomascube:
I don't know how common it is to use (different) subpaths on a reverse proxy. The proposed solution with a config option `request_uri_field` is a generic approach to resolve issues with wrong guessing of the request URL. It's not directly related to x-fowarded-* headers and thus the `proxy_whitelist` check might not be appropriate.

program/include/rcmail.php
```diff
@@ -1154,17 +1158,24 @@ public function url($p, $absolute = false, $full = false, $secure = false)
...
         }
 
         return $prefix . $url;
     }
 
+    /**
+     * Get the 'request_uri_field' config option
+     * with an additional check against the 'proxy_whitelist' config
+     */
+    protected function get_requesr_uri_field() {
```
alecpl, line 1171: A typo in the function name. Also, add a new line before `{`

program/include/rcmail.php
```diff
@@ -1154,17 +1158,24 @@ public function url($p, $absolute = false, $full = false, $secure = false)
             $prefix = rtrim($prefix, '/') . '/';
         }
         else {
-            if (isset($_SERVER['REQUEST_URI'])) {
```
alecpl, line 1157: You moved that block to another place, I'll have to test this PR.

config/defaults.inc.php
```diff
@@ -817,6 +817,13 @@
 // Note: Use assets_path to not prevent the browser from caching assets
 $config['use_secure_urls'] = false;
 
+// Specify the $_SERVER field that contains the full path of the original HTTP request.
+// This might be changed when Roundcube runs behind a reverse proxy using a subpath.
+// The reverse proxy config can specify a custom header (e.g. X-Forwarded-Path) containing
+// the path under which Roundcube is exposed to the outside world (e.g. /rcube/).
+// This header vaue is then availbale in PHP with $_SERVER['HTTP_X_FORWARDED_PATH'].
+$config['request_uri_field'] = 'REQUEST_URI';
```
alecpl, line 825: I wonder whether this makes sense to be set at all. And whether this option is useful without proxy. Maybe it should be `$config['proxy_uri_field'] = null;` or `$config['request_path'] = null;` I'm not sure.

alecpl (changes requested):
Looks like two tests fail.

program/include/rcmail.php
thomascube: Yes that was indented in order to make this work also for the `$absolute || $full` mode.

config/defaults.inc.php
thomascube: Depending on the PHP/webserver configuration the superglobal fields with path info may vary. The fact that we check for `REDIRECT_SCRIPT_URL` or `SCRIPT_NAME` already in the code underlines that. And there's `PATH_INFO` or `PHP_SELF` out there, too. Using a generic config option might help mitigating other potential problems with PHP environments not only related to reverse proxies.

alecpl (approved):

alecpl:
Approved. Though I'd have to find out why I still have `./phpunit` instead of `/vendor/bin/phpunit` on my setup.

ghnp5:
Hey guys. Thank you for the fix. If this was approved, can this be merged so that this can go on the next release, please? 😊
Cheers!

config/defaults.inc.php
msimerson: Big +1 from me for declaring exactly what the request path base should be. It will work reliably in any environment and doesn't require sending additional mostly duplicated and mostly useless tokens with every proxied HTTP request.

```diff
- $config['request_path'] = null;
+ $config['request_path'] = '/rcmail';
```

It's explicit. There's no parsing of it required.

alecpl:
I renamed the option to `request_path` and added possibility to set it just to a real path. `$_SERVER` field is also supported.

ghnp5:
Hey. Any update on when will this be released to the Docker version? I'm using the latest version of Roundcube and this is still broken. I haven't been able to use Roundcube at all since this has been broken. Thank you very much!

PR #8773, opened 2022-11-02
Use json_decode to prevent exception

vitoshalabs:
On our install ( Debian testing with rouncube 1.6.0+dfsg-1.1 ) we get:
```
PHP Fatal error:  Uncaught Error: Call to undefined function GuzzleHttp\json_decode() in /usr/share/roundcube/program/include/rcmail_oauth.php:237
```
 changing the function call from `\GuzzleHttp\json_decode` to `json_decode` fixes it and works fine

### DIFF
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -234,7 +234,7 @@ class rcmail_oauth
                         ],
                 ]);
 
-                $data = \GuzzleHttp\json_decode($response->getBody(), true);
+                $data = json_decode($response->getBody(), true);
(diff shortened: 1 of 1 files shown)


alecpl:
This indicates that you're using Guzzle < 6.2, but we require 7.3.

I guess we could merge this anyway.

vitoshalabs:
> This indicates that you're using Guzzle < 6.2, but we require >=7.3.
> 
> I guess we could merge this anyway.

Thanks we will upgrade Guzzle

PR #8778, opened 2022-11-08
Fixing max_filesize calculation for upload.

rokj:


alecpl:
This does not make sense.

rokj:
> This does not make sense.

What does not make sense?

In https://github.com/roundcube/roundcubemail/pull/8778/commits/14b12ed8565b165328833b61cbda2dd5a61b9366#diff-4264f6df68ee68aa17181f78aa680b7c968c4f854df0b88f9ea49dce2fca4249L477

there is check in if case where you are checking with logical operator && the same variable. This always returns 1.

For example:
```
$max_size = 12323543454454542354;

if ($max_size && $max_size < $max_filesize) ...

$max_size && $max_size  this returns 1
```

alecpl:
I think you should learn PHP some more.

alecpl:
Maybe if the code was `if ($max_size > 0 && $max_size < $max_filesize)` it would be more clear.

rokj:
> Maybe if the code was `if ($max_size > 0 && $max_size < $max_filesize)` it would be more clear.

Then fix it in your way, because it calculates wrong max_filesize and this is a bug. Try setting post_max_size > 35MB and upload_max_filesize > 35 for example... and user will show wrong max filesize he can upload.

alecpl:
No, it doesn't.

rokj:
You're right. I should learn PHP.

PR #8802, opened 2022-11-25
Disable `gc_temp` in favor of an external gc

sebastka:
Use an external garbage collector to clean up the `temp_dir`.
Useful to avoid `readdir` over large directories on network shares


alecpl:
The same (almost) could be done by setting `session.gc_probability=0` and using bin/gc.sh periodically.

PR #8828, opened 2022-12-23
Optimize SVG files

dilyanpalauzov:
by calling

`svgo --multipass $(find -name '*.svg')`

svgo ⇔ https://github.com/svg/svgo

This will reduce the data transferred to the browser.


alecpl:
The savings aren't that big. I prefer to keep the structure easy to read in case of svg.

PR #8839, opened 2022-12-29
add script to manage identities

vladasko-g:


bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
...
+  $new_identity['standard'] = $options['S'];
+}
+
+$rcmail = rcube::get_instance();
+
+$args = rcube_utils::get_opt([
+        'u' => 'user',
+        'h' => 'host',
+        'd' => 'dir',
+        'x' => 'dry-run',
```
alecpl, line 83: 'dir' and 'dry-run' aren't used here. Also, why are you using both getopt() and rcube_utils::get_opt()? Please, ose rcube_utils::get_opt() only.

bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env php
+
```
alecpl, line 2: You should copy the copyright header from other scripts, please.

bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
...
+# -e email (new identity) - e.g. jonesmith@example.com
+# -n name                 - e.g. John Smith
+# -o organization         - e.g. Company ABC
+# -s signature            - e.g. '<h2>Best Wishes</h2>'
+# -b bcc                  - e.g. jonesmith@example.com
+# -r reply-to             - e.g. jonesmith@example.com
+# -H is html signature    - e.g. 0 (off) or 1 (on)
+# -S standard (default)   - e.g. 0 (off) or 1 (on)
+
+<?php
```
alecpl, line 18: I'd start the PHP script just after  the shebang.

bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
...
+
+define('INSTALL_PATH', realpath(__DIR__ . '/..') . '/' );
+
+require INSTALL_PATH.'program/include/clisetup.php';
+
+$options = getopt('u:e:n:o:s:b:r:H:S:');
+
+if (isset($options['u'])) {
+  $username = $options['u'];
+} else {
```
alecpl, line 28: Please, 4-space indentation.

bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
...
+
+function validateEmail($email, $fieldName) {
+  if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
+    echo "invalid $fieldName format";
+
+    exit;
+  }
+}
+
+function get_host($args)
```
alecpl, line 99: We're doing the same in deluser.sh. I guess we could de-duplicate this and create rcube_utils::get_imap_host_opt().

bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
...
+
+  $new_identity['bcc'] = $options['b'];
+}
+if (isset($options['r'])) {
+  validateEmail($options['r'], 'reply-to email');
+  
+  $new_identity['reply-to'] = $options['r'];
+}
+if (isset($options['S'])) {
+  $new_identity['standard'] = $options['S'];
```
alecpl, line 74: If you allow setting this flag you should unset this flag from the existing default identity, if such exists. I.e. you should use `rcube_user::set_default()` and not make it part of the `$options` array.

bin/addidentity.sh
```diff
@@ -0,0 +1,139 @@
...
+}
+
+if (isset($options['o'])) {
+  $new_identity['organization'] = $options['o'];
+}
+if (isset($options['s'])) {
+  $new_identity['signature'] = $options['s'];
+}
+if (isset($options['H'])) {
+  $new_identity['html_signature'] = $options['H'];
```
alecpl, line 61: Please, make sure this is either '1' or '0'.

alecpl (changes requested):
We should also check the `identities_level` option. It may disallow some actions.

bin/addidentity.sh
vladasko-g: thanks for all the useful feedback, I forgot to mark this as draft, I'll update the PR

bin/addidentity.sh
vladasko-g: updated

bin/addidentity.sh
vladasko-g: updated

bin/addidentity.sh
vladasko-g: updated

bin/addidentity.sh
vladasko-g: I moved this part to utils, but only changed the usage here

bin/addidentity.sh
vladasko-g: updated

bin/addidentity.sh
vladasko-g: updated

vladasko-g:
moved here https://github.com/roundcube/roundcubemail/pull/8887

Issue #8642, opened 2022-08-01
bug: updater does not update the vendor packages


damnms:
because of some unrelated comments in #8639 i'm going to create this bug report.

steps to reproduce:
1) have an install of 1.5.3 configured with smtp_server
2) download the 1.6-complete tar.gz 
3) run the bin/installto script and point it to the 1.5.3 location
4) answer 'y' when it asks if it should migrate the old configs

expected:
the vendor packages should be updated too (e.g. net_smtp from 1.7 to 1.10)

actual behaviour:
it does not update these files

additional information:
i have a quiet restricted system, so maybe its an permission problem, but then i would expect an error that these files could not be overwritten

alecpl:
I see where the problem might be. The script indeed does not sync the vendor folder if `composer.json` file exists in the target location. This is because your installation might have other packages (plugins and their dependencies) installed, and we don't want to break that.

One problem is that since 1.4.0 the release packages include `composer.json` (and `composer.lock`) files. So, this "update vendor folder if user is not using composer" idea does not work anymore.

I'm not sure there is an easy fix. Maybe we should just require composer to upgrade Roundcube. Then we should probably bump version of our dependencies to the required value in composer.json and run `composer update` or tell the user to do so.

@thomascube, any ideas?

ps. a workaround for everyone not using composer and want to upgrade, remove `composer.json` file from your installation before you execute `installto.sh` script. For those using composer the way is to bump required versions in your `composer.json` file and run `composer update`. Required versions can be found in `composer.json-dist` file.

damnms:
If roundcube requires any higher version of a vendor package, i would update that in any way - IMO it makes no sense to update roundcube and not the required vendor package - because then it is for sure broken (like it was on my installation).
Maybe it would be a good way to just print out a big warning "Found custom plugins, make sure they work with the latest vendor packages roundcube ships" or so.

seanm:
We had the same `Call to undefined method Net_SMTP::starttls` error in bug #8698 (using postfix+dovecot) after updating roundcube 1.5.3 to 1.6.0.

We moved `vendor/pear` (since a grep for `Net_SMTP` found stuff in that folder) from the download tarball into our installation and that fixed it.  Subsequently we nuked our `vendor` and replaced it all.

So indeed it seems `bin/installto.sh` does not do some necessary things.  We saw it logged a message:

`NOTICE: Update dependencies by running php composer.phar update --no-dev`

[...]

PR #8840, opened 2022-12-29
Update vendor dir from installto.sh if untouched (#8642)

thomascube:
* installto.sh checks for a marker indicating that composer was not run/used.
* update.sh tries to find and run composer


bin/update.sh
```diff
@@ -299,3 +308,25 @@ function repo_key($repo)
 
     return $key;
 }
+
+function vendor_dir_untouched()
+{
+    system('grep -q "generated by Roundcube install" ' . INSTALL_PATH . 'vendor/autoload.php 2>/dev/null', $exit_code);
```
alecpl, line 314: INSTALL_PATH might include spaces.

bin/update.sh
```diff
@@ -299,3 +308,25 @@ function repo_key($repo)
...
+    return $exit_code === 0;
+}
+
+function find_composer()
+{
+    if (is_file(INSTALL_PATH . 'composer.phar')) {
+        return 'php composer.phar';
+    }
+    else {
```
alecpl, line 323: If there's a return, `else` can be removed.

bin/installto.sh
```diff
@@ -146,3 +146,10 @@ if (strtolower($input) == 'y') {
 else {
     echo "Update cancelled. See ya!\n";
 }
+
+
+function vendor_dir_untouched($target_dir)
+{
+    system('grep -q "generated by Roundcube install" ' . $target_dir . '/vendor/autoload.php 2>/dev/null', $exit_code);
```
alecpl, line 153: `$target_dir` might include spaces.

Edit: On the other hand we don't care about this in other places where we deal with INSTALL_PATH. So, nevermind.

bin/update.sh
```diff
@@ -267,7 +267,16 @@ if ($RCI->configured) {
...
 
-        echo "NOTICE: Update dependencies by running `php composer.phar update --no-dev`\n";
+        if (!vendor_dir_untouched()) {
+            $exit_code = 1;
+            if ($composer_bin = find_composer()) {
+                echo "Executing " . $composer_bin . " to update dependencies...\n";
+                echo system('(cd ' . INSTALL_PATH . "; $composer_bin update --no-dev)", $exit_code);
+            }
+            if ($exit_code != 0) {
+                echo "ATTENTION: Update dependencies by running `php composer.phar update --no-dev`\n";
```
alecpl, line 277: I was thinking that we could add extra "-------------------" line on to and bottom to make it even more visible.

bin/update.sh
```diff
@@ -299,3 +308,25 @@ function repo_key($repo)
 
     return $key;
 }
+
+function vendor_dir_untouched()
```
alecpl, line 312: I suppose we could move these extra functions to rcmail_install.

alecpl (commented):
This solution makes sense, but will not properly update existing installations (that do not have the text in autoload file, e.g. 1.5.3, 1.6.0). So, we might still need to add some notes to the UPGRADING file.

alecpl (approved):

PR #8853, opened 2023-01-09
Clean duplicates in autocomplete list

AGoubs:
When a contact is displayed multiple time in the autocompletion list, this will clean the list to display this user only once.
There is an 'autocomplete_clean_duplicates' (false by default) parameter to activate this function in the configs.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -1230,6 +1230,9 @@ $config['autocomplete_threads'] = 0;
 // Max. number of entries in autocomplete popup. Default: 15.
 $config['autocomplete_max'] = 15;
 
+// Clean duplicates in automcomplete list. Default: false
+$config['autocomplete_clean_duplicates'] = true;
+
 // show address fields in this order
 // available placeholders: {street}, {locality}, {zipcode}, {country}, {region}
 $config['address_template'] = '{street}<br/>{locality} {zipcode}<br/>{country} {region}';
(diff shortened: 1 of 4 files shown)


dilyanpalauzov:
I have the same contact in two different address books: once from an LDAP address book, once from a CardDAV address book.  In the mail composer window, when the autocompletion is utilized, that contact appears only once.

In which use case does this changeset make a difference?

dilyanpalauzov:
The above information is not very correct.  I have configured the LDAP addressbook as 
```
$config['ldap_public']['ldap'] = array(
  'name' => 'LDAP (AEGEE Bodies)',
  'hosts' => array('tls://ldap.aegee.org'),
  'user_specific' => false,
…
$config['ldap_debug'] = true;
```
When the CardDAV address book is disabled, the AB-search does not find the contact./tmp/ldap.log contains:
```
[20-Jan-2023 15:39:08 +0000]: <cfb25602> C: Connect [ldap://ldap.aegee.org:389:389]
[20-Jan-2023 15:39:08 +0000]: <cfb25602> S: OK
[20-Jan-2023 15:39:09 +0000]: <cfb25602> C: Close
```
so 389 is added twice.

pabzm:
@AGoubs Thank you for the contribution!

Are you still interested in this PR? (It's been some time since you posted.)

I noticed your code compares names or display-names. Shouldn't it rather compare email-addresses? Or even both, name and email-address? I might have one person with different addresses in my addressbooks, or different people with an identical name, which should definitely be shown separately.

Another issue I see is that it changes code that's also used by the `acl`-plugin, which might have unintended side-effects. It might be better to filter the list of contacts in `program/actions/mail/autocomplete.php` directly. Do you feel comfortable to move the filtering to PHP?
If the filtering gets more conservative we might not even put it behind a config option, because it's just right to not show duplicate addressbook entries.

PR #8860, opened 2023-01-14
Add BIMI plugin (#8143)

### LINKED ISSUE #8143: Display Brand Indicator Message Identification (BIMI) images

candrews:
Brand Indicator Message Identification (BIMI) is an industry-wide standards effort to use brand logos as indicators to help email recipients recognize and avoid fraudulent messages.

See: https://datatracker.ietf.org/doc/draft-brand-indicators-for-message-identification/

Closes: https://github.com/roundcube/roundcubemail/issues/8143

### DIFF
--- a/plugins/bimi/README.md
+++ b/plugins/bimi/README.md
@@ -1,0 +1,23 @@
+Roundcube Webmail BIMI Plugin
+=============================
+This plugin will use [Brand Indicators for Message Identification (BIMI)](https://bimigroup.org/) icons contact icons if a contact doesn't otherwise have an icon.
+
+Brand Indicator Message Identification (BIMI) is an industry-wide standards effort to use brand logos as indicators to help email recipients recognize and avoid fraudulent messages.
+
+[IETF: Brand Indicators for Message Identification (BIMI)](https://datatracker.ietf.org/doc/draft-brand-indicators-for-message-identification/)
+
+License
+=======
+This plugin is released under the GNU General Public License Version 3
+or later (http://www.gnu.org/licenses/gpl.html).
+
+Install
+=======
+* Place this plugin folder into plugins directory of Roundcube
+* Add bimi to $config['plugins'] in your Roundcube config
+
+Config
+======
+The default config file is plugins/bimi/config.inc.php.dist
+Rename this to plugins/bimi/config.inc.php
+
--- a/plugins/bimi/bimi.php
+++ b/plugins/bimi/bimi.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * BIMI
+ *
+ * Plugin to display Display Brand Indicators for Message Identification (BIMI) icons
+ * for contacts/addresses that do not have a photo image.
+ *
+ * @license GNU GPLv3+
+ * @author Craig Andrews <candrews@integralblue.com>
+ * @website http://roundcube.net
+ */
+class bimi extends rcube_plugin
+{
+    public $task = 'addressbook';
+
+
+    /**
+     * Plugin initialization.
+     */
+    function init()
+    {
+        $this->add_hook('contact_photo', [$this, 'contact_photo']);
+    }
+
+    /**
+     * 'contact_photo' hook handler to inject a bimi image
+     */
+    function contact_photo($args)
+    {
+        // pre-conditions, exit if photo already exists or invalid input
+        if (!empty($args['url']) || !empty($args['data'])
+            || (empty($args['record']) && empty($args['email']))
+        ) {
+            return $args;
+        }
+
+        $rcmail = rcmail::get_instance();
+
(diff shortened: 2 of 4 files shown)


candrews:
> Rely on the mail-headers.

Which mail header? I don't think there is one that specifies the necessary BIMI information.

pep-un:
Hello, 

It seams needed to query live DNS record for some information (as the policy) not present in the header. 
is some teasting needet on this PR ? I will be happy to help for that part.

Regards,

LauraTaylorUK:
It would be great to support BIMI without requiring paid certification of course :)

I looked at the plugin code and its missing all the validation steps. Off the top of my head, the plugin should at least perform the following things:


**Authentication requirements**

- If more than one RFC5322.From headers are present or if there is one RFC5322.From header but with multiple addresses, then BIMI processing must not be performed.

- The DMARC result must be a pass or a quarantine with a percentage tag of 100% (pct=100) else BIMI processing must not be performed.


**Assertion Record Discovery**

- Check for the BIMI selector header, use it if it exists otherwise use 'default' as the selector.

- Records must start with a "v=" tag and current version else BIMI processing must not be performed.


**Indicator Discovery Without Evidence**

- The TLS server certificate must be valid and trusted else BIMI processing must not be performed.


**Indicator Validation**

- Check the size of the downloaded file and if found too big then BIMI processing must not be performed.

- Check the file format is SVG or SVGZ and if not properly formatted then BIMI processing must not be performed.

alecpl:
I agree it needs more validation. Also maybe `brandicon` would be a better name for the plugin. Finally, the code style needs to be fixed according to our new rules.

Neustradamus:
@candrews: Can you update your PR with latest BIMI draft version?
- https://datatracker.ietf.org/doc/html/draft-brand-indicators-for-message-identification

InputOutputZ:
I just had thought about this implementation and @ghost comment and recent draft, from user experience point of view which doesn't break the draft standards, I don't think it looks the best when loading multiple avatars for email account and displaying them, it just feels better if there is just one avatar, using following logic, if there is bimi present from the recipient message, current avatar should be replaced with bimi, and then alternative perhaps using gravatar or roundcube addressbook avatars. 

Here are a relevant snapshots attached below from the bimi draft expiring on the 8th of May 2025.
 
It looks like the indicator can be displayed without evidence, as long as the uri with https transport call was made to retrieve the SVG/SVGZ, Also, size limit is not an issue, and receiver can proceed to display SVG as long as its valid even without evidence.

 Screenshot 2025-02-27 at 20 55 14 

Here it touches on personal avatars and BIMI, and it looks like if tag s=bimi is not present, then it should treated as the sender require BIMI to have preference while if was set to personal, then avatar should be used instead of BIMI, therefore its per this PR implementation I think its considered valid BIMI integration, since we still have choice to display personal avatar over BIMI, i.e. don't participate in BIMI whenever we want.

Screenshot 2025-02-27 at 20 56 14
Screenshot 2025-02-27 at 20 56 30

Lastly, BIMI processing is performed at MTA level and the only thing Roundcube needs to do either need to check message headers, which can be considered the secure way to fetch the SVG and decide whether to display it or not, when BIMI SVG is getting hashed and its hash truncation included in AR headers, otherwise retrieving SVG from DNS records with https call, its too considered safe perhaps safest since this bypasses the risk of message headers modification, and should be considered as the most valid BIMI source, and finally the SVG needs just to be validated and filtered from any other non-SVG tags and per comments, it looks like displaying the SVG via img tag src attribute is safer than displaying it via raw SVG element tag.
Screenshot 2025-02-27 at 21 39 56

Also, this bimi validator could be useful, I dont have experience writing code in perl and it would be great if anyone can convert it to PHP package.
https://github.com/marcbradshaw/bimi-validator
https://bimivalidator.authmilter.org

Also, this package, it might be helpful
https://github.com/biohzrdmx/bimini-php

I recognise that many of the MTAs have not integrated BIMI up until now and difficult to test, but I hope to see this PR closed, with fully rounded plugin ready to include as part Roundcube anytime soon.

sarabjeet121:
its working 
Screenshot 2026-03-21 170416

Neustradamus:
Some people ask news about BIMI:
- https://github.com/roundcube/roundcubemail/issues/8143

PR #8887, opened 2023-01-25
add identity management script

vladasko-g:
A shell script used to manage identities.

use case - automation that aims to synchronize roundcube settings with mailserver settings

### DIFF
--- a/bin/identity.sh
+++ b/bin/identity.sh
@@ -1,0 +1,469 @@
+#!/usr/bin/env php
+
+<?php
+/*
+ +-----------------------------------------------------------------------+
+ | This file is part of the Roundcube Webmail client                     |
+ |                                                                       |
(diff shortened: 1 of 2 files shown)


vladasko-g:
not 100% sure how identities_level should be managed so I try to be strict, any comments would be useful

vladasko-g:
I didn't update all other places where get_host might possibly be used as I'm not really comfortable with all other places of the app

vladasko-g:
I updated the code according to @alecpl comments in the former PR https://github.com/roundcube/roundcubemail/pull/8839 it would be great to have more feedback

matejak:
Hello @alecpl and perhaps @thomascube, could you please check this PR out? I also thing that it is useful, especially when it comes to the need of advanced automated administration.
Are you aware of any bits or perhaps documentation/rationale that is missing?

Issue #7742, opened 2020-11-24
Select addresses from a group when composing


rdacn:
Hi
we have a problem that might not be very common
Our users are essentially teachers
They import  pupils in their contacts using csv, from a student management software 
Every pupil has 2 mail addresses (mother and father, or parent1 and parent2), except in a few cases
Teachers then create groups (one class = one group)
But when they want to send a message to the whole group, they go to "insert contact(s)" they search for the group, they select the group and... only one address per pupil is retrieved (presumably the first one in the list)

The only solution I found is to tell them to edit each pupil contact entry and add a keyword (name of the class or group) to its first or last name, then to search in contacts with this keyword : this way, they can select all addresses

The real solution would not be to retrieve all addresses from every contact when selecting a group, as it would obviously lead to duplicates, but simply, when searching a group in 'insert contact(s)' , to display all addresses from this group and let the user select which ones (s)he wants to choose.

Thanks

alecpl:
Most likely what should happen here is that 'group-expand' action should consider "Skip alternative email addresses in autocompletion" option. I.e. it should use all email addresses when `autocomplete_single=false` (default), but it would use one address when `autocomplete_single=true`.

So, it could be argued that this actually is a bug.

rdacn:
That's true for auto-completion, but I'd like to point out that my primary target was not auto-completion but the menu that appears when you click on the "Add/insert contact(s)" button on the right.

In this menu, when you search for a group, you can only select the group itself. A better option would be to display the group, but also, each individual entry belonging to the group, with all its addresses, therefore providing a way to select the group by clicking it (as for now) OR select each individual address.

clesmian:
We are facing a similar issue. After looking for a solution I stumbled upon this issue. 

I implemented a quick solution, which solves the problem of sending emails to multiple guardians of a child for us. I also created a PR. Maybe it also helps in your case.

If there are any comments, I'm always open to suggestions.

Neustradamus:
@rdacn: Have you tested the @clesmian PR?

gthieleb:
We are also facing the same issue. Teacher that saving two mail adresses in one contact can only select one. @rdacn where you able to test this PR?

PR #8906, opened 2023-02-02
Add option to add all addresses of contacts to recipients when their group is added

clesmian:
As discussed in issue #7742 it is difficult to add all email-addresses of contacts to the recipients when they are organized in groups. To make it easier this PR implements an option to expand all addresses of all contacts in a group when the group is added to the recipients of an email.


Neustradamus:
To follow this PR.

PR #8916, opened 2023-02-10
[managesieve] allow custom USER-scriptname in kolab-mode

slangenhorst:
Hello,

we have an kolab-like sieve enviroment, where there's an generated sieve-script file which then includes the users own script.
With this patch the hardcoded USER-Scriptname can be customized, so that users can use all the roundcube sieve-features and we still have the company-generated in place and activated.


Neustradamus:
@slangenhorst: Why have you closed your PR?

Issue #7141, opened 2020-01-03
Elastic: delete button too far from list of mails


bes-internal:
I think many people are familiar with the situation when you can delete a letter-notification without even reading it. Now it is difficult to make it using mouse: you need to select the letter, find toolbar on the right side, delete it and then return to the list of letters.

![0](https://user-images.githubusercontent.com/2046529/71722777-31b33e00-2e3b-11ea-8e28-1277c0d33be8.png)

I suggest the following improvements:
1. Add delete button to the top of the list of letters to delete an opened letter 
2. Add delete on hover to delete without reading

![1](https://user-images.githubusercontent.com/2046529/71722850-82c33200-2e3b-11ea-8cc6-e23970e77708.png)

alecpl:
1. You can select message(s) with click and CTRL key to select message(s) without opening.
2. You can install contextmenu plugin.

Anyway, maybe we should consider other options, but I don't really like proposed options.

manfredu:
My opinion: Two Delete buttons with the same functionality are confusing. You can also use DEL key from the keyboard for fast action.

Also there are more buttons on the right column above the open mail which work on a list of selected mails (middle column)...

I do agree that the placement of the buttons is a bit confusing in the beginnung, but I got used to it.

The only other option I can imagine of would be to have all buttons together in one place. But I would leave it as it is.

bes-internal:
consider the logic of a similar interface
![2020-01-10_14-23](https://user-images.githubusercontent.com/2046529/72149841-17d2a780-33b5-11ea-8218-3cc6b927e692.png)

manfredu:
So the buttons on the right side affects only the currently displayed mail? 

This means all buttons which can affect several selected mails would need to be copied to the middle column (i.e. forward, delete mark, more... ).

manfredu:
Or have a row of buttons, which can affect one or more mails, across both columns and move buttons, which can be only used for the currently open mail, into the window of the displayed mail... Similiar to above screenshot.

bes-internal:
Yes. On above screenshot this is first button for forward. 
The meaning of any UI is to reduce the time and number of gestures of the most popular functions.
Let's do delete at least on hover of message in Elastic. Mark or delete. This is most popular quick functions.

P.S. I speak not only from myself. But from a significant number of our customers who use the new interface. I have already received more than 20 applications about the inconvenience of deleting in the new interface.

alecpl:
[...]

PR #8930, opened 2023-02-19
add delete field to message list

johndoh:
Inspired by #7141

Untitled-1

Currently when hovering over a message in the list, for messages which are not flagged, a flag icon is displayed in the top right corner of the row. Also the message date is replaced with the message size.

This PR adds a new delete column to the message list so that in widescreen view a delete icon can be shown in additional to the flag icon. This delete icon is shown directly below the flag icon. For messages which have attachments the attachment icon is replaced by the delete icon on hover.

*This is a breaking change*.
A new span for the delete icon is added into the `flags` \<td\>.  I can create PRs for the legacy Classic and Larry skins to prevent display issues.

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -389,7 +389,7 @@ class rcmail_action_mail_index extends rcmail_action
         $rcmail->output->set_env('sort_order', $_SESSION['sort_order']);
         $rcmail->output->set_env('messages', []);
         $rcmail->output->set_env('listcols', $listcols);
-        $rcmail->output->set_env('listcols_widescreen', ['threads', 'subject', 'fromto', 'date', 'size', 'flag', 'attachment']);
+        $rcmail->output->set_env('listcols_widescreen', ['threads', 'subject', 'fromto', 'date', 'size', 'flag', 'attachment', 'delete']);
 
         $rcmail->output->include_script('list.js');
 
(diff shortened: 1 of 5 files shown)


alecpl:
The hover effect on the icon is wrong: i.e. the color should not be red and the icon should not change. Also make sure to use the same icon as in the toolbar.

I wonder whether, after we merge this, it would make sense to use this new icon/position for the Undelete action (now we're using the icon on the left of the subject line). I guess it would have pros and cons.

johndoh:
I have changed the delete icon to match the one in the toolbar exactly and removed the hover effects I added.

I have also added the restore icon for undelete action

alecpl:
@johndoh I did not forgot about this. The solution I had in mind was something like this:

hover-menu

It will take time for me to implement this though. I think it will have some adventages.

Plugins could add icons there easier (of course too many might be problematic, in future we may make this configurable). I removed the icon/button from the subject line, maybe we could also remove the flag icon/button at all and keep only the red color as a flag indication.

For keyboard-only users we'd need a way to display the menu, maybe using Right-Arrow and Left-Arrow keys.

Potentially we could put folder name in this menu, but it might not look that good. Something to investigate.

Then it might be a good idea for your swipe plugin's default behavior. Swipe left displays this menu. Swipe right hides it.

johndoh:
I like the idea of a "message actions" field where the actions in the field could be customised.

I think the flag icon should stay, people are used to seeing that in mail clients and it might not be so obvious what only highlighting a message in red means, also for accessibility with colour blind people.

While the Elastic skin does not support the desktop/list view I think this also needs to be considered as we know some users like that display, we should not deprecate the functionality.

PR #8902, opened 2023-01-31
Fix a small typo in el_GR/labels.inc

ghost:
I just happened to notice this tiny typo on a local installation of Roundcube. I just sent a request to join the Transifex organisation to fix this there as well and help with the Greek translation. 

If typo fixes should exclusively be done via Transifex as well, feel free to close this PR instead of merging.

PR #8946, opened 2023-03-15
Typo fix in password plugin and password plugin driver interface extended

szabozs:
There was a typo in $length variable in minimal lenght message for password.


bandisast:
Greetings,

From previous experience with a [similar](https://github.com/roundcube/roundcubemail/pull/8902) issue, any and all localisation-related typo fixes must first be published on Transifex and *then* be transferred to the github repo. Did you fix the same exact typo on Transifex as well? If not, your patch would be overwritten if someone pushed new strings directly from Transifex.

ghost (changes requested):
Greetings,

From previous experience with a [similar](https://github.com/roundcube/roundcubemail/pull/8902) issue, any and all localisation-related typo fixes must first be published on Transifex and *then* be transferred to the github repo. Did you fix the same exact typo on Transifex as well? If not, your patch would be overwritten if someone pushed new strings directly from Transifex.

alecpl:
I fixed the issue at Transifex. As for the other part of this PR... there's no need to extend the API, you can get username from a driver like this:
```
$username = rcmail::get_instance()->get_user_name();
```

Issue #7950, opened 2021-03-16
Feature request: search results showing the folder each result is in


PenelopeFudd:
Hi;

When I search all my folders for messages, I can find them, but I have no idea which folder they're in.  Would it be easy to append the folder name to the message header and/or add it to the search results?

Thanks.

thomascube:
There's an extra column in Larry there but the list view in Elastic is different.

alecpl:
Here's some options:
1. Display the folder name in an extra line between sender/date and subject
2. Replace sender/recipient with folder name on hover event. Just like we replace date with size.
3. Use the folder name as tr's title attribute.
4. Extend the contextmenu plugin to display message information (like folder name, size, something else?) in the context menu (on right click). Or implement our own context menu and allow the plugin to extend it.

I don't like the option 1. Option 3. would be simplest to implement, but because sender and subject have it's own title it might not be that good solution. Option 4. might be tricky. So, probably option 2. is the winner.

ps. only option 1. would work on mobile, but I think that we can agree that mobile UI does not have to have all features.

manfredu:
Currently I use the following workaroud: Moving the mouse cursor to the subject of a mail from the result list displays a url at the bottom with _action=show. This url also contains the _mbox parameter with the folder name!

Even with your suggestion it is still confusing that if a message from the search result list is shown in the third column, the selected folder in the first column may not match and it is not clear to which folder the mail belongs.

alecpl:
We implmented the title attribute-based solution.

PR #8956, opened 2023-03-25
include source folder name in message details when multi folder search acitve

johndoh:
possible solution for https://github.com/roundcube/roundcubemail/issues/7950

adds a "Folder" entry to the message details (To, From, Date, etc) show at the top of the message display when a multi folder search is active. Only visible in the details (expanded) view, no the summary.


antiuser:
@johndoh hi! will be great if you do this in search results :)

PR #8975, opened 2023-04-08
markasjunk: add SpamAssassin v4 compatibility

johndoh:
Add compatibility with new pref names in SAv4 to the sa_blacklist driver for markasjunk plugin

### DIFF
--- a/plugins/markasjunk/drivers/sa_blacklist.php
+++ b/plugins/markasjunk/drivers/sa_blacklist.php
@@ -52,6 +52,10 @@ class markasjunk_sa_blacklist
         $this->sa_preference_field = $rcube->config->get('sauserprefs_sql_preference_field');
         $this->sa_value_field      = $rcube->config->get('sauserprefs_sql_value_field');
 
+        // SAv4 compatibility
+        $blocklist_pref_name       = $rcube->config->get('sauserprefs_sav4', false) ? "blocklist_from" : "blacklist_from";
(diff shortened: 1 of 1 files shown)


plugins/markasjunk/drivers/sa_blacklist.php
```diff
@@ -52,6 +52,10 @@ private function _do_list($uids, $spam)
         $this->sa_preference_field = $rcube->config->get('sauserprefs_sql_preference_field');
         $this->sa_value_field      = $rcube->config->get('sauserprefs_sql_value_field');
 
+        // SAv4 compatibility
+        $blocklist_pref_name       = $rcube->config->get('sauserprefs_sav4', false) ? "blocklist_from" : "blacklist_from";
+        $welcomelist_pref_name     = $rcube->config->get('sauserprefs_sav4', false) ? "welcomelist_from" : "whitelist_from";
```
alecpl, line 57: This will be a BC break. I think we should make the old options the default.

alecpl (changes requested):

plugins/markasjunk/drivers/sa_blacklist.php
johndoh: `False` is the default result and `blacklist_from` is the default pref name so I don;t think it is a break. SAv4 supports the original blacklist_from and the new blocklist_from so if users do add the new `sauserprefs_sav4` they get blacklist_from which still works no different to now.

alecpl: Sorry, I've read this code incorrectly. You're right.

PR #8983, opened 2023-04-19
add signature template to new_user_dialog plugin

CvH:
This adds the possibility to define a default signature. Additionally the signature email gets replaced with the user email if `@EMAIL@` get added at the signature.

To configure the default signature add this to the `config/config.inc.php`

`$config['new_user_dialog_signature_text'] = ' ... ';`

Just for full transparency, I have no PHP skills. So while it works perfectly for me I have no real idea if this is "properly" done.

### DIFF
--- a/plugins/new_user_dialog/new_user_dialog.php
+++ b/plugins/new_user_dialog/new_user_dialog.php
@@ -19,9 +19,12 @@ class new_user_dialog extends rcube_plugin
 
     function init()
     {
+        $rcmail = rcmail::get_instance();
+        $this->load_config();
         $this->add_hook('identity_create', [$this, 'create_identity']);
         $this->add_hook('render_page', [$this, 'render_page']);
         $this->register_action('plugin.newusersave', [$this, 'save_data']);
+        $this->signature_text = $rcmail->config->get('new_user_dialog_signature_text');
     }
 
     /**
@@ -89,6 +92,7 @@ class new_user_dialog extends rcube_plugin
                     'name' => '_signature',
                     'rows' => '5',
                 ],
+                $out = str_replace('@EMAIL@', rcube_utils::idn_to_utf8($identity['email']) , $this->signature_text ?? ''),
                 $identity['signature']
             ));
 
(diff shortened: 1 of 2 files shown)


Neustradamus:
@CvH: Interesting!

@alecpl: What do you think?

plugins/new_user_dialog/new_user_dialog.php
```diff
@@ -19,9 +19,12 @@ class new_user_dialog extends rcube_plugin
 
     function init()
     {
+        $rcmail = rcmail::get_instance();
+        $this->load_config();
```
alecpl, line 23: FOr better performance loading config file should be done only when it's needed, i.e. not in init().

plugins/new_user_dialog/new_user_dialog.php
```diff
@@ -19,9 +19,12 @@ class new_user_dialog extends rcube_plugin
...
         $this->add_hook('identity_create', [$this, 'create_identity']);
         $this->add_hook('render_page', [$this, 'render_page']);
         $this->register_action('plugin.newusersave', [$this, 'save_data']);
+        $this->signature_text = $rcmail->config->get('new_user_dialog_signature_text');
```
alecpl, line 27: non-existing property warning. This can be a normal variable if you move it to render_page().

plugins/new_user_dialog/new_user_dialog.php
```diff
@@ -89,6 +92,7 @@ function render_page($p)
                     'name' => '_signature',
                     'rows' => '5',
                 ],
+                $out = str_replace('@EMAIL@', rcube_utils::idn_to_utf8($identity['email']) , $this->signature_text ?? ''),
                 $identity['signature']
```
alecpl, line 96: This is wrong. If signature already exists (which could be set by another plugin) it should be used. $out variable is not needed here.

Also, it was missing before, but I think the content need to be encoded with rcube::Q().

alecpl (changes requested):

Neustradamus:
@CvH: Have you progressed on it?

CvH:
tried to address the comments for change, like said I am not a php dev so pls review it with a grain of salt :)
I didnt rebased against master yet and just pushed a additional commit to address the changes - if this is correct it would rebase.

PR #9005, opened 2023-05-08
markasjunk's cmd_learn driver can't find some headers

nsoggia:
I added a few escapeshellarg() here and there just for good practice. I couldn't think of a POC for a RCE, but one day someone else could find one.

The real point of this PR is that the %h:headername argument works only with some headers because rcube_imap_generic::fetch() sometimes returns a key that is equal to the one in the query and sometimes it doesn't.
For example, the RFC header "message-id" becomes "messageid", so
using "%h:message-id" the header is found by fetch() but _do_salearn() can't find the answer,
using "%h:messageid" the header is not found by fetch().
Either way you don't get what you deserve.


alecpl:
Thanks for the PR. I just merged the security part separately. I'll see if I can find a better fix for the headers issue.

nsoggia:
It works, and your solution is much more clever than mine

alecpl:
Fixed.

PR #9038, opened 2023-06-25
create template object for message list actions

johndoh:
I acknowledge that 9427ec1d35445170cd5d25b3aae9b69950610c3b is very new, I thought it might be helpful to setup a place to discus the possibilities of this new menu. I think the idea is great, better than adding extra columns for every action a user wants.

The first commit in this PR moves creation of the menu from the Elastic JS code into the core so in the future other skins could benefit from it and also plugins. Note: there is more to do, if Roundcube wants to go this way, more of the JS for the menu needs to move into `app.js` or may be `list.js` so it becomes a core function rather than a skin one.

There is a lot more to do, I wanted to post this draft now to start the discussion and get feedback on if its worth working on.

Unrelated to the organization of the code I have also been playing with the look of the menu. At the moment I have this:

Untitled-1

- Position is moved to the top right corner of the message row
- The message row border color is used as a background color to try and make it look like a tab coming down from the corner of the row
- The size of the icons has been reduced and the text and icons have the same size so its not so intrusive

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -719,14 +719,14 @@ class rcmail_action_mail_index extends rcmail_action
         $config = $rcmail->config->get('message_actions', ['size', 'date', 'flag', 'delete']);
 
         $actions = [
-            'size'      => null,
-            'date'      => null,
-            'read'      => ['label' => 'markasread', 'action' => 'mark/read'],
-            'unread'    => ['label' => 'markasunread', 'action' => 'mark/unread'],
-            'flag'      => ['label' => 'markasflagged', 'action' => 'mark/flagged'],
-            'unflag'    => ['label' => 'markasunflagged', 'action' => 'mark/unflagged'],
-            'delete'    => ['label' => 'deletemessage', 'action' => 'delete'],
-            'undelete'  => ['label' => 'undeletemessage', 'action' => 'mark/undelete'],
+            'size' => null,
+            'date' => null,
+            'read' => ['label' => 'markasread', 'action' => 'mark/read'],
+            'unread' => ['label' => 'markasunread', 'action' => 'mark/unread'],
+            'flag' => ['label' => 'markasflagged', 'action' => 'mark/flagged'],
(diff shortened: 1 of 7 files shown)


alecpl:
I'm not sure I like `messagelistactions` part and I'd also like to have normal size icons in the menu.

johndoh:
> I'm not sure I like messagelistactions part

You don't like the name or your don't like having a template object for it?

> I'd also like to have normal size icons in the menu.

Understood. Nothing in this PR changes the look, the image I posted was created with `_styles.less` and only for the purposes of discussion.

johndoh:
I am now looking at the idea of allowing users to customise the actions available in the hover menu via their settings. I was thinking about something like this:

Untitled-2

And use drag & drop to allow users to customise the order of the icons (with some hint text inside the box to inform users).
But first I think some decisions need to be made about if/how this feature will be implemented, can it go into the core or should it a plugin?

Either way there are some other details like the skin needs to somehow register that it supports the object so we know to show the settings section. May be this could be done in the config section of the meta file like with the layouts.

I think there are 3 options:

1. The current implementation (in master branch). This is Elastic skin specific.
2. This PR, add a template object to allow all skins to use the same code, easier extensibility to option 1. Styling stays in the skin just as it is in option 1.
3. Hybrid option. Move the code this PR puts into the core into a core plugin. Styling stays in the skin just as it is in option 1.
4. Third party plugin. Move everything to a third party plugin.

I think I prefer 2 or 3 but I am happy work on any of them. I think the hover menu is a good solution, it adds useful functionality to the mail screen and will be useful in any skin with widescreen layout. In developing this PR I have also been trying things with different skins and it does work in list mode as well – rather than taking up screen space with an additional column for more options the options only come up on hover. I’m also trying to make it a little more generic at least on the JS side, so similar menus could easily be added to other lists if required.

alecpl:
As for the configuration, I would put it in "List options" dialog instead. I would not bother with order/drag-n-drop/icons.

johndoh:
i have simplified this PR, the only aim now is to create a template object for the list actions widget so that plugins and extended skins can interact with it easier.

PR #9057, opened 2023-07-09
Fix sort_folder_comparator

adrienbeau:
In case $path2 is longer than $path1, and all folders in $path1 are
equal to the corresponding folder in $path2 (in other words: $path2 is a
subfolder of $path1), sort_folder_comparator currently considers them
equal because it stops comparing when it reaches the end of $path1.

This breaks the ordering done by uasort() above, and breaks code that
relies on the folder list being sorted, notably the folders settings
page.

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -4517,7 +4517,10 @@ class rcube_imap extends rcube_storage
         $path1 = explode($this->delimiter, $str1);
         $path2 = explode($this->delimiter, $str2);
 
-        foreach ($path1 as $idx => $folder1) {
+        $len = max(count($path1), count($path2));
+
+        for ($idx = 0; $idx < $len; $idx++) {
+            $folder1 = $path1[$idx] ?? '';
             $folder2 = $path2[$idx] ?? '';
 
             if ($folder1 === $folder2) {


alecpl:
Could you provide a sample list of folders input that is not sorted properly?

adrienbeau:
Unfortunately, my current list of folders reveals a bit too much about me. I will try to reproduce the issue with a made-up list.

Another way to highlight the issue is to shuffle $folders in sort_folder_list right before calling uasort. Here's how I've done it:

--- ~/software/roundcubemail-1.6.2/program/lib/Roundcube/rcube_imap.php     2023-07-01 13:30:01.000000000 +0200
+++ /opt/roundcube/program/lib/Roundcube/rcube_imap.php 2023-07-09 09:58:28.011245392 +0200
@@ -4418,6 +4418,7 @@
 
         // sort folders
         // asort($folders, SORT_LOCALE_STRING) is not properly sorting case sensitive names
+        $this->shuffle_assoc($folders);
         uasort($folders, [$this, 'sort_folder_comparator']);
 
         $folders = array_keys($folders);
@@ -4531,6 +4535,24 @@
     }
 
     /**
+     * Added to hightlight the incorrect sort_folder_collator
+     */
+    protected function shuffle_assoc(&$array) {
+        rcube::write_log('plop', '### before shuffle array:');
+        rcube::write_log('plop', $array);
+        $keys = array_keys($array);
+        shuffle($keys);
+        $new = array();
+        foreach($keys as $key) {
+            $new[$key] = $array[$key];
+        }
+        $array = $new;
+        rcube::write_log('plop', '### after shuffle array:');
+        rcube::write_log('plop', $array);
+        return true;
+    }
+
+    /**
      * Find UID of the specified message sequence ID
      *
      * @param int    $id     Message (sequence) ID

adrienbeau:
Here is a list of folders that triggers the issue on my server.

(By the way: this is on Ubuntu 22.04 with the official Ubuntu packages for PHP 8.1.2 and Dovecot 2.3.16.)

I created a fresh test account and added the following folders manually using the Roundcube folder settings page, in this order:

```
INBOX/G Folder 2
INBOX/S Folder 4 accentué
INBOX/S Folder 4 accentué/B Level 2 Folder 4.1
INBOX/S Folder 4 accentué/H Level 2 Folder 4.2
INBOX/A Folder 1
INBOX/G Folder 2/T Level 2 Folder 2.3
INBOX/S Folder 4 accentué/H Level 2 Folder 4.2/K Level 3 Folder 4.2.1
INBOX/S Folder 4 accentué/H Level 2 Folder 4.2/B Level 3 Folder 4.2.4
INBOX/A Folder 1/C Level 2 Folder 1.3
INBOX/A Folder 1/C Level 2 Folder 1.3/C Level 3 Folder 1.3.2
INBOX/M Folder 3
INBOX/G Folder 2/D Level 2 Foldère 2.4
INBOX/G Folder 2/T Level 2 Folder 2.3/U Level 3 Folder 2.3.1
INBOX/M Folder 3/A Folder 3.3 Level 2
INBOX/G Folder 2/T Level 2 Folder 2.3/H Levèle 3 Folder 2.3.3
```

Here is the resulting folder list in the settings view:
Screenshot_20230709_160846_folders_in_settings_view

_M Folder 3_ is greyed out and disabled in the screenshot because it is considered virtual. It is considered virtual because the folder _INBOX/M Folder 3/A Folder 3.3 Level 2_ was encountered before _INBOX/M Folder 3_ in the "sorted" folder list.

Depending on how folders are "sorted", more breakage can be visible. The biggest one I've seen is INBOX being considered virtual, and other inbox subfolders getting displayed as children of the Drafts folder (even though their parent folder was still correctly shown as being INBOX in the folders detail pane). In that case, I also had subfolders shown as children of the wrong folder.

The breakage is rather unpredictable, adding one more folder to my example gives us a working list again. The breakage is consistent however, removing that additional folder gives us the broken list again. This is not surprising, since PHP lists are hash tables. This also means that my example may not work on your machine, it may depend on the exact PHP compiled binary.

I managed to reproduce the breakage with another set of test folders. It seems the breakage only starts to be visible when 15 folders have been added.

Here is the contents of the $out variable right at the end of sort_folder_list. This confirms line 15 is not at the correct place:

```
[09-Jul-2023 16:35:59 +0200]: <kiccbjm0> ### out:
[09-Jul-2023 16:35:59 +0200]: <kiccbjm0> array (
  0 => 'INBOX',
  1 => 'INBOX/A Folder 1',
  2 => 'INBOX/A Folder 1/C Level 2 Folder 1.3',
  3 => 'INBOX/A Folder 1/C Level 2 Folder 1.3/C Level 3 Folder 1.3.2',
  4 => 'INBOX/G Folder 2',
  5 => 'INBOX/G Folder 2/D Level 2 Fold&AOg-re 2.4',
  6 => 'INBOX/G Folder 2/T Level 2 Folder 2.3',
  7 => 'INBOX/G Folder 2/T Level 2 Folder 2.3/H Lev&AOg-le 3 Folder 2.3.3',
  8 => 'INBOX/G Folder 2/T Level 2 Folder 2.3/U Level 3 Folder 2.3.1',
  9 => 'INBOX/M Folder 3/A Folder 3.3 Level 2',
  10 => 'INBOX/S Folder 4 accentu&AOk-',
  11 => 'INBOX/S Folder 4 accentu&AOk-/B Level 2 Folder 4.1',
  12 => 'INBOX/S Folder 4 accentu&AOk-/H Level 2 Folder 4.2',
  13 => 'INBOX/S Folder 4 accentu&AOk-/H Level 2 Folder 4.2/B Level 3 Folder 4.2.4',
  14 => 'INBOX/S Folder 4 accentu&AOk-/H Level 2 Folder 4.2/K Level 3 Folder 4.2.1',
  15 => 'INBOX/M Folder 3',
  16 => 'Drafts',
  17 => 'Sent',
  18 => 'Junk',
  19 => 'Trash',
)

```

alecpl:
I wasn't able to reproduce and create a test case that would fail. Maybe PHP 8.1.2 (quite old) has a bug, maybe it's related to the fact that in some cases `sort_folder_comparator()` can return null, but it should always return an integer.

The change and your explanation makes sense, so I'll merge this.

PR #9074, opened 2023-07-19
Managesieve plugin: add spam settings engine

frozen-sky:
This PR adds in a similar way as the out-of-office and forward engines, the option to render a more sophisticated/specific form to set spam filtering values. It allows to configure enable/disable spam-filtering and adjust the threshold when filtering should happen.
There is an option to add ACL Allow list entries, which will allow emails to skip the filtering action.

Configuration defaults and options are added to the managesieve configuration example.

As for localization: It is unknown to me how RoundCube works with localizations. For now I did not touch the localizations except English (US). I can do an effort to fill in most translations, however, I do not speak all the languages in the table


plugins/managesieve/tests/Forward.php
```diff
@@ -10,9 +10,9 @@ static function setUpBeforeClass(): void
     }
 
     /**
-     * Test vacation_form()
+     * Test forward_form()
```
frozen-sky, line 13: When making the test, i saw this was incorrect labelled in the Forward test, as such corrected this. Its not related to the goal of this PR

plugins/managesieve/tests/Spam.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        $rcube  = rcube::get_instance();
+        $output = $this->initOutput(rcmail_action::MODE_HTTP, 'settings', 'managesieve');
+
+        $plugin  = new managesieve($rcube->plugins);
+        $spam = new rcube_sieve_spam($plugin);
+
+        setProperty($spam, 'spam', ['list' => []]);
+
+        $result = $spam->forward_form([]);
```
alecpl, line 25: ->spam_form().

plugins/managesieve/lib/Roundcube/rcube_sieve_spam.php
```diff
@@ -0,0 +1,528 @@
...
+                        'not' => true,
+                        'arg1' => 'from',
+                        'arg2' => $item,
+                    ];
+                }
+
+            }
+
+            $rule['type']       = 'if';
+            $rule['name']       = "Spam filter settings";
```
alecpl, line 268: This needs to be localized.

alecpl (changes requested):

Neustradamus:
@frozen-sky: Have you progressed on your PR?

frozen-sky:
> @frozen-sky: Have you progressed on your PR?

Thank you for reminding me. The local version is a bit progressed but not finalized yet. I will try to push a bit on it next week

pabzm:
@frozen-sky Are you still interested in this PR?

PR #9081, opened 2023-07-30
Config option to save drafts silently

### LINKED ISSUE #8989: Feature Request, Save Draft Silently (without notification that its being saved)
I recently had somebody have trouble with losing an email that they had spent time typing.

I changed the save draft frequency to 1 minute, so that they would not lose what they had worked on typing.

The problem is that they are now getting annoyed by the popup message that says saving to drafts....

Could we get an option to save to drafts without the popup message? like a silent mode.

The option could go just below the save frequency.

Jieiku:
This adds a config option to save drafts silently.

closes #8989

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -1344,6 +1344,9 @@ $config['prettydate'] = true;
 // save compose message every 300 seconds (5min)
 $config['draft_autosave'] = 300;
 
+// save compose message silently
+$config['draft_autosave_silent'] = false;
+
 // Interface layout. Default: 'widescreen'.
 //  'widescreen' - three columns
 //  'desktop'    - two columns, preview on bottom
--- a/program/actions/mail/send.php
+++ b/program/actions/mail/send.php
@@ -272,8 +272,10 @@ class rcmail_action_mail_send extends rcmail_action
                         'folder' => $store_target
                 ]);
 
-                // display success
-                $rcmail->output->show_message(!empty($plugin['message']) ? $plugin['message'] : 'messagesaved', 'confirmation');
+                if ($rcmail->config->get('draft_autosave_silent') != true) {
+                    // display success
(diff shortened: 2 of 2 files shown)


Jieiku:
it should be good to go.

pabzm:
@Jieiku Thank you for the contribution and the instant feedback on the reminder!

@alecpl You commented in the linked issue that you would prefer to not introduce another option. I can share that feeling, but on the other hand is here a simple PR ready to merge that solves a problem for at least six people, so I would merge it.
Do you have an opinion on it? (If I don't hear from you, I'll act in two weeks.)

/remind me to act on this in two weeks

pabzm:
@alecpl This is neither "silly" nor a "burden". Please be more considerate with language to not alienate contributors.

alecpl:
You're right, this PR is in the "trivial" category. It would be silly to accept this change willy-nilly. Another person will come and say it should be a user preference, let's add it do the Preferences UI.

As for "burden", well, probably too heavy of a word, indeed, but I'm not a native English speaker, so forgive me. Every line of code, and especially every config option increases technical debt that someone has to support.

pabzm:
> I'm not a native English speaker, so forgive me

It's alright. In good old [netiquette](https://datatracker.ietf.org/doc/html/rfc1855#page-3) terms, let's  "be conservative in what you send and liberal in what you receive", especially since most of us are not English native speakers.

> Another person will come and say it should be a user preference, let's add it do the Preferences UI.

If people take the time to think of a solution, implement it, open a reasonable pull-request and gather several positive reactions, I think we should accept that, too.

> Every line of code, and especially every config option increases technical debt that someone has to support.

That ist true in general, but to be honest: if you bring that argument against such a small and simple PR it very much conveys that you want Roundcubemail to not change. From my point of view *that* behaviour will make Roundube die, not connecting with a lively community and embracing change.

pabzm:
@Jieiku Sorry this is happening in your PR. We'll sort this out elsewehere and come back to you later.

PR #9125, opened 2023-09-05
Fix PHP8 deprecation warning in the markasjunk's "email_learn" driver

sebastka:
```
str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /var/www/rc/plugins/markasjunk/drivers/email_learn.php on line 162
```


alecpl:
Fixed via c38c75c0d4860.

PR #9150, opened 2023-09-21
Fix #5051 by always showing inline attachments in list of attachments

bohwaz:
When a message has attachments with a Content-Id or a Content-Location, they are added as inline, expecting the attachment to show up in the message HTML body.

```
--==_=_650c1f3a7e07d-b7578f07ad
Content-Type: image/png; name="2023-09-21 12_35_30-XXXX Pierre (Membres actifs) - Chromium.png"
Content-Transfer-Encoding: base64
Content-Disposition: inline; filename="2023-09-21 12_35_30-XXX Pierre (Membres actifs) - Chromium.png"
Content-ID: <f_lmt1smiu1>
```

But, in some cases the attachments are not referred to in the HTML body.
And when you are viewing emails as plaintext only, you still won't see the attachments.

It's weird as the message is seen as having attachments in the messages list, but no attachments show up in the message view:

image

This patch always adds the inline attachment in the list of attachements, as is the case in most email software, so that you can see and download the attachments.

### DIFF
--- a/program/lib/Roundcube/rcube_message.php
+++ b/program/lib/Roundcube/rcube_message.php
@@ -939,6 +939,13 @@ class rcube_message
                         && (!empty($mail_part->content_id) || !empty($mail_part->content_location))
                     ) {
                         $this->add_part($mail_part, 'inline');
+
+                        // Always show inline attachements in the list of attachments,
+                        // as sometimes the attachment has a content-id but is not referred to in
+                        // the HTML. Also if you switch to plaintext you won't be able to
+                        // see the attachments. #5051
+                        // https://github.com/roundcube/roundcubemail/issues/5051
+                        $this->add_part($mail_part, 'attachment');
                     }
 
                     // Any non-inline attachment


alecpl:
What other email software? Thunderbird for example does not list such images as attachments when they are used in the HTML body. And I don't like a solution proposed here.

The discrepancy between list and preview is indeed an issue, but for performance reasons we don't do an extensive check on the message to find out whether there's a real attachment or not. We do a simple content-type check instead. I doubt this will change.

bohwaz:
For now, the attachements are just hidden and the user experience is poor and confusing.

I checked with Zimbra, Claws Mail, GMail and Mutt, they all display the attachments and don't leave them hidden like Roundcube does.

Neustradamus:
@alecpl: Any progressed on this @bohwaz PR?

Thanks in advance.

bohwaz:
On my side, I'm using this patch in production since September with success :)

But it is annoying having to apply the patch after each update.

I'm considering pushing the patch in the Debian package if there is no progress here.

Neustradamus:
Dear @roundcube team, @alecpl, @thomascube, @karlitschek, @nextcloud team: Can you look this PR done by @bohwaz which solves an imported ticket from 2012 (more 11 years and 3 months)?

Thanks in advance.

Linked to:
- https://github.com/roundcube/roundcubemail/pull/9150
- https://github.com/roundcube/roundcubemail/issues/5051
- https://github.com/roundcube/roundcubemail/issues/6279
- https://github.com/roundcube/roundcubemail/issues/7117
- https://github.com/roundcube/roundcubemail/issues/7122
- https://github.com/roundcube/roundcubemail/issues/7214
- https://github.com/roundcube/roundcubemail/issues/7293
- https://github.com/roundcube/roundcubemail/issues/7414
- https://github.com/roundcube/roundcubemail/issues/7440
- https://github.com/roundcube/roundcubemail/pull/7441

alecpl:
The proposed solution is wrong.

bohwaz:
> The proposed solution is wrong.

Why?

alecpl:
Because it adds all inline attachments to the attachments list.

bohwaz:
And how is that an issue?

alecpl:
In Roundcube inline images (used in HTML content) aren't listed as attachments. It's been this way always. Precisely "Content-Disposition: inline" makes the message part a non-attachment. What to do with such parts could be discussed, but adding them to the attachments list is wrong. As I said before: Thunderbird for example does not list such images as attachments when they are used in the HTML body.

Unfortunately a better fix is not an easy fix.

bohwaz:
I don't see how giving users more control on the attached files is a bad thing?

Neustradamus:
@pabzm, @alecpl: Have you progressed on a fix?

pabzm:
@Neustradamus Unfortunately not, due to other priorities and a lack of time. If you want to help with the issue, please let it be known in a related, still open issue like https://github.com/roundcube/roundcubemail/pull/9585, and please avoid commenting in closed issues or pull requests. Thank you!

PR #9169, opened 2023-10-15
Treat INBOX case-insensitively

### LINKED ISSUE #9166: Duplicate Inbox folder
When the IMAP server reports the Inbox folder without using all-caps ("INBOX"), Roundcube shows Inbox twice. Note that the IMAP spec says that it has to be handled case-insensitively, as explained here: https://github.com/mjl-/mox/issues/80#issuecomment-1759489304

> For reference, INBOX being case-insensitive has been the case forever. See the mailbox ABNF syntax at the very old https://datatracker.ietf.org/doc/html/rfc1730#page-58
>
> ```
>    mailbox         ::= "INBOX" / astring
>                        ;; INBOX is case-insensitive; other names may be
>                        ;; case-sensitive depending on implementation.
> ```
> 
> And the more recent https://datatracker.ietf.org/doc/html/rfc9051#name-formal-syntax
> 
> ```
> mailbox         = "INBOX" / astring
>                     ; INBOX is case insensitive.  All case variants
>                     ; of INBOX (e.g., "iNbOx") MUST be interpreted as
>                     ; INBOX, not as an astring.  An astring that
>                     ; consists of the case-insensitive sequence
>                     ; "I" "N" "B" "O" "X" is considered
>                     ; to be an INBOX and not an astring.
>                     ; Refer to Section 5.1 for further
>                     ; semantic details of mailbox names.
> ```

In my testing, Dovecot reports INBOX in all caps and Roundcube shows Inbox once, but https://github.com/mjl-/mox reports it as Inbox and Roundcube shows it twice. Please implement case-insensitivity.

mattfbacon:
This should fix #9166.

### DIFF
--- a/plugins/archive/archive.php
+++ b/plugins/archive/archive.php
@@ -411,7 +411,7 @@ class archive extends rcube_plugin
                         'maxlength'     => 30,
                         'folder_filter' => 'mail',
                         'folder_rights' => 'w',
-                        'onchange'      => "if ($(this).val() == 'INBOX') $(this).val('')",
+                        'onchange'      => "if ($(this).val().toUpperCase() == 'INBOX') $(this).val('')",
                         'class'         => 'custom-select',
                 ]);
             }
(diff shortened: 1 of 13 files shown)


alecpl:
I think that more appropriate solution might be to uppercase the inbox in a single place i.e. `rcube_imap_generic::_listMailboxes()`.

mattfbacon:
OK, let's do that :)

PR #9177, opened 2023-10-21
unify message font settings

johndoh:
Inspired by https://www.roundcubeforum.net/index.php/topic,30621.msg77615.html

Move all the font settings for HTML messages to one place so that what is shown on the Composing Messages settings screen and in the HTML editor are the same. Currently this is only true for the font size settings which are hard coded in 2 places.

- create a `fontsize_defs()` function, similar to `font_defs()` to provide the list of font sizes
- pass the result of the `font_defs()` function to TinyMCE so that the font lists are consistent
- add a hidden config option (similar to `editor_css_location` for example) to allow admins to override the default font family and font size lists if they want

### DIFF
--- a/program/actions/settings/index.php
+++ b/program/actions/settings/index.php
@@ -1000,8 +1000,8 @@ class rcmail_action_settings_index extends rcmail_action
                     $blocks['main']['options']['default_font'] = [
                         'title' => html::label($field_id, rcube::Q($rcmail->gettext('defaultfont'))),
                         'content' => html::div('input-group',
-                            $select_font->show($rcmail->config->get('default_font', 1)) .
-                            $select_size->show($rcmail->config->get('default_font_size', 1))
+                            (!isset($no_override['default_font']) && count($fonts) > 1 ? $select_font->show($rcmail->config->get('default_font', 1)) : '') .
+                            (!isset($no_override['default_font_size']) && count($fontsizes) > 1 ? $select_size->show($rcmail->config->get('default_font_size', 1)) : '')
                         )
                     ];
                 }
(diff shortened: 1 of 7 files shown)


alecpl:
Makes sense, but I have some notes:
1. Add/Describe new config options in defaults.inc.php (move default list of fonts and sizes there)
2. What if any of the lists is set to an empty array? Will the TinyMCE selectors got hidden? It would be good if they did.

As I'm not a big fan of heaving the font format selector at all, maybe we should consider setting `font_defs` to be an empty array by default.

Also, for reference #5700.

johndoh:
> What if any of the lists is set to an empty array? Will the TinyMCE selectors got hidden? It would be good if they did.

The selectors are not hidden automatically, they need to be removed from the `toolbar` conf var. What if rather than using empty arrays, adding `default_font` or `default_font_size` to `dont_override` also hid the TinyMCE selectors?

alecpl:
> The selectors are not hidden automatically, they need to be removed from the `toolbar` conf var. What if rather than using empty arrays, adding `default_font` or `default_font_size` to `dont_override` also hid the TinyMCE selectors?

That's not how I understand `dont_override`. So, no.

johndoh:
My thinking was that `dont_override` prevents the user from changing the default. And if having an empty font list (or a list with 1 font in) also hides the selection box then that's the same result. But I can imagine may be a admin would want to prevent the user changing the default but still allow them to change font when composing an individual message which I guess is your point.

program/actions/settings/index.php
```diff
@@ -998,8 +1000,8 @@ public static function user_prefs($current = null)
                     $blocks['main']['options']['default_font'] = [
```
alecpl, line 1000: We should skip the whole element if we skip both selects.

johndoh: I have reworked the whole section to make a cleaner flow and remove the double dont_override checks

PR #9189, opened 2023-10-27
get rid of jquery diff files

progsmile:
In order to cleanup repo, unused diffs were removed.


alecpl:
Why would you do that?

progsmile:
Updated.

alecpl:
We're using them after we update jquery-ui, because they contain our changes to the vanilla code.

progsmile:
Ok, supposed they are unused. Another solution might be creating one more css file with overriden rules for jquery original styles.

PR #9190, opened 2023-10-27
Remove imap readonly folder check

progsmile:
At this point `$this->conn` can't reach `setError` function as it is `protected`, so making it public

https://github.com/progsmile/roundcubemail/blob/master/program/lib/Roundcube/rcube_imap.php#L2869


alecpl:
That's the only place where we call setError(). I think we should consider removing this whole `if` instead.

progsmile:
Fixed.

PR #9208, opened 2023-11-08
patch-for-displaying-header-title

lenasterg:
This fix allows the .header-title to be visible also in large screens. This is helpful, specially in the settings screens.


alecpl:
This does not make sense. It breaks the toolbar.

PR #9213, opened 2023-11-11
Add possibility to change ATTR_EMULATE_PREPARES via config file

ShaiMagal:
In past there were this commit which add this _ATTR_EMULATE_PREPARES_:
https://github.com/roundcube/roundcubemail/commit/9faea49f827a8b3e765513a023dafb5ece0c9fc0

But in some cases there is necessary to use bool "true", not only bool "false".
Every time after update RC I must manually change this line (OK, well, not manually, I am using "sed" command, but...)...
This is really annoying. So, using config file for "override" this behaviour is logical step for improving "administrator user experience".

I got this error with _ATTR_EMULATE_PREPARES_ on value "false":
`[1615] Prepared statement needs to be re-prepared`
This config make problem:
`$config['password_query'] = 'UPDATE XXX SET `password` = %P WHERE `user` = %u';`
After change ATTR_EMULATE_PREPARES on value "true", problem dissapears.
This change I am doing from year 2021, without any problems or BC. So, now is ideal to solve this via config file :)

**For this reasons, there should be possibility to change this behavior via config file. There is no BC with this change.**

### DIFF
--- a/program/lib/Roundcube/db/mysql.php
+++ b/program/lib/Roundcube/db/mysql.php
@@ -135,7 +135,9 @@ class rcube_db_mysql extends rcube_db
         $result[PDO::ATTR_AUTOCOMMIT] = true;
 
         // Disable emulating of prepared statements
-        $result[PDO::ATTR_EMULATE_PREPARES] = false;
+        // configured value has higher prio
+        $conf_value = rcmail::get_instance()->config->get('attr_emulate_prepares');
+        $result[PDO::ATTR_EMULATE_PREPARES] = (($conf_value) ? $conf_value : false);
 
         return $result;
     }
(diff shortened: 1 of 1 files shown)


alecpl:
I'd suggest configuring this via DSN, like we do with 'verify_server_cert' a few lines above.

But it's strange this is needed as Roundcube is not using prepared queries.

Did you try `SET GLOBAL table_definition_cache = 1024` instead?

ShaiMagal:
@alecpl Hi,
 
 - why this config isn't OK? It's already using this way on other places.
 So, I should edit this PR and change "config" to $dsn?

 - it's not possible to change table_definition_cache, it is variable from webhosting, not possible to change. What impact it can have for this problem?

ShaiMagal:
@alecpl Hi, I updated code on your suggestion to using "$dsn". Is it now "correct" way?

alecpl:
I changed the option from `attr_emulate_prepares` to `emulate_prepares`.

ShaiMagal:
@alecpl ok :) thx!

ShaiMagal:
@alecpl Today we updated to RC 1.7.0 and tested it - works like a charm!

Issue #8201, opened 2021-09-15
OAuth2/OIDC discovery support


azmeuk:
Hi. Currently several URIs are mandatory in the oauth configuration:

```php
// Mandatory: URI for OAuth user authentication (redirect)
$config['oauth_auth_uri'] = null;

// Mandatory: Endpoint for OAuth authentication requests (server-to-server)
$config['oauth_token_uri'] = null;

// Optional: Endpoint to query user identity if not provided in auth response
$config['oauth_identity_uri'] = null;
```

However there is an [OAuth2 RFC](https://www.rfc-editor.org/rfc/rfc8414.html) and a [OIDC spec](https://openid.net/specs/openid-connect-discovery-1_0.html) that define standard URLs for json files describing the server metadata, and among other things, those URLs.

I suggest supporting those specs in roundcube, guess those URLs when possible, and make the configuration entries optional.

What do you think?

alecpl:
It might not be good for performance reasons to do the discovery on every logon, so we'd need to cache this somewhere. Maybe, a better approach would be to provide a discovery script that then returns the urls, or maybe puts them into the config file.

alecpl:
Implemented.

Issue #9109, opened 2023-08-28
Support for OpenID Connect RP-Initiated Logout


azmeuk:
I suggest implementing the [OIDC Connect RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) specification.

> This specification complements the [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#OpenID.Core) [OpenID.Core] specification by enabling the Relying Party to request that an End-User be logged out by the OpenID Provider.

This is the "mirror" spec of #9110. This would allow users to be disconnected from other services when they log out from roundcube. This is very convenient for service providers offering several services (say nextcloud and roundcube for instance).

Technically this would consist in:
- Sending a request with a payload to the Identity Provider (IdP) when users disconnect from roundcube

Related to #8057

alecpl:
Done.

Issue #9110, opened 2023-08-28
Support for OpenID Connect Back-Channel Logout


azmeuk:
I suggest to implement helpers for [OIDC Back-Channel Logout](https://openid.net/specs/openid-connect-backchannel-1_0.html):

> This specification defines a logout mechanism that uses direct back-channel communication between the OP and RPs being logged out; this differs from front-channel logout mechanisms, which communicate logout requests from the OP to RPs via the User Agent.

This is the "mirror" spec of #9109. This would allow users to be disconnected from roundcube when they are disconnected from other services. This is very convenient for service providers offering several services (say nextcloud and roundcube for instance).

Technically that would consist in:
- providing a logout endpoint that the Identity Provider (IdP) will be able to reach
- decode and validate a JWT sent by the IdP at the logout endpoint
- logout the user

EdouardVanbelle:
Hello I have started a POC where I can receive a backchannel logout request from the identity provider
I can found the ODIC sub and found the associated Roundcube's user 
(for that purpose I had to associate user to the OIDC sub)

The problem I have is that I cannot easily find all associated sessions for a given user. 
What I am thinking to do is something like: storing the logout event timestamp for a given sub. 
Then on any next refresh, if the associated user's token is older than the event I could raise a "session expired" without needing to query any server

Before going ahead it brings changes in database and I will appreciate feedbacks because I don't know the future evolution of Roundcube and I want to ensure a move forward aligned with the maintainers' vision

While waiting and seeing differently the problem: currently each time Roundcube does a refresh, it queries the IMAP server which may raise immediately a failure from the authentification request is OIDC session is burnt.

EdouardVanbelle:
I have a simple solution:

each time we receive a back channel logout request I will store in cache the `sub` and the logout request `timestamp`
when a customer trigger a refresh I will check if the current oauth session for the same sub was generated before the logout request, if so I just have to kill_session 

Will test it later

Doing this no need to patch any database

alecpl:
Implemented via #9217.

Issue #8214, opened 2021-10-01
OAUTH: Do not delete `id_token` from `$_SESSION['oauth_token']`


azmeuk:
Hi. I plugged roundcube with [rcmcarddav](https://github.com/mstilkerich/rcmcarddav) so it can fetch the contacts of my nextcloud instance, using [nextcloud-oidc-login](https://github.com/pulsejet/nextcloud-oidc-login/).

Nextcloud needs a JWT to authenticate the users, and check the `aud` claim in the token. rcmcarddav cannot give the `id_token` to nextcloud as roundcube deletes it and only let the `access_token`. Some providers like [canaille](https://gitlab.com/yaal/canaille/) fills `id_token` with a JWT, but `access_token` with a regular non JWT token (as [the spec illustrates](https://openid.net/specs/openid-connect-core-1_0.html#id_token-tokenExample)).

https://github.com/roundcube/roundcubemail/blob/1d2ad936a729ff7fec9d595b35f8e92ce7c5e44d/program/include/rcmail_oauth.php#L250

I could not figure what purpose it serves to delete `id_token`, but removing this line ([plus another tweak in rcmcarddav](https://github.com/mstilkerich/rcmcarddav/issues/354)) does fix the situation, as it allows rcmcarddav to forward the ID token to nextcloud.

I suggest not removing `id_token`. I volunteer to provide a patch if it is OK with you.

alecpl:
As mentioned in https://github.com/mstilkerich/rcmcarddav/issues/354 using id_token in this scenario is probably not the way it should be done. Therefore, I'm for a wontfix on this.

@thomascube for the last word. Maybe we could pass the `id_token` to the `oauth_login` hook, but not keep it in session.

thomascube:
I think we should not always store `id_token` in session but we might pass the full JWT `$data` to the `'oauth_login'` plugin hook. Then plugins like rcmcarddav can pick it from there and store it themselves if necessary.

PR #9217, opened 2023-11-16
oauth improvements

EdouardVanbelle:
* Add `oauth_config_uri` to parse the .well-known/openid-configuration and discover `oauth_*_uri`
  * Add `oauth_logout_uri` support of https://github.com/roundcube/roundcubemail/issues/9109  (security improvement) (credit: @Westie)
  * Add support of OAUTHBEARER 
  * Fix: missing config `oauth_provider_name` in rcmail_oauth's constructor
  * Refactor: move display to the rcmail_oauth class and use `loginform_content` hook


EdouardVanbelle:
Dear team I a proposing you this patch to improve the OIDC use, I am currently using it for my personal use on the 1.6.5 version

I would like to thanks also @Westie for his work (I included it in this PR)

Let me know if you prefer non squashed PR

Kind regards
Edouard

EdouardVanbelle:
Hello I am going to check these two requests:

- https://github.com/roundcube/roundcubemail/issues/9109
- https://github.com/roundcube/roundcubemail/issues/9110

I guess only 1 is satisfied here

program/include/rcmail_oauth.php
```diff
@@ -65,21 +82,65 @@ static function get_instance($options = [])
...
-            'timeout'     => $this->rcmail->config->get('oauth_timeout', 10),
+            'timeout'         => $this->rcmail->config->get('oauth_timeout', 10),
             'verify_peer'     => $this->rcmail->config->get('oauth_verify_peer', true),
             'auth_parameters' => $this->rcmail->config->get('oauth_auth_parameters', []),
             'login_redirect'  => $this->rcmail->config->get('oauth_login_redirect', false),
         ];
+
+        // fetch data directly from config_uri if specified (the .well-known/openid-configuration) and fill in URIs
+        $config_uri = $this->options['config_uri'];
+        if ( !empty($config_uri)) {
```
alecpl, line 107: that extra space all over the place needs to go out.

program/include/rcmail_oauth.php
```diff
@@ -28,6 +28,11 @@
  */
 class rcmail_oauth
 {
+    // prepare the OAUTHBEARER which is now the official protocol (rfc 7628)
+    // but currently implement mostly the formal XOAUTH2
+    /** @var string */
+    protected $prefered_auth = 'XOAUTH2';
```
alecpl, line 34: `auth_type` would be better.

alecpl (changes requested):

config/defaults.inc.php
```diff
@@ -349,15 +349,21 @@
...
 
-// Mandatory: Endpoint for OAuth authentication requests (server-to-server)
+// Mandatory or Optional if $oauth_config_uri is specified: Endpoint for OAuth authentication requests (server-to-server)
 $config['oauth_token_uri'] = null;
 
-// Optional: Endpoint to query user identity if not provided in auth response
+// Optional or Optional if $oauth_config_uri is specified: Endpoint to query user identity if not provided in auth response
 $config['oauth_identity_uri'] = null;
 
+// Optional or Optional if $oauth_config_uri is specified: Endpoint for OIDC Logout propagation
```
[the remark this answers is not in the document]
EdouardVanbelle: result of a 2 o'clock commit ;)

program/include/rcmail_oauth.php
EdouardVanbelle: you mean you prefer I move the highlighted block in a protected discover() method and call it in constructor ?

azmeuk:
Thank you very much for this patch! This would also fix #8201.

config/defaults.inc.php
```diff
@@ -349,15 +349,22 @@
...
 
-// Mandatory: Endpoint for OAuth authentication requests (server-to-server)
+// Mandatory or Optional if $oauth_config_uri is specified: Endpoint for OAuth authentication requests (server-to-server)
 $config['oauth_token_uri'] = null;
 
 // Optional: Endpoint to query user identity if not provided in auth response
 $config['oauth_identity_uri'] = null;
 
+// Optional: Endpoint for OIDC Logout propagation
+$config['oauth_logout_uri'] = null;
```
azmeuk, line 366: Why not call this var `oauth_end_session_endpoint` to fit to the OIDC spec vocabulary?

azmeuk: :thinking: Actually I don't know, `oauth_logout_uri` is similar to the other vars in rouncube :man_shrugging:

EdouardVanbelle: I kept same vocabulary as the current one, and @alecpl is right we need to setup a cache to avoid unnecessary call to OIDC

azmeuk:
I have tested this PR. Discovery works well!

As `id_token` is needed to pass as an `id_token_hint` parameter in the IdP end_session_endpoint, the `refresh_access_token` method is automatically called at logout since the `id_token` is absent from the session. I find this very strange to actually need to refresh a token just to logout, and invalid the token a few seconds after.

It was discussed a few years ago in #8214, at the time the maintainers preferred not to keep the id_token in the session. @thomascube would you reconsider this? I don't know if that was for security reasons, but nowadays with [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html) and people using JWT access tokens containing more or less the same payload than the `id_token`, I don't know if this is still relevant.

program/include/rcmail_oauth.php
```diff
@@ -533,9 +636,45 @@ public function managesieve_connect($options)
...
+        // if no ID token, refresh token
+        if (empty($_SESSION['oauth_token']['id_token'])) {
+            $this->refresh_access_token($_SESSION['oauth_token']);
+        }
+
+        // generate redirect URL for post-logout
+        $params = [
+            'post_logout_redirect_uri' => $this->rcmail->url([], true, true),
+            'id_token_hint' => $_SESSION['oauth_token']['id_token'],
+        ];
```
azmeuk, line 675: Out of curiosity, why not adding `client_id` and a `state` parameter here?

EdouardVanbelle: I did not challenged this part, need to check RFCs :)

EdouardVanbelle:
As we have the discovery, I took opportunity to fetch jwks (certificates) to verify JWT signatures, it is important for backchannel call to trust foreign calls

A future PR will come, for manual setup it will be config option `oauth_jwks_uri`

azmeuk:
I think I met one weak point: if the `refresh_access_token` returns `false` (because of wrong credentials or IdP misconfiguration for instance, or expired refresh token), then `$_SESSION['oauth_token']['id_token']` fails, and the roundcube user stays logged.

If `refresh_access_token` fails for whatever reason, maybe fill `$this->logout_redirect_url` anyways (after all the spec indicates the `id_token_hint` is *recommended* but not *mandatory*). 
Perhaps the `logout_hint` parameter would be useful in that situation? However I am not sure which value it should have. The obvious one is the email because it is one thing we are sure that roundcube has, but this may not be the value expected by the IdP. I don't know.

EdouardVanbelle:
@azmeuk I will check your remarks, currently the `refresh_token` is saved which is more sensitive than the `id_token`
I consider in a such case that we could store `id_token` too via the `encrypt()` method

azmeuk:
One last thing. On my setup, it seems that `id_token` is always null, even after calling `refresh_access_token`. After having a look at the `refresh_access_token` method, I cannot see anything that would save the `id_token` in `$_SESSION`, unlike `request_access_token`.

Looking closer at [the OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse), it seems the refresh token request should not return an id_token anyways:

> Upon successful validation of the Refresh Token, the response body is the Token Response of [Section 3.1.3.3Successful Token Response](https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse) except that it might not contain an id_token.

Thus, to be able to use `id_token_hint`, it seems that keeping the `id_token` somewhere is actually mandatory.

EdouardVanbelle:
here is some improvements:

-  add cache to Discovery and JWKS
- add debug, activated vi òauth_debug` = true
- improve refresh token
- fetch sub (will be used later)
- fix leak of access_token
- use firebase/php-jwt JWT livrary and check signature if `oauth_jwks_uri` present
- provide `id_token_hint` and `client_id` on OpenID Connect RP-Initiated Logout

@azmeuk if you test you can activate the debug option

program/include/rcmail_oauth.php
EdouardVanbelle: you now have the recommended options ;)

azmeuk:
Hi @EdouardVanbelle, I will try to test this soon.
The PR has become quite large. I wonder if splitting it up would not make the whole process easier for maintainers? :shrug:

EdouardVanbelle:
Hello @azmeuk ,  I pushed the version with the backchannel logout, you can add in your IDP a callback to `/index.php/oauth/backchannel` to test the functionality 

regarding the pull request need to check with @alecpl
I don't really know how to plug a such page which requires no session, and the way I did it may be an ugly hack

due to sensitivity of this patch it could be time to add tests before and after this patch, for that I need to learn how to code in php :) 

Right now I considered that it is overkill to redevelop a full JWT parser, so I used `firebase/php-jwt` which breaks current compatibility with **php 7.3**

EdouardVanbelle:
Hello @alecpl let me know if you need to split the pull request or rebase/squash it to make it more easier to review
I will appreciate any suggestion on how to catch a call for the backchannel url rather than the hack I did

index.php
```diff
@@ -95,8 +95,16 @@
 
 $session_error = null;
 
+// XXX HACK XXX handle backchannel oauth requests
+if ($RCMAIL->oauth->is_enabled() && (
+    $_SERVER['REQUEST_URI'] === '/index.php/oauth/backchannel' ||
+    $_SERVER['REQUEST_URI'] === '/oauth/backchannel' )) {
```
alecpl, line 101: I'm not very familiar with our OAuth related code.

Here I don't fully understand what is the hack? We already use `PATH_INFO` instead of `REQUEST_URI` in include/iniset.php, and it sets the task/action accordingly. Task will be overwritten in set_task(), but action should stick, so all we need is a code below this existing block:
```
// handle oauth login requests
else if ($RCMAIL->task == 'login' && $RCMAIL->action == 'oauth' && $RCMAIL->oauth->is_enabled()) {
    $oauth_handler = new rcmail_action_login_oauth();
    $oauth_handler->run();
}
```
similar to this, but with `$RCMAIL->action == 'backchannel'`. 

If the `$_SERVER['REQUEST_URI'] === '/oauth/backchannel'` part is required, I'd set action=backchannel  in the startup hook.

index.php
```diff
@@ -95,8 +95,16 @@
...
+    $oauth_handler = new rcmail_action_oauth_backchannel();
```
alecpl, line 102: Please, make it `rcmail_action_login_oauth_backchannel`. I see it's not exactly 'login', but it's related and I'd prefer to have them both in program/actions/login folder.

program/actions/oauth/backchannel.php
```diff
@@ -0,0 +1,102 @@
...
+                    ], true, false
+                );
+
+                $answer['error_description'] = "Error decoding JWT";
+            }
+
+        }
+        else {
+             rcube::raise_error([
+                    'message' => "oidc backchannel called without any parameter", // FIXME show IP from
```
alecpl, line 89: `rcube_utils::remote_ip()`

program/include/rcmail_oauth.php
```diff
@@ -28,18 +30,58 @@
  */
 class rcmail_oauth
 {
+    //FIXME: php seems supporting enum
```
alecpl, line 33: We're not ready to jump to PHP 8.1 yet, I think. We might decide later what will be the minimum for the next Roundcube version, but it would rather be 8.0, I suppose.

program/include/rcmail_oauth.php
```diff
@@ -66,20 +134,148 @@ static function get_instance($options = [])
...
+                if ($this->cache) {
+                    $this->cache->set("discovery", $data);
+                }
+            }
+
+            // map discovery to our options
+            foreach (self::$config_mapper as $config_key => $options_key) {
+                if(empty($data[$config_key])) {
+                    rcube::raise_error([
+                            'message' => 'key '.$config_key.' not found in answer of '.$config_uri,
```
alecpl, line 216: For such I usually prefer double quotes and variables in form of `{$var}` inside.

program/include/rcmail_oauth.php
```diff
@@ -66,20 +134,148 @@ static function get_instance($options = [])
...
+        $this->jwks = json_decode($response->getBody(), true);
+
+        //sanity check
+        if (!isset($this->jwks['keys'])) {
+            $this->log("incorrect jwks response from %s", $this->options['jwks_uri']);
+        }
+        else if ($this->cache) {
+            // this is a hack because we cannot specify the TTL in the shared_cache
+            // and cache must not be too high as the Identity Provider can rotate it's keys
+            $this->jwks['expires']=time() + self::JWKS_CACHE_TTL;
```
alecpl, line 276: Code style. space `=` space.

program/include/rcmail_oauth.php
```diff
@@ -66,20 +134,148 @@ static function get_instance($options = [])
...
+
+        if (empty($config_uri)) {
+            return;
+        }
+
+        if ($this->cache === null) {
+            rcube::raise_error([ 'message' => 'Warning you activated `oauth_config_uri` it is highly recommanded to setup `oauth_cache`' ], true, false);
+        }
+
+
```
alecpl, line 192: Code style. Please remove all unneeded empty lines, like here and in the beginning of the method body. Also, `if` space `(`, and I see an unneeded space after `(` in a few places.

alecpl (changes requested):

index.php
EdouardVanbelle: I used `oauth` task because I saw in `program/include/rcmail.php`    
```
public static $main_tasks = ['mail','settings','addressbook','login','logout','utils','oauth','dummy'];
```

I am correcting following your recommendations

alecpl (changes requested):

program/include/rcmail_oauth.php
```diff
@@ -58,47 +99,213 @@ static function get_instance($options = [])
...
+                $data = json_decode($response->getBody(), true);
+
+                //sanity check
+                if (!isset($data['issuer'])) {
+                    throw new RuntimeException("incorrect response from %s", $config_uri);
+                }
+
+                //cache answer
+                if ($this->cache) {
+                    $this->cache->set("discovery", $data);
```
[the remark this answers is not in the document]
EdouardVanbelle: ok, I will apply same behavior with `jwks` cache

program/include/rcmail_oauth.php
```diff
@@ -58,47 +99,213 @@ static function get_instance($options = [])
...
+        if ($this->jwks !== null) {
+            // already defined
+            return;
+        }
+
+        if ($this->cache === null) {
+            rcube::raise_error([ 'message' => 'Warning you activated `oauth_config_uri` it is highly recommanded to setup `oauth_cache`' ], true, false);
+        }
+
+        $this->jwks = $this->cache ? $this->cache->get("jwks") : null;
```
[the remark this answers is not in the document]
EdouardVanbelle: will use same mechanism for discovery:
```php
  $key_cache = "jwks".md5($jwks_uri);
```

program/include/rcmail_oauth.php
```diff
@@ -433,6 +739,125 @@ public function refresh_access_token(array $token)
...
+            $this->log_debug('all tokens generated before %s timestmp are comromised', $data['not-before-policy']);
+            // TODO
+        }
+
+        // please note that id_token / identity may have changed, could be interesting to grab it and refresh values, right now it is not used
+
+        //creation time. Information also present in JWT, but it is faster here
+        $data['created_at'] = time();
+
+        $min_refresh_interval = $this->rcmail->config->get('min_refresh_interval');
```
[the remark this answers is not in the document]
EdouardVanbelle: here I do the assumption that user don't do any action (passive on his browser)
so I am using the refresh frequency to be sure the access_token will be refreshed before it's expiration, it is not the best way to do and saw that user can change his refresh frequency via $a_user_prefs['refresh_interval']

need to do deeper tests to be sure that if I don't do this trick, that the refresh_token will occurs **before** the next IMAP sync

index.php
```diff
@@ -198,6 +198,13 @@
     $oauth_handler->run();
 }
 
+// handle oauth login requests
+else if ($RCMAIL->task == 'login' && $RCMAIL->action == 'backchannel' && $RCMAIL->oauth->is_enabled()) {
```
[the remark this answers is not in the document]
EdouardVanbelle: I have this working on `rcmail_oauth`

```php 
    public function startup(array $args)
    {
        if (!$this->is_enabled()) {
            return $args;
        }

        if ($args['task'] == 'login' && $args['action'] == 'oauth') {
            // handle oauth login requests
            $oauth_handler = new rcmail_action_login_oauth();
            $oauth_handler->run();
        }
        else if ($args['task'] == 'login' && $args['action'] == 'backchannel') {
            // handle oauth login requests
            $oauth_handler = new rcmail_action_login_oauth_backchannel();
            $oauth_handler->run();
        }
        else if ($args['task'] == 'logout') {
            //handle only logout task
            $this->handle_logout();
        }

        return $args;
    }
```

which remove oauth code in index.php

I could also migrate `rcmail_action_login_oauth` and `rcmail_action_login_oauth_backchannel` into `rcmail_oauth`
Doing this rcmail_oauth becomes an autonomous class but this code is becoming deeply long...

That said, it seems that that rcmail_oauth can become a plugin 
Let me know your preferences

program/include/rcmail_oauth.php
```diff
@@ -456,14 +887,73 @@ protected function mask_auth_data(&$data)
...
+    }
+
+
+    /**
+     * Returns the auth_type to use
+     *
+     * @return string
+     *   the auth type: XOAUTH or OAUTHBEARER
+     */
+    public function get_auth_type()
```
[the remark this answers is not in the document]
EdouardVanbelle: I have this kind of POC:

```diff
@@ -965,18 +952,24 @@ class rcube_imap_generic
         $auth_methods = [];
         $result       = null;

         // check for supported auth methods
-        if (!$auth_method || $auth_method == 'CHECK') {
+        if (!$auth_method || $auth_method == 'CHECK' || $auth_method == 'CHECK_OAUTH') {
             if ($auth_caps = $this->getCapability('AUTH')) {
                 $auth_methods = $auth_caps;
             }

+            if ($auth_method == 'CHECK_OAUTH') {
+                $all_methods = ['XOAUTH2','OAUTHBEARER'];
+            }
+            else {
                 // Use best (for security) supported authentication method
                 $all_methods = ['DIGEST-MD5', 'CRAM-MD5', 'CRAM_MD5', 'PLAIN', 'LOGIN'];

                 if (!empty($this->prefs['gssapi_cn'])) {
                     array_unshift($all_methods, 'GSSAPI');
                 }
+            }

             foreach ($all_methods as $auth_method) {
                 if (in_array($auth_method, $auth_methods)) {
```

doing this I set auth_type='CHECK_OAUTH' and let library choosing the authentification method according server's capabilities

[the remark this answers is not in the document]
alecpl: Yes. Makes sense. I think we'll need this also for Sieve and SMTP. One thing though, I don't like `CHECK_OAUTH`. I'd make it just `OAUTH`?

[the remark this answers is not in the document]
EdouardVanbelle: We are now in the most complexe situation: where we have to find the best naming :)
ok for OAUTH

I submitted yesterday some merge requests for that purpose on Sieve & SMTP libraries

Do you agree to submit this change via another pull request ? 
Current MR is already including too many features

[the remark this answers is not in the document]
alecpl: Sure.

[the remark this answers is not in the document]
EdouardVanbelle: Hello @alecpl, I took a look on `program/actions/login/oauth.php` code is 90% same as the block 
```php
if ($RCMAIL->task == 'login' && $RCMAIL->action == 'login') {
...
``` 
in `index.php` which means if this part is changed, it may break oauth login

I will check to move logic into `authenticate` hook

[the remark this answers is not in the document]
EdouardVanbelle: Here is a work in progress: (based on current merge request) https://github.com/EdouardVanbelle/roundcubemail/commit/dcfd29d318ef1feefc5d8c800a08d9684a242259
I am going to mark this thread as resolved
Once merged we will focus on this other commit

Neustradamus:
To follow

EdouardVanbelle:
Hello @alecpl you have the next coming changes on another branch (once this one is approved)
https://github.com/EdouardVanbelle/roundcubemail/compare/master...EdouardVanbelle:roundcubemail:feat/hooks-and-oauth?expand=1

- choose the most appropriate auth mechanism for Sieve, IMAP and SMTP (will require latest libraries versions)
- simplify action/login/oauth in order to use hooks

alecpl:
@EdouardVanbelle So, the last part here that is not clear is that one with 'min_refresh_interval'.

Since you already know this code better than me, could you take a look at #9244 for feedback, please.

program/include/rcmail_oauth.php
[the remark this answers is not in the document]
EdouardVanbelle: Good catch, I mixed up `min_refresh_interval` and `refresh_interval`, now corrected

alecpl:
Please, fix the code style issues. We deployed code style checking in CI.

EdouardVanbelle:
@alecpl here it is !

alecpl:
@EdouardVanbelle Thank you for this work and I'm waiting for more!

EdouardVanbelle:
Thank you @alecpl  !
I prepare the next merge request
This also close this request: https://github.com/roundcube/roundcubemail/issues/9110

Neustradamus:
@EdouardVanbelle: Good job!

oculos:
@EdouardVanbelle Could you share a bit how to configure logout for keycloak? I've updated Roundcube, and attempted to use the new configs, but I can't seem to log off.

EdouardVanbelle:
Hello @oculos , here is my configuration:

tip concerning your `oauth_config_uri`: if you open this url in your browser you should have a json containing all your OIDC config

### Roundcube config:
(on my side I have enabled the debugging)

```php
$config['oauth_debug'] = true;
$config['oauth_provider'] = 'generic';
$config['oauth_provider_name'] = 'Vanbelle';
$config['oauth_client_id'] = 'YOUR-CLIENT-ID';
$config['oauth_client_secret'] = '#######';
$config['oauth_config_uri'] = 'https://YOUR-KEYCLOAK-DOMAIN/realms/YOUR-REALM/.well-known/openid-configuration';
$config['oauth_verify_peer'] = true;
$config['oauth_scope'] = "email openid";
$config['oauth_auth_parameters'] = [];
$config['oauth_identity_fields'] = ['email'];
// setup to true once config ok to remove user/pass prompt and redirect to oauth login
$config['oauth_login_redirect'] = false;
$config['oauth_cache'] = 'db';
$config['oauth_cache_ttl'] = '1d';
```


### Keycloak config:
```
General Settings
    Client ID: YOUR-CLIENT-ID
    Root URL : https://YOUR-ROUNDCUBE-DOMAIN/
    Home URL : https://YOUR-ROUNDCUBE-DOMAIN/
    Valid redirect URIs : https://YOUR-ROUNDCUBE-DOMAIN/index.php/login/oauth
    Valid post logout redirect URIs : https://YOUR-ROUNDCUBE-DOMAIN/?_task=logout
    Web origins : YOUR-ROUNDCUBE-DOMAIN
    Admin URL : https://YOUR-ROUNDCUBE-DOMAIN/

Capability config
    Client authentication : On
    Authorization : Off
    Authentication flow
        [X] Standard flow
        [ ] Direct access grants  <------ BEWARE: use it only if you need your MTA has a proxy in user/password gateway, usefull for a transition, I do not recommand it.
        [ ] Implicit flow
        [X] Service accounts roles
        [ ] OAuth 2.0 Device Authorization Grant
        [ ] OIDC CIBA Grant

Logout settings
    Front channel logout : Off
    Backchannel logout URL : https://YOUR-ROUNDCUBE-DOMAIN/index.php/oauth/backchannel
    Backchannel logout session required : On
    Backchannel logout revoke offline sessions : Off
```

### Debugging backchannel


To try backchannel logout: If you are on your Keycloak account in user:
https://YOUR-KEYCLOAK-DOMAIN/realms/YOUR-REALM/account/#/ > logout

on your roundcube's log directory, if you enabled debug on config (set `oauth_debug` to `true`), you should have:
```
$ tail -f oauth.log
[10-Jun-2024 17:30:34 +0000]: <0mb25nud> DEBUG: [ip=<IP FROM KEYCLOAK SERVER> sub=- ses=-] backchannel: logout event received, schedule a revocation for token's sub: <ID Of THE SUB>
[10-Jun-2024 17:30:39 +0000]: <6m4s24lp> DEBUG: [ip=<IP OF THE CUSTOMER (next refresh)> (X-Forwarded-For: 88.175.55.47) sub=<ID Of THE SUB> ses=6da52dcb-855b-4776-a813-6552f1275f2e] abort, token for sub <ID OF THE SUB> has been revoked
```

### Debugging postlogout

once logged in Roundcube, just logout, it should propagate to Keycloak the logout request:

```
$ tail -f oauth.log
[10-Jun-2024 17:38:41 +0000]: <oetqnpfm> DEBUG: [ip=<IP OF THE CUSTOMER>  sub=<ID Of THE SUB> ses=fa2a21b9-5f69-43c5-829a-a2bf1c22c1b7] creating logout call: https://YOUR-KEYCLOAK-DOMAIN/realms/YOUR-REALM/protocol/openid-connect/logout?post_logout_redirect_uri=https%3A%2F%2FYOUR-ROUNDCUBE-DOMAIN%2F%3F_task%3Dlogout&client_id=YOUR-CLIENT-ID&id_token_hint=YOUR-TOKEN
```

and your session on keycloak should be kicked

oculos:
Thanks a lot @EdouardVanbelle 
What is weird here is that I don't think roundcube is indeed picking up the configurations.
As I said, having only the `oauth_config_uri` doesn't work, even if the link is correct (I clicked on it from the terminal and it does show all the links - it's just that Roundcube doesn't pick them.

Not even the debug is working: 

```
-rw-r--r--  1 www-data www-data 677K Jun 10 12:03 errors.log
-rw-r--r--  1 www-data www-data    0 Jun  8 19:00 oauth.log
```

I created the oauth.log file because it wasn't created at all. and it's still empty despite debug is on. So something is not working as it should.

EdouardVanbelle:
`oauth_config_uri` works only if you enabled the cache, which means ensuring:
```php
$config['oauth_cache'] = 'db';
$config['oauth_cache_ttl'] = '1d';
```
could you confirm you enabled it ?

if config uri does not work, you should have the reason  your `errors.log` (errors are kept in default errors.log)
something like: 
```Error fetching {$config_uri} : ...```

**Comment by oculos** [comment]
This is my configuration: 

```
$config['oauth_provider'] = 'generic';
$config['oauth_provider_name'] = 'mydomain SSO';
$config['oauth_client_id'] = 'roundcube';
$config['oauth_client_secret'] = 'somesecret';
$config['oauth_auth_uri'] = 'https://auth.mydomain.eu/realms/mydomain/protocol/openid-connect/auth';
$config['oauth_token_uri'] = 'https://auth.mydomain.eu/realms/mydomain/protocol/openid-connect/token';
$config['oauth_identity_uri'] = 'https://auth.mydomain.eu/realms/mydomain/protocol/openid-connect/userinfo';
$config['oauth_scope'] = 'Roundcube_email openid';
$config['oauth_config_uri'] = 'https://auth.mydomain.eu/realms/mydomain/.well-known/openid-configuration';
$config['oauth_logout_uri'] = 'https://auth.mydomain.eu/realms/mydomain/protocol/openid-connect/logout';
$config['oauth_cache'] = 'db';
$config['oauth_debug'] = true;
$config['oauth_verify_peer'] = true;
$config['oauth_cache_ttl'] = '8h';
$config['oauth_identity_fields'] = ['postfixMailAddress'];
$config['oauth_login_redirect'] = true;
```

If I comment out `$config['oauth_token_uri'] = 'https://auth.mydomain.eu/realms/mydomain/protocol/openid-connect/token';`, I don't get pass the Roundcube login window with the login button.

Absolutely no new errors on `errors.log` when doing so.

**Comment by EdouardVanbelle** [comment]
@oculos even if you comment out `oauth_token_uri`and provided the `oauth_config_uri` the  *discover()* method should grab it and explain any error in errors.log

Could you let met know which commit you are working on ?  so I could check to reproduce
I just need the hash:
```
git rev-parse --short HEAD
```

(feature is right now only available on master's branch which is still in development )

oculos:
Ohhh now I see. I mistakenly thought this was already added to the package that is publicly available (1.6.7). You see, I don’t install from the repo, I just download the package from Roundcube’s website. 
so sorry for wasting your time @EdouardVanbelle

EdouardVanbelle:
No worries @oculos, at least I provided a sample of use ;)

PR #9220, opened 2023-11-17
Adding log entry for a succesful password change

kuchtamartin:
Added log for password change success


alecpl:
An option already exists:
```
// Enables logging of password changes into logs/password
$config['password_log'] = false;
```

PR #9227, opened 2023-11-26
Ignore mbstring.func_overload in PHP 8

cgzones:
The feature `mbstring.func_overload` is deprecated since PHP version 7.2 and removed since 8.0.0
(see https://www.php.net/manual/en/mbstring.overload.php).

Skip trying to set it for PHP version 8, e.g. to avoid failures with the sloppy comparison hardening of Snuffleupagus
(https://snuffleupagus.readthedocs.io/config.html#prevent-sloppy-comparison):

    ERROR: Wrong 'mbstring.func_overload' option value and it wasn't possible to set it to required value (0). Check your PHP configuration (including php_admin_flag).

### DIFF
--- a/program/lib/Roundcube/bootstrap.php
+++ b/program/lib/Roundcube/bootstrap.php
@@ -28,11 +28,16 @@
 $config = [
     'display_errors'  => false,
     'log_errors'      => true,
-    // Some users are not using Installer, so we'll check some
-    // critical PHP settings here. Only these, which doesn't provide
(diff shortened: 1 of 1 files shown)


alecpl:
I doubt the whole Roundcube is compatible with this Snuffleupagus setting.

Indeed on PHP8 both `init_get('mbstring.func_overload')` and `filter_var(false, FILTER_VALIDATE_INT)` return false. So, we might merge this anyway. Thanks.

Issue #9204, opened 2023-11-05
Enigma + Mailvelope: option "Use Mailvelope main keyring" not retained => no key found


sphakka:
Trying to use Enigma from Roundcube v1.6.4 (OIDC auth) with Mailvelope v5.1.1 on Firefox v116.0.3. My domain is correctly authorized in Mailvelope's configuration with API access; Roundcube correctly enables the "Mailvelope options" section in "Settings > Preferences > Encryption". 

However, if I enable option "Use Mailvelope main keyring" and then click "Save", that option resets itself to disabled. As a consequence, no key is available in menu "Settings > PGP keys".

## Steps to reproduce
1. Activate and configure Enigma
2. Install Mailvelope addon in a clean Firefox profile (no other addon installed)
4. Authorize your domain, say,  "*.example.com" with + API access in Mailvelope's configuration page
5. Generate/import a key pair in Mailvelope for, say, 'user@example.com'
7. Log in to Roundcube as 'user@example.com'
9. Go to "Settings > Preferences > Encryption",
10. Toggle (enable) option "Use Mailvelope main keyring" and click "Save"

## Expected result
Option "Use Mailvelope main keyring" is **kept enabled**. User's key is available in menu "Settings > PGP keys". 

## Actual result
Option "Use Mailvelope main keyring" is **reset to disabled**. No key is available in menu "Settings > PGP keys".

No relevant information found in logs...

alecpl:
"Settings > PGP keys" is controlled by Enigma and has nothing to do with Mailvelope. Also, I don't think it makes sense to use both Enigma and Mailvelope.

That being said, looks like there might be an issue with keeping that setting enabled.

sphakka:
Right, thanks. The reason for trying Mailvelope came from the fact that I'm unable to generate keys with Enigma -- I'll open a different issue for that.

BTW, even with Mailvelope disabled, the option "Use Mailvelope main keyring" is still actionable, whereas it should be greyed-out.

sphakka:
Gosh, it looks like `openpgp.js` is not installed with Debian 12's `roundcube-plugins` package!

PR #9240, opened 2023-12-05
Fix saving other encryption settings besides enigma's

pabzm:
Previously, if the enigma plugin was enabled, it wasn't possible to save the option whether Roundcube should use Mailvelope's main keyring (`mailvelope_main_keyring`), because this code overwrote the whole array for the "encryption"-section on saving. That is now fixed.


pabzm:
This issue was mentioned in #9204 already, but apparently didn't get fixed yet.

PR #9241, opened 2023-12-07
Use native composer autoloader

mvorisek:
In #9407 and #9412 composer autoloading was introduced to coexist with the legaly autoloader.

This PR replaces the legacy autoloader so composer is newly needed.

### DIFF
--- a/plugins/database_attachments/database_attachments.php
+++ b/plugins/database_attachments/database_attachments.php
@@ -29,8 +29,6 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-require_once INSTALL_PATH . 'plugins/filesystem_attachments/filesystem_attachments.php';
-
 class database_attachments extends filesystem_attachments
 {
     // Cache object
--- a/plugins/debug_logger/debug_logger.php
+++ b/plugins/debug_logger/debug_logger.php
@@ -65,8 +65,6 @@ class debug_logger extends rcube_plugin
 
     public function init()
     {
-        require_once __DIR__ . '/runlog/runlog.php';
-
         $this->runlog = new runlog();
 
         if (!rcmail::get_instance()->config->get('log_dir')) {
--- a/plugins/enigma/enigma.php
+++ b/plugins/enigma/enigma.php
@@ -103,11 +103,6 @@ class enigma extends rcube_plugin
 
         $this->env_loaded = true;
 
-        // Add include path for Enigma classes and drivers
-        $include_path = $this->home . '/lib' . \PATH_SEPARATOR;
-        $include_path .= ini_get('include_path');
-        set_include_path($include_path);
-
         // load the Enigma plugin configuration
         $this->load_config();
 
--- a/plugins/example_addressbook/example_addressbook.php
+++ b/plugins/example_addressbook/example_addressbook.php
@@ -1,7 +1,5 @@
 <?php
 
-require_once __DIR__ . '/example_addressbook_backend.php';
-
 /**
  * Sample plugin to add a new address book
  * with just a static list of contacts
--- a/plugins/identicon/identicon.php
+++ b/plugins/identicon/identicon.php
@@ -55,8 +55,6 @@ class identicon extends rcube_plugin
             }
 
             if ($email) {
-                require_once __DIR__ . '/identicon_engine.php';
-
                 if (!empty($args['attrib']['bg-color'])) {
                     $bgcolor = $args['attrib']['bg-color'];
                 } else {
--- a/plugins/managesieve/managesieve.php
+++ b/plugins/managesieve/managesieve.php
@@ -281,11 +281,6 @@ class managesieve extends rcube_plugin
     public function get_engine($type = null)
     {
         if (!$this->engine) {
-            // Add include path for internal classes
-            $include_path = $this->home . '/lib' . \PATH_SEPARATOR;
-            $include_path .= ini_get('include_path');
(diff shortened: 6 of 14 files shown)


alecpl:
What's the reason for replacing composer.json-dist with composer.json. On upgrade we can't just sync that file. It will remove all plugins/skins/packages added by the user.

mvorisek:
Yes, upgrade script is the last place with `composer.json-dist`. The reason I renamed the file is only `composer.json` is read by packagist/composer installer and this repo is impossible to be used as a `roundcube/roundcubemail` dep in other projects otherwise. I would be grateful if you can help me with migrating the upgrade script, probably we want to copy the `composer.json` as `composer.json-dist` or rename the current `composer.json` before the upgrade is started.

mvorisek:
I have reworked/simplified this PR to update the autoloading only, eg. to replace the custom autoloaders in favor of native composer autoloaders. The advantages are summarized in the PR description.

alecpl:
Ok, less contentious, I guess, but still
1. We need to move/merge "autoload" and "autoload-dev" sections into composer.json in here https://github.com/roundcube/roundcubemail/blob/98ac6e02f5b057ed5b355dbdbf2fe637aa9b299d/bin/update.sh#L177
2. As this might be a BC break for some plugins, I'd have to make some more tests, to find out what the impact is.
3. I'm not sure, but this might be a problem for Roundcube packages in distros like Debian, but I don't really know.
4. Will that work at all when using "Dependent" (non-complete) Roundcube package to upgrade your installation? I.e. will bin/ scripts work without composer.json file and vendor folder? I guess it will not work. Does that mean we have to stop providing "Dependent" Roundcube packages?

alecpl:
FYI, I know two plugins that depend on `rcube_autoload`. It's our `managesieve` plugin, and external `sauserprefs` plugin.

mvorisek:
PR is done, all autoloaders are migrated to native composer.

External plugins needs to be updated if they do not utilize native composer already.

johndoh:
With this PR external plugins which utilize the current Roundcube autoloader stop working unless they are changed to use the composer autoloader. This is a breaking change.

Not everyone uses composer to install plugins, not everyone can, not everyone has command line access to their hosting. This PR prevents those people from being able to use external plugins which use the autoloader.

Please consider retaining the Roundcube autoloader or do something else to maintain the functionality of external plugins used in none composer environments.

If Composer is to become a requirement then please update the documentation e.g. https://github.com/roundcube/roundcubemail/wiki/Plugin-API#installing-and-activating-plugins to reflect this.

alecpl:
Internal plugins should be in "require" not "require-dev", I suppose. Anyway, I'm not sure I like such a big composer.json file. I'd prefer to keep it simpler. More complicated it becomes, more error-prone updates become.

I understand it allows to remove some code in tests, but other than that it's not a big win. That part regarding plugins, I mean.

As for the general autoloader, I like that part. That being said, I would accept this only for Roundcube 2.0. We might decide to go with 2.0 instead of 1.7 if we merge other breaking changes, specifically #9294.

mvorisek:
> Internal plugins should be in "require" not "require-dev"

The plugins are not hosted on Packagist, ie. `composer require roundcube/roundcube` would fail otherwise. `repositories` composer config is for composer root package only.

Other reason to keep them in "require-dev" is to allow plugins customization using composer in child packages.

Becauose of this I belive it is wanted to the bundled plugins listed in "require-dev". For distribution/downloads, the `roundcube/roundcube` is a composer root package, so nothing changes.

> Not everyone uses composer to install plugins, not everyone can, not everyone has command line access to their hosting or is confident enough to use it. This PR prevents those people from being able to use internal or external plugins which use the autoloader.

I support this. I will rework this PR to keep the old autoloading working. I am fairly confident it can coexist until v2.0 at least.

mvorisek:
I have reviewed the removed/replaced original autoloader methods.

Let's separate the autoloading to two domains.

ROUNDCUBE INTERNAL AUTOLOADING - all autoloading code loading files withing this repo. With this PR, such loading is replaced with composer native autoloading mechanism.

There any many advantages, it is faster, more secure, static analysers can rely on "descriptive autoloading", and it also allows to use this package as a dependency easily.

The "disadvantage"/BC break is composer is needed. I do not think this is a problem as long as the "composer update" is run before the files are packged and distributed. This is already done for "complete distribution bundle" - https://roundcube.net/download/.

> Not everyone uses composer to install plugins, not everyone can, not everyone has command line access to their hosting or is confident enough to use it. This PR prevents those people from being able to use internal or external plugins which use the autoloader.

With this said, composer autoloader is fine to be used for v1.7/master. Composer files are distributed and no composer is needed for the install (on hosting).

CUSTOM PLUGIN AUTOLOADING - autoloading of files outside this repo. The preferred autoloading is using composer for the same reason as for internal files. However, I agree with https://github.com/roundcube/roundcubemail/pull/9241#issuecomment-2041016781 above, currently the Roundcube does not require composer to be installed/rerun when a plugin is added and it should stay so.

This has been fixed in https://github.com/roundcube/roundcubemail/commit/fd0d66989127b68a414e80289aa08c49617246c4. AFAK the plugin main file (file named like the plugin directory) is kept loaded by Roundcube as before and if the plugin relies on autoloading using custom `set_include_path` like `enigma` did - https://github.com/roundcube/roundcubemail/blob/7713b7c1bd/plugins/enigma/enigma.php#L109 - it is kept supported as well.

With this said, I belive this PR can be landed. Feel free to test this PR and let me know if there are any concerns.

PR #9246, opened 2023-12-09
Fix and assert basic CS using CI

mvorisek:
Assert basic coding style according to the latest standards and reduce merge conflicts in a long term.

This PR lands changes for non-functional (and mostly whitespace) changes only. I plan to fix more CS in upcoming PRs.


alecpl:
This is cool. I'd keep old switch/case indentation if possible. Not that I have strong opinion about the syntax, but I'd like to keep the diff less invasive. Please.

mvorisek:
> This is cool. I'd keep old switch/case indentation if possible. Not that I have strong opinion about the syntax, but I'd like to keep the diff less invasive. Please.

The `case` indentation is fixed by https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/whitespace/statement_indentation.rst PHP CS Fixer rule. It is not configureable for `case`, thus I cannot serve.

In a long them, this should pay off, as the PSR-12 CS [1] recommends/allows only this format. If you do a lot of merged into bugfix branches, I advise to add and run the PHP CS Fixer on these branches too, so the whitespace CS is the same and the diffs minimal.

[1] https://www.php-fig.org/psr/psr-12/#52-switch-case

mvorisek:
PR is done and rebased on the latest master.

mvorisek:
@alecpl can this PR be merged?

alecpl:
Probably not before this weekend.

mvorisek:
Let's land #9258 first, I wil then rebase this PR.

PR #9248, opened 2023-12-10
Fix explode_quoted_string for multibyte delimiter

mvorisek:
Also much more performant, previously the string was processed byte by byte.

What is the controbution policy, what branch should it target?


alecpl:
Use master unless the fix does not apply to it, then you can use release-1.x branch.

PR #9250, opened 2023-12-10
Fix format_date with "x" format for Windows

mvorisek:
the 1st param of `IntlDateFormatter::__construct` is Intl locale, not system locale

on Windows and PHP 8.3 and I getting `Found unconstructed IntlDateFormatter` exceptions - https://github.com/php/php-src/issues/12912

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -212,6 +212,7 @@ class rcmail extends rcube
 
         // set localization
         setlocale(LC_ALL, $lang . '.utf8', $lang . '.UTF-8', 'en_US.utf8', 'en_US.UTF-8');
+        ini_set('intl.default_locale', $lang);
 
         // Workaround for http://bugs.php.net/bug.php?id=18556
         // Also strtoupper/strtolower and other methods are locale-aware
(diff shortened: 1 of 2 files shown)


alecpl:
The intention of the code was to return "date representation based on locale" (i.e. `%x` from `strftime()`). Using English US date format is just wrong.

That being said, we dropped strftime-compatible format support in 1.6. Also, the 'x' formatter does something completely different in `date()`.

So, we have tho options:
1. remove this whole if branch
2. fix the issue properly (and use another letter?)

mvorisek:
What about `new IntlDateFormatter(null)` to use `intl.default_locale`?

docs: https://www.php.net/manual/en/intldateformatter.create.php#refsect1-intldateformatter.create-parameters

alecpl:
Maybe, if you set `intl.default_locale` in `rcmail::set_user()`.

alecpl:
In other words, this has to be a user locale, not system locale.

mvorisek:
updated

alecpl:
The tests pass, but we should set `intl.default_locale` there too. I.e. in Rcmail_Rcmail::test_format_date().

PR #9249, opened 2023-12-10
Fix Sieve scripts comment parse with CRLF

mvorisek:

PR #9251, opened 2023-12-10
Run unit tests in CI on Windows too

mvorisek:
merge after #9249, #9250 and #9253

### DIFF
--- a/.github/workflows/browser_tests.yml
+++ b/.github/workflows/browser_tests.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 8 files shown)


.github/workflows/browser_tests.yml
```diff
@@ -14,8 +14,10 @@ jobs:
 
     strategy:
       fail-fast: true
+      matrix:
+        php: ["8.1", "8.3"]
```
alecpl, line 18: I don't think we need to run browser tests with more than one php version.

mvorisek:
PR is done. All non-tests changes has been landed.

PR #9252, opened 2023-12-10
Fix invalid phpdocs

mvorisek:


mvorisek:
PR is done. I reviewed all the changed manually and tested them by PHP CS Fixer and PHPStan.

PR #9253, opened 2023-12-10
Fix rcube_imap::sort_folder_comparator() warning on Windows

mvorisek:


mvorisek:
needed for/tested by https://github.com/roundcube/roundcubemail/pull/9251

PR #9268, opened 2023-12-17
Assert expected data types in tests

mvorisek:
Use `TestCase::assertSame()` instead of `TestCase::assertEquals()` which does not assert expected data type.

Fix [`php_unit_strict`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/php_unit/php_unit_strict.rst) rule.

### DIFF
--- a/tests/Framework/Bootstrap.php
+++ b/tests/Framework/Bootstrap.php
@@ -74,7 +74,7 @@ class Framework_Bootstrap extends PHPUnit\Framework\TestCase
 
         foreach ($data as $value => $expected) {
             $result = slashify($value);
-            $this->assertEquals($expected, $result, "Invalid slashify() result for $value");
+            $this->assertSame($expected, $result, "Invalid slashify() result for $value");
         }
 
     }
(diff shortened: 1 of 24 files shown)


tests/Framework/Bootstrap.php
```diff
@@ -56,7 +56,7 @@ function test_parse_bytes()
 
         foreach ($data as $value => $expected) {
             $result = parse_bytes($value);
-            $this->assertEquals($expected, $result, "Invalid parse_bytes() result for $value");
+            $this->{'assertEquals'}($expected, $result, "Invalid parse_bytes() result for $value");
```
alecpl, line 59: ???? here and in some other places.

alecpl (changes requested):

tests/Framework/Bootstrap.php
mvorisek: The (fixed) `php_unit_strict` rule is risky, so I reviewed them manually to prevent bad changed. Here, I would expect bytes to be parsed to integer, so I kept original `assertEquals`, so with a fix in future, the test will still pass. The fix/improvement is here: #9268. In that PR, the test is hardened to `assertSame`.

PR #9270, opened 2023-12-17
Modernize more CS III

mvorisek:
This PR focuses on rules with a few changes only.


program/actions/login/oauth_backchannel.php
```diff
@@ -1,24 +1,24 @@
...
- | PURPOSE:                                                              |
- |    Implementation of backchannel logout from IDP                      |
- |                                                                       |
- | @see https://openid.net/specs/openid-connect-backchannel-1_0.html     |
- |                                                                       |
- | URL to declare: <roundcube instance>/index.php/login/backchannel      |
- +-----------------------------------------------------------------------+
- | Author: Thomas Bruederli <roundcube@gmail.com>                        |
- +-----------------------------------------------------------------------+
+ * +-----------------------------------------------------------------------+
```
alecpl, line 4: This change looks like a false positive. This is a file comment.

plugins/jqueryui/themes/elastic/jquery-ui.css.diff
```diff
@@ -8,7 +8,7 @@ index 2f0601a..819a492 100644
  	opacity: 0;
 -	-ms-filter: "alpha(opacity=0)"; /* support: IE8 */
```
alecpl, line 9: I suppose the changes in diff files aren't intended, right?

alecpl (changes requested):

program/actions/login/oauth_backchannel.php
mvorisek: I reviewed it, but did not compare it to other files, good catch!

plugins/jqueryui/themes/elastic/jquery-ui.css.diff
mvorisek: Damm, there are EOL changes GitKraken did/does not shown me!

mvorisek: It seems the CI passes even /w the EOL changed. So I prefer this text files to be normalized.

mvorisek: There changes should be fine. The dark red is syntax/diff highlighting of .diff file, not highlighting of the actual changes made.

The only changes are removed trailing spaces, which should be completely fine, as ` ` (space) at the line beginning is useless in diff if not followed by any data.

alecpl: Ah, indeed. That's misleading. Anyway, would it make sense to exclude .diff and .js files for now?

mvorisek: I did not include them into the CS fixer, I only did some experiment to include all text files localy and in https://github.com/roundcube/roundcubemail/pull/9270/commits/e2cfaa12a753e9efdc27c20d9314ca3b40c05be9 I commited the files where trailing spaces are normally better to be removed.

PR #9271, opened 2023-12-17
Return `integer|false` from parse_bytes() function

mvorisek:
Bytes are discrete, however `2.5MiB` decimal value should still be accepted, at least for BC.


alecpl:
I just added `convertDeprecationsToExceptions="true"` in master. I suppose the null case will throw a deprecation warning.

PR #9275, opened 2023-12-18
Refer native constants unambiguously

mvorisek:


.php-cs-fixer.dist.php
```diff
@@ -31,7 +31,25 @@
             'equal' => false,
             'identical' => false,
         ],
-        'native_constant_invocation' => true,
+        'native_constant_invocation' => [
+            'include' => [
```
mvorisek, line 35: FYI this list is not for CI, this list is to be able to run CS Fixer with stable results for more environments like local/Windows.

mvorisek:
@alecpl I want to base others PRs on this one, can this PR be merged?

PR #9279, opened 2023-12-19
Rename composer.json.dist to composer.json

mvorisek:
needed for #9241, for using this project/libs as a composer/Packagist dependency in general


bin/installto.sh
```diff
@@ -78,8 +78,9 @@ if (strtolower($input) == 'y') {
         }
     }
 
-    foreach (['index.php', 'config/defaults.inc.php', 'composer.json-dist', 'jsdeps.json', 'CHANGELOG.md', 'README.md', 'UPGRADING', 'LICENSE', 'INSTALL'] as $file) {
-        $command = "rsync -a --out-format=%n " . INSTALL_PATH . "$file $target_dir/$file";
+    foreach (['index.php', 'config/defaults.inc.php', 'composer.json.dist', 'jsdeps.json', 'CHANGELOG.md', 'README.md', 'UPGRADING', 'LICENSE', 'INSTALL'] as $file) {
```
mvorisek, line 81: This ensures the `composer.json` is downloaded still to `composer.json.dist` as expected by https://github.com/roundcube/roundcubemail/blob/1.6.5/bin/update.sh#L177 update script.

I couldn't find more places to fix, but please review this PR carefully, there might be more install usecases/docs I am not aware.

composer.json
```diff
@@ -3,7 +3,7 @@
     "description": "The Roundcube Webmail suite",
     "license": "GPL-3.0-or-later",
     "require": {
-        "php": ">=7.3.0",
+        "php": ">=7.3 <8.4",
```
alecpl, line 6: I'm not sure about this change.

alecpl:
This will make working directly on source code more complicated. I normally have my composer.json file with additional packages/plugins added. I won't be able to do that. I need to investigate what options there are to fix that. One I found is use of `COMPOSER` env pointing to a different file.

mvorisek:
`COMPOSER` env. var [1] is for root package [2] only. Dependencies require to use `composer.json` strictly. I belive there is no workaround and this PR is needed.

[1] https://getcomposer.org/doc/03-cli.md#composer
[2] https://getcomposer.org/doc/04-schema.md#root-package

composer.json
mvorisek: landed in #9298

bin/installto.sh
```diff
@@ -79,7 +79,8 @@ if (strtolower($input) == 'y') {
     }
 
     foreach (['index.php', 'config/defaults.inc.php', 'composer.json-dist', 'jsdeps.json', 'CHANGELOG.md', 'README.md', 'UPGRADING', 'LICENSE', 'INSTALL'] as $file) {
-        $command = 'rsync -a --out-format=%n ' . INSTALL_PATH . "$file $target_dir/$file";
+        $source_file = $file === 'composer.json-dist' ? 'composer.json' : $file;
```
mvorisek, line 82: This ensures the `composer.json` is downloaded still to `composer.json-dist` as expected by https://github.com/roundcube/roundcubemail/blob/1.6.5/bin/update.sh#L177 update script.

I couldn't find more places to fix, but please review this PR carefully, there might be more install/update usecases/docs I might be not fully aware.

alecpl:
I created https://github.com/composer/composer/issues/11784 but I don't expect it to be implemented soon or at all.

The package I mentioned there is not of a great quality. However, I tested the `COMPOSER` variable and it looks like it might actually work. But it's not an easy decision.

mvorisek:
@alecpl can I ask you to review this PR? Based on https://github.com/roundcube/roundcubemail/pull/9279#issuecomment-1868383175 and https://github.com/composer/composer/issues/11784 the `composer.json` name is the only one name possible to make this project working with composer natively.

mvorisek:
I have rebased this PR. @alecpl, can you please check? According to the sources above, there are no other options, the only supported filename by composer is `composer.json` and the place that needs to be adjusted is the distrubution/upgrade.

mvorisek:
@alecpl, would you mind to look into this PR?

PR #9282, opened 2023-12-22
Update SECURITY.md

miaulalala:
Add Nextcloud HackerOne link instead of mail addresses

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,8 +6,7 @@ Check our website's [download page](https://roundcube.net/download/) to see whic
 
 ## Reporting a Vulnerability
 
-If you found a security issue or vulnerability of the software, please report with direct and encrypted email to *thomas[at]roundcube.net*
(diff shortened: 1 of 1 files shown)


xkill:
Report on nextcloud? Why?

alecpl:
It supposed to be added as an extra channel, not replacing the old one. I still want to be a target for security reports. We can remove Thomas's email, and/or add another one i needed.

xkill:
But Nextcloud is not Roundcube. I don't know why the security report related to Roundcube have to be reported on Hackersone Nexcloud. Moreover, nextcloud has it's own mail app (it also supports to integrate Roundcube, but is a 3rd party app, not the official one).

alecpl:
https://roundcube.net/news/2023/11/30/nextcloud-the-new-home-for-roundcube

xkill:
> https://roundcube.net/news/2023/11/30/nextcloud-the-new-home-for-roundcube

Thanks for the clarification!!!

miaulalala:
@alecpl can we merge? Since we discussed this :)

ChristophWurst (approved):

PR #9283, opened 2023-12-23
Fix single quotes CS

mvorisek:


bin/update.sh
```diff
@@ -62,7 +62,7 @@ if ($RCI->configured) {
         // list obsolete config options (just a notice)
         if (!empty($messages['obsolete'])) {
             echo "NOTICE: Obsolete config options:\n";
-            echo "(You still have some obsolete or inexistent properties set."
+            echo '(You still have some obsolete or inexistent properties set.'
```
alecpl, line 65: I don't know. My preference was to use double quotes for long/longish text, like error messages or sql queries. And single quotes for short labels. I don't like mixing single quotes with double quotes like here.

mvorisek: There is no fixer for it to maintain such style. Symfony and other projects use single quotes when possible (no `\n`, variables... inside a string) so IMHO we should use this CS and assert it using PHP CS Fixer.

mvorisek: @alecpl let's merge this PR, it asserts the CS wel and is commontly used by most of PHP projects (Symfony, Lavarel, ...).

PR #9274, opened 2023-12-17
OAuth improvements (+ add user_create hook)

EdouardVanbelle:
Dear team, please find OAuth improvements

 * Core: Improv.: can specify an array on *_auth_type, lib will elect the best auth type according server's capabilities
 * OAuth: Refact.: migrate login flow into hooks (more evolutive code and less dependency to core code)
 * OAuth: Fix: logger prefix (include prefix during login phase)
 * OAuth: Feat: add user_create hook (map OIDC claims when creating a new user)
 * OAuth: Lean: prepare OAUTHBEARER or XOAUTH2 in Sieve, SMTP, IMAP
     
This pull request also prepare the support of OAUTHBEARER and can already be merged, I will add OAUTHBEARER once:
 - pear Net_SMTP https://github.com/pear/Net_SMTP/pull/79 is accepted + included in a new release
 - pear Net_Sieve will have a new release (thank you @alecpl for the merge of https://github.com/pear/Net_Sieve/pull/10 )
 
 Many thanks for your time

PR #9286, opened 2023-12-26
oauth: feat: use OIDC claims on user creation

EdouardVanbelle:
Hello please find the user_create hook that uses the OIDC's identify when creating a new user

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -407,6 +407,17 @@ $config['oauth_cache'] = 'db';
 // Optional: cache ttl
 $config['oauth_cache_ttl'] = '8h';
 
+// Optional: map OIDC claims to Roundcube keys during the account creation
+// format: roundcube_key => array of claims (the first claim found and defined will be used)
+// more informations on claims: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+// roundcube_key can be: user_name, user_email, language
+// default value:
+$config['oauth_user_create_map'] = [
+    'user_name' => ['name'],
+    'user_email' => ['email'],
+    'language' => ['locale'],
+];
+
 ///// Example config for Gmail
 
 // Register your service at https://console.developers.google.com/
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -70,6 +70,9 @@ class rcmail_oauth
     /** @var ?array parameters used during the login phase */
     protected $login_phase;
 
+    /** @var array list of allowed keys in user_create_map (note that user and host are protected) */
+    protected static $user_create_allowed_keys = ['user_name', 'user_email', 'language'];
+
     /** @var array map of .well-known entries to config (discovery URI) */
     static protected $config_mapper = [
         'issuer'                 => 'issuer',
@@ -156,6 +159,13 @@ class rcmail_oauth
             'client_secret'   => $this->rcmail->config->get('oauth_client_secret'),
             'identity_uri'    => $this->rcmail->config->get('oauth_identity_uri'),
             'identity_fields' => $this->rcmail->config->get('oauth_identity_fields', ['email']),
+            'user_create_map' => $this->rcmail->config->get('oauth_user_create_map', [
+                //rc key => OIDC Claim @see: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims )
+                'user_name' => ['name'],
+                'user_email' => ['email'],
+                'language' => ['locale'],
+            ]),
+
             'scope'           => $this->rcmail->config->get('oauth_scope', ''),
             'timeout'         => $this->rcmail->config->get('oauth_timeout', 10),
             'verify_peer'     => $this->rcmail->config->get('oauth_verify_peer', true),
@@ -170,6 +180,7 @@ class rcmail_oauth
             $options['http_options'] = [];
         }
 
+        // sanity check on PKCE value
         if ($this->options['pkce'] && !array_key_exists($this->options['pkce'], self::$pkce_mapper)) {
             // will stops on error
             rcube::raise_error([
@@ -177,6 +188,18 @@ class rcmail_oauth
                 'file'    => __FILE__,
                 'line'    => __LINE__,
             ], true, true);
+        }
+
+        // sanity check that configuration user_create_map contains only allowed keys
+        foreach ($this->options['user_create_map'] as $key => $ignored) {
+            if (!in_array($key, self::$user_create_allowed_keys)) {
+                // will stops on error
+                rcube::raise_error([
+                    'message' => "use of key `{$key}` in `oauth_user_create_map` is not allowed",
+                    'file'    => __FILE__,
+                    'line'    => __LINE__,
+                ], true, true);
+            }
         }
 
         // prepare a http client with the correct options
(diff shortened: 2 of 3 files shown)


program/include/rcmail_oauth.php
```diff
@@ -1066,6 +1074,46 @@ public function login_after($options)
...
+        $identity = $this->login_phase['token']['identity'];
+
+        foreach ($this->options['user_create_map'] as $rc_key => $oidc_claims) {
+            if (!is_array($oidc_claims)) {
+                // normalize to array
+                $oidc_claims = [$oidc_claims];
+            }
+            foreach ($oidc_claims as $oidc_claim) {
+                // use the first defined claim
+                if (isset($identity[$oidc_claim]) && $identity[$oidc_claim] !== null) {
```
alecpl, line 1104: Check for null is redundant. Maybe we should just use `!empty()` or check whether it's a non-empty string?

program/include/rcmail_oauth.php
```diff
@@ -1066,6 +1074,46 @@ public function login_after($options)
...
+
+        if (!isset($this->login_phase['token']['identity'])) {
+            $this->log_debug("identity not found, was the scope 'openid' defined ?");
+            return $data;
+        }
+
+        $identity = $this->login_phase['token']['identity'];
+
+        foreach ($this->options['user_create_map'] as $rc_key => $oidc_claims) {
+            if (!is_array($oidc_claims)) {
```
alecpl, line 1098: I usually do just a casting to array in such cases.

config/defaults.inc.php
```diff
@@ -403,6 +403,17 @@
...
+// Optional: map OIDC claims to Roundcube keys during the account creation
+// format: roundcube_key => array of claims (the first claim found and defined will be used)
+// more informations on claims: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+// default value:
+// [
+//   'user_name' => ['name'],
+//   'user_email' => ['email'],
+//   'language' => ['locale'],
+// ]
+$config['oauth_user_create_map'] = null;
```
alecpl, line 415: Set the default array here instead of null.

program/include/rcmail_oauth.php
```diff
@@ -1066,6 +1074,46 @@ public function login_after($options)
...
+     * @return array $data key/values to setup user's identity
+     */
+    public function user_create($data)
+    {
+        if (!$this->login_phase) {
+            return $data;
+        }
+
+        if (!isset($this->login_phase['token']['identity'])) {
+            $this->log_debug("identity not found, was the scope 'openid' defined ?");
```
alecpl, line 1091: no space before `?`.

program/include/rcmail_oauth.php
```diff
@@ -1066,6 +1074,46 @@ public function login_after($options)
...
+            foreach ($oidc_claims as $oidc_claim) {
+                // use the first defined claim
+                if (isset($identity[$oidc_claim]) && $identity[$oidc_claim] !== null) {
+                    $data[$rc_key] = $identity[$oidc_claim];
+                    break; //no need to continue
+                }
+            }
+        }
+
+        $this->log_debug("setup account creation of user:%s host:%s with user_name:%s user_email:%s lang:%s",
```
alecpl, line 1111: Maybe it would be better to just log the items we got from the token identity.

program/include/rcmail_oauth.php
```diff
@@ -1066,6 +1074,46 @@ public function login_after($options)
...
+
+        foreach ($this->options['user_create_map'] as $rc_key => $oidc_claims) {
+            if (!is_array($oidc_claims)) {
+                // normalize to array
+                $oidc_claims = [$oidc_claims];
+            }
+            foreach ($oidc_claims as $oidc_claim) {
+                // use the first defined claim
+                if (isset($identity[$oidc_claim]) && $identity[$oidc_claim] !== null) {
+                    $data[$rc_key] = $identity[$oidc_claim];
```
alecpl, line 1105: Maybe we should validate the values coming from outside?

alecpl (changes requested):

program/include/rcmail_oauth.php
EdouardVanbelle: good to know, thanks

program/include/rcmail_oauth.php
EdouardVanbelle: just a french reflex :)

program/include/rcmail_oauth.php
EdouardVanbelle: I opted for non empty string

program/include/rcmail_oauth.php
EdouardVanbelle: Good catch, here what I did:
1. ensure config setup only allowed keys (user & host are for example forbidden)
2. normalizing user_email & language, then I check validity

I also added tests

Neustradamus:
@EdouardVanbelle: Nice, good job!

Linked to:
- https://github.com/roundcube/roundcubemail/pull/9274
- https://github.com/roundcube/roundcubemail/pull/9286
- https://github.com/roundcube/roundcubemail/pull/9287
- https://github.com/roundcube/roundcubemail/pull/9288
- https://github.com/roundcube/roundcubemail/pull/9289
- https://github.com/pear/Net_SMTP/pull/79
- https://github.com/pear/Net_Sieve/pull/10
- https://github.com/pear/Net_Sieve/pull/11

alecpl:
@Neustradamus Please, stop adding this references. It's misleading, not everything is related.

program/include/rcmail_oauth.php
EdouardVanbelle: I just changed the language check, the list_languages() was too strict

```diff
-                            // check that language is supported
-                            $languages = $this->rcmail->list_languages();
-                            if (!array_key_exists($value, $languages)) {
+                            // check that language is at the correct format
+                            if (!preg_match('/^[a-z]{2}(\_[A-Z]{2,})?$/', $value)) {
                                 rcube::raise_error([
```

program/include/rcmail_oauth.php
```diff
@@ -1132,6 +1156,72 @@ public function login_after($options)
...
+                                    'line'    => __LINE__,
+                                ], true, false);
+                                continue 2; // continue on next foreach iteration
+                            }
+                            break;
+                        case 'language':
+                            // normalize language
+                            $value = strtr($value, '-', '_');
+                            // check that language is at the correct format
+                            if (!preg_match('/^[a-z]{2}(\_[A-Z]{2,})?$/', $value)) {
```
alecpl, line 1204: That's not going to work with languages like "ast" or "es_419"

EdouardVanbelle: What about this simplification?

```php
   // check that language is in the correct format (over simplification of RFC5646)
   if (!preg_match('/^[a-zA-Z]{2,8}(\_[a-zA-Z0-9]{1,8})*$/', $value)) {
  ...
```

Which already sanitize data and ensure you are not going to have temptative hacks like `$language="../../../password"` or similar 

What is the requirement on the user's property `language` 
- does it need to be the wished language from the customer? (allow language to be added lately) 
- or does it need to be a truly and currently supported language by roundcube?

If it as to be a supported language, the best option I see is to change `rcube->language_prop()` from `protected` to `public` and use it in this hook

What do you think ?

alecpl: In another place we're using `/^[a-zA-Z0-9_-]+$/`, but I'd add a length limit, lets say 8, and it will be fine.

EdouardVanbelle: done

please note, I saw that `rcube->language_prop()` is capturing browser's accepted language via
```php
            if (preg_match('/^([a-z]+)[_-]([a-z]+)$/i', $lang, $m))
```            

I added this: https://github.com/roundcube/roundcubemail/pull/9292

Issue #8757, opened 2022-10-22
Support PKCE with OAuth2


JE4GLE:
[PKCE](https://oauth.net/2/pkce/) is an important security measure of the OAuth 2.0 workflow.

Roundcube currently does not support it and I would suggest integrating it.

Thank you!

EdouardVanbelle:
Hello @JE4GLE, good idea !
As I am currently working on OAuth improvements, I will check your request

Neustradamus:
@EdouardVanbelle: Nice idea!

alecpl:
Implemented.

vaminakov:
Seems it does not working with some IAMs.
I'm using zitadel, and when I set roundcube to use PKCE, the error in logs:
`roundcube[]: {"error":"invalid_request","error_description":"code_challenge required"}`

PR #9287, opened 2023-12-26
oauth: feat: add support of PKCE

EdouardVanbelle:
Hello please find the PKCE implementation (RFC 7636)

- by default it is enabled, if server do not implement it, PKCE extra parameters will be ignored (as requested by RFC)
- I did not implemented plain as requested as we can directly support `S256` hash method
- I used a hash array to prepare future hahses (even if RFC specifies only `S256` right now) 

This covers this request: https://github.com/roundcube/roundcubemail/issues/8757

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -353,7 +353,7 @@ $config['oauth_client_id'] = null;
 $config['oauth_client_secret'] = null;
 
 // Optional: the OIDC discovery URI (the 'https://.../.well-known/openid-configuration')
-// if specified, the discovery will supersede `oauth_issuer`, `auth_auth_uri`, `oauth_token_uri`, `oauth_identity_uri`, `oauth_logout_uri`, `oauth_jwks_uri`)
+// if specified, the discovery will supersede `oauth_issuer`, `auth_auth_uri`, `oauth_token_uri`, `oauth_identity_uri`, `oauth_logout_uri`, `oauth_jwks_uri`
 // it is recommanded to activate a cache via `oauth_cache` and `oauth_cache_ttl`
 $config['oauth_config_uri'] = null;
 
@@ -365,6 +365,10 @@ $config['oauth_jwks_uri'] = null;
 
 // Mandatory: URI for OAuth user authentication (redirect)
 $config['oauth_auth_uri'] = null;
+
+// Optional: PKCE protection, by default it is enabled to S256 method, to disable it use `false`
+// please note that `plain` method is voluntarily not implemented
+$config['oauth_pkce'] = 'S256';
 
 // Mandatory or Optional if $oauth_config_uri is specified: Endpoint for OAuth authentication requests (server-to-server)
 $config['oauth_token_uri'] = null;
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -80,6 +80,12 @@ class rcmail_oauth
         'jwks_uri'               => 'jwks_uri',
     ];
 
+    /** @var array map PKCE code_challenge_method to hash method */
+    static protected $pkce_mapper = [
+        'S256' => 'sha256',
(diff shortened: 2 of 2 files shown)


config/defaults.inc.php
```diff
@@ -366,6 +366,10 @@
 // Mandatory: URI for OAuth user authentication (redirect)
 $config['oauth_auth_uri'] = null;
 
+// Optional: PKCE protection, by default it is enabled to S256 method, to disable it use `false`
+// please note that `plain` method is not implemented
+$config['oauth_pkce'] = null;
```
alecpl, line 371: Set the default here.

program/include/rcmail_oauth.php
```diff
@@ -163,6 +170,19 @@ public function __construct($options = [])
             $options['http_options'] = [];
         }
 
+        if ($this->options['pkce'] === true) {
```
alecpl, line 173: Handling `true` is not needed.

program/include/rcmail_oauth.php
```diff
@@ -474,22 +511,40 @@ public function login_redirect()
...
+            // do not store it in clear, do not want it to be readable
+            $_SESSION['oauth_code_verifier'] = $this->rcmail->encrypt($code_verifier);
+
+            $query += [
+                'code_challenge_method' => $code_challenge_method,
+                'code_challenge' => self::base64url_encode(hash($hash_method, $code_verifier, true)),
+            ];
+        }
+
+        $this->log_debug("requesting authorization code via a redirect to %s with scope='%s' and pkce method=%s",
```
alecpl, line 540: I'm just thinking maybe logging the full url instead would be better.

alecpl (changes requested):

program/include/rcmail_oauth.php
EdouardVanbelle: Yes, I am always very permissive for admin/users, I convert 'true' to the default value
Removing it

program/include/rcmail_oauth.php
EdouardVanbelle: I did it to avoid leak of sensitive data like the `nonce` in other pull request

Neustradamus:
@EdouardVanbelle: Nice, good job!

Linked to:
- https://github.com/roundcube/roundcubemail/pull/9274
- https://github.com/roundcube/roundcubemail/pull/9286
- https://github.com/roundcube/roundcubemail/pull/9287
- https://github.com/roundcube/roundcubemail/pull/9288
- https://github.com/roundcube/roundcubemail/pull/9289
- https://github.com/pear/Net_SMTP/pull/79
- https://github.com/pear/Net_Sieve/pull/10
- https://github.com/pear/Net_Sieve/pull/11

program/include/rcmail_oauth.php
```diff
@@ -367,6 +395,11 @@ public static function base64url_decode($encoded)
         return base64_decode(strtr($encoded, '-_', '+/'), true);
     }
 
+    public static function base64url_encode($payload)
```
alecpl, line 398: I think both these methods should be not public.

program/include/rcmail_oauth.php
```diff
@@ -474,22 +507,40 @@ public function login_redirect()
         }
 
         // create a secret string
-        $_SESSION['oauth_state'] = rcube_utils::random_bytes(12);
+        $_SESSION['oauth_state'] = rcube_utils::random_bytes(32);
```
alecpl, line 510: Does it have to be 32. Is it in some standard?

alecpl (changes requested):

program/include/rcmail_oauth.php
EdouardVanbelle: Your right I kept the same login as the initial version which was public, changing it to protected

program/include/rcmail_oauth.php
EdouardVanbelle: juste my paranoid reflex

EdouardVanbelle:
Thank you !

Issue #8851, opened 2023-01-09
securing log file locations


iseries:
Good day!

I am working as a security researcher. I found during [a review of web servers of German municipalities and authorities](https://renerehme-dev.translate.goog/blog/informationssicherheit-in-den-kommunen?_x_tr_sl=de&_x_tr_tl=en&_x_tr_hl=de&_x_tr_pto=wapp) that many Roundcube instances in production are not configured correctly. Also a recent investigation of German universities showed the same picture. I was able to view log files and temporary email attachments on over 30 instances.

The data that was drained during my investigation amounted to several GB. Among them were ID cards, passports, medical cards, various applications, birth certificates, and sick notes.

In other words, highly sensitive personal data.

Of course, the responsibility for a correct configuration lies with the operator of the application. After exchanging information with the [CCC](https://www.ccc.de/en/) and various data protection authorities, I would like to point out this issue and suggest that it would make sense to change the software in future Roundcube versions so that log files and temporary attachments are stored outside the documentRoot by default. This would not only contribute to the security of data, but also strengthen the Roundcube software in general in its confidentiality.

Best regards,
Rene Rehme

LauraTaylorUK:
So in other words, you have access to their roundcube `/logs` and `/temp` directories? If that is so, it implies one of the following:

- the .htaccess files have been removed
- AllowOverride is set to `none` (or not enough permissions)
- a required Apache module is not loaded
- they don't use Apache, thus there is no support for .htaccess files

I would guess, that it is the administrators responsibility to install roundcube correctly and if they don't use .htaccess files then they should take the appropriate action to block access to `/logs` and `/temp` directories.

Your suggestion to store these things out of the public_html directory is reasonable, but it could complicate roundcube's installation?

alecpl:
There's `public_html` folder in Roundcube for a reason. We do not force users to use it. I'm not sure we should (e.g. by providing `index.php` file under `public_html` only). But we recommend to use it in the INSTALL file (`SECURE YOUR INSTALLATION` section).

iseries:
> There's `public_html` folder in Roundcube for a reason. We do not force users to use it. I'm not sure we should (e.g. by providing `index.php` file under `public_html` only). But we recommend to use it in the INSTALL file (`SECURE YOUR INSTALLATION` section).

Sorry to bring this up again. The question is: Why should you not provide an index.php file under public_html only?

[...]

Issue #5738, opened 2017-04-27
Enhancement: static asset separation


corrideat:
Roundcube currently comes in a package that includes some PHP code and some static assets. Additional plugins and themes also follow this structure.

I believe it would be useful to provide some separation between these. This could allow for the following:

1. Storing Roundcube's executable files from outside the web root. This can help, for example, for easier file permission management and reduces the risk of exposing sensitive information (e.g., database configuration, passwords, etc.) in misconfigured or misbehaved servers.
2. With an additional configuration option (base resource path), much as it is currently done for the logo image, static assets could be served from a different domain or a CDN. 

Having a look, I don't believe this should be too difficult to implement in code, although it can be a greater issue ensuring that themes and plugins' files remain accessible (as, basically, they'd need to place files in two locations instead of a single directory, as it is now.)

In addition, a(n optional) [subresource integrity](https://w3c.github.io/webappsec-subresource-integrity/) tag can be added to the code generating HTML elements to prevent potential vulnerabilities that could arise from implementing this feature (such as a malicious CDN.)

alecpl:
We have already assets_path and assets_dir settings. So, you can separate assets, but you have to do it by your own. We have also public_html/ folder. The only issue is with plugins that as you noticed would need to split resources too, and I don't really like this idea.

BTW, Kolab already does this on a packaging level.

corrideat:
Yes, while `assets_path` kind of allows separation, the manual process is a bit tedious and error-prone. (I have written a simple script to catch most files, by extension, but it won't work if some plugin uses an extension I didn't think of.)

Hence, my idea was more along the lines of having two main directories, say `application` (code) and `resources` (static files). Then, the process for implemeting separation is server-side is very simple.

For plugins, I'm thinking a compromise would be to recommend this same separation _within the plugin directory_. Then, the files would still need to be manually copied, but at least won't require hunting down the right files.

alecpl:
#9294 has been merged. Now all resources are served by public_html/static.php. Imo, a separation is not needed.

ner00:
> Now all resources are served by public_html/static.php.

Not all resources, apparently.

For example, when I test the `newmail_notifier` plugin in the Settings, to test notification popup or test notification sound, I get a 403 error because the plugin is redirecting the user directly to the original path of the resource, bypassing `static.php`.

Here's an excerpt from `newmail_notifier.js`:
```js
// Sound notification
function newmail_notifier_sound() {
[...]

PR #9294, opened 2023-12-31
Add static files server

alecpl:
In an attempt to improve an overall security of Roundcube installation related to the fact that people aren't using our `public_html` folder, I prepared an initial implementation of a static files server.

The idea is that we do not allow direct access to static resources at all. I.e. you must use the `public_html` as your root 
folder, and that folder will contain only two files: `index.php` and `static.php` (and maybe `.htaccess`).

It also solves problems people have with the links (`skins` and `plugins`) in public_html folder.

I'm interested in your opinion. Please, note it's not a complete solution. We'd have to:
- make it working with `use_secure_urls` feature,
- make it working with `assets_path` feature,
- get rid of `assets_dir`, as it won't be needed anymore,
- move `.htaccess` and index.php into the public_html folder.

Let me know what do you think.

Reference: #8851, #5738.


Neustradamus:
Nice PR!

alecpl:
@johndoh @mvorisek @EdouardVanbelle @cmollekopf what do you think?

mvorisek:
a) `.htaccess` can solve any security issue as long as no dynamic ACL based on like the logger user is needed. `.htaccess` is supported more or less by Apache only.

b) On Windows, symlinks are very mildly supported and even on NTFS they are not usually extracted (by common archive utilities) from the release archvives.

c) It will add more load to the webserver as the static content requests will have to go thru PHP, but it seems like caching is implemented so this should not be an issue, actually it can improve performance on webservers with wrongly configured caching.

To sum this up:

Given a) this PR helpful for non-Apache based webservers like nginx (with non-ignoreable marker share).

Given a) and c) this PR should be not the default/only static files routing solution when `.htaccess` is supported.

Given b), this PR is needed (for restricting access to all files). All symlinks can/should be removed then.

People are lazy and how many of them configure `public_html` as output dir at all? If this PR should solve all file access issues, it would make sense to add a runtime check if `.htaccess` is supported (ie. if served thru Apache) or if accessed thru `public_html/*.php`. If satisfied, we are safe. If not, we should inform the user the installation can be insecure by default. User (administrator) should still be able to disable this check, as he might want to mimic the `.htaccess` by a custom webserver config.

alecpl:
The point is to remove current `./index.php` file and have `public_html/index.php` file so admins are forced to use `public_html` as the entry point (it won't be optional as it is now). And after that there might be only one `.htaccess` file in `public_html`.

mvorisek:
Inexperienced users should not be forced to change their webserver route to `public_html` and experienced users should be allowed to host Roundcube in a subfolder.

So there needs to a be `.htaccess` in the root.

`index.php` in the root removed (or kept with some simple info) can effectively act as the "runtime check", as either `.htaccess` in the root will be honored or the request was made thru `public_html`.

The problem of serving static files and securing other files is not related to this repo only. I would prefer some solution using separate repo. Maybe there is some package which solves it well already.

johndoh:
I second @mvorisek's thoughts about Windows. If there are any contributors who run a Windows bassed setups may be they could comment?

If `index.php` is removed from the root then people have to setup `public_html` as the root, right? May be adding some security checks into the installer, to make sure thing like the `logs` are not readable by the client, would also help people secure their setups?

> If file type is not on the list it will have to be served by custom code in your plugin.

I tried serving a non-supported file from a test plugin. Its a very rough test but if you have `$this->include_script($this->local_skin_path() . '/foo.bar');` in your plugin that result still has `static.php` in it but then fails because `static.php` does not support `.bar` files and plugins can't interact with it.

Small thing but `str_starts_with()` is PHP8 only.

alecpl:
We could make use of https://github.com/Stadly/FileWaiter. I didn't find other library that isn't part of some big framework.

I really see no problem in forcing users to use `public_html`.

As for performance, I think it will be good enough. Users that require something really fast can setup a separate static assets server, I suppose.

Regarding Windows and links. The only problem (link) would be the installer. I suppose we could provide install.php instead of a link.

melroy89:
I do want to use `public_html` already, but the resources like the skins will no longer load. Example: https://mail.melroy.org/

Changing the document root back to the actual root (without public_html) will load all the static files correctly. What do I wrong today? I understand why people won't use `public_html` today even without this PR...

melroy89:
> I found it.. [The wiki guide told me to use FTP](https://github.com/roundcube/roundcubemail/wiki/Installation#get-it-ready) (eg. FileZilla) to copy the data to the server, but it seems like it's not copying the symlinks from `public_html` !

Please be sure to update the documention wiki page with some additional information that copying those files will not copy symlinks. Or either describe a better installation method instead..?

program/lib/Roundcube/bootstrap.php
```diff
@@ -436,3 +436,9 @@ function rcube_autoload(string $classname): bool
 
     return false;
 }
+
+if (!function_exists('str_starts_with')) {
+    function str_starts_with(string $haystack, string $needle) {
+        return strlen($needle) === 0 || strpos($haystack, $needle) === 0;
+    }
+}
```
mvorisek: https://github.com/atk4/core/blob/5.2.0/composer.json#L35-L38 would be better solution as installed dependencies are required in all cases

alecpl: You're right. Anyway, we should consider bumping PHP version requirement to 8.1 for the next Roundcube version, especially if it will be a bump to 2.0.

pabzm: > we should consider bumping PHP version requirement to 8.1 for the next Roundcube version

I'm all for it!

alecpl:
This is ready for a final review. I think the only issue/regression left is with the `assets_path` setting, but I think it's rarely used. We'll have to consider limiting it's use (i.e. be specific about what can be put there).

public_html/static.php
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * @param string $path File location
+ */
+function serveStaticFile($path): void
+{
+    $lastModifiedTime = filemtime($path);
+
+    header('Last-Modified: ' . gmdate('D, d M Y H:i:s \G\M\T', $lastModifiedTime));
+    if (!empty($_SERVER['HTTP_IF_MODIFIED_SINCE']) && strtotime($_SERVER['HTTP_IF_MODIFIED_SINCE']) == $lastModifiedTime) {
+        header('HTTP/1.1 304 Not Modified', true, 304);
```
mvorisek, line 132: not sure if `http_response_code` is needed as well

public_html/static.php
```diff
@@ -0,0 +1,209 @@
...
+
+    // Potential hack attempts, don't allow ".."
+    if (strpos($path, '..') !== false) {
+        return null;
+    }
+
+    $ext = pathinfo($path, \PATHINFO_EXTENSION);
+
+    // Only supported file types
+    if (empty($ext) || !isset(SUPPORTED_TYPES[strtolower($ext)])) {
```
mvorisek, line 91: I do not think lowecase here is needed/wanted. Better to be strict, many FS as case sensitive anyway...

public_html/.htaccess
```diff
@@ -1 +0,0 @@
-../.htaccess
```
mvorisek, line 1: for some reasons I was not able to submit comment on `public_html/.htaccess` file...

`.htaccess` is supported on Apache only. As we do not do any mod rewrite here (except favicon which can be specified in html header to have different path), all these headers should be specified in route script to work on non-Apache webservers too.

public_html/static.php
```diff
@@ -0,0 +1,209 @@
...
+serveStaticFile($path);
+
+/**
+ * Validate that the file exists and can be served
+ *
+ * @param string $path File location
+ *
+ * @return ?string Verified and resolved file location
+ */
+function validateStaticFile($path): ?string
```
mvorisek, line 76:
```suggestion
 * @return ?string Verified and resolved file location
 */
function validateStaticFile(string $path): ?string
```

cosmetics only...

VVD:
Replaced value for `DocumentRoot` to `/usr/local/www/roundcube/public_html` and got error in browser:
```
Oops... something went wrong!

An internal error has occurred. Your request cannot be processed at this time.

For administrators: Please check the application and/or server error logs for more information.
```
There are no errors in logs.

VVD:
Fix:
```
update `system` set roundcube-version='2025092300';
```
All scripts from `SQL/mysql/*.sql` were run manually during previous updates.

kamzar1:
This PR has rendered all our auto installations for clients based on nginx and roundcube served from a subdirectory useless.
Forced us to move on to another webmail client.

PR #9296, opened 2024-01-01
Fix "increment_style" CS

mvorisek:


alecpl:
I don't like it. I know Symfony uses it, but Laravel does not. So, there's no wide consensus. Let's skip this change.

mvorisek:
Let's assert the "post" increment style then for now.

PR #9297, opened 2024-01-01
Fix CS (whitespace, visibility)

mvorisek:


alecpl:
I used to have a strong opinion about a newline before `else`, not anymore, so I'll accept that.

However, I don't like these extra blank lines. A blank line after a brace always makes sense, but not in other cases. It might sometimes make sense, but I would not force this rule.

mvorisek:
> A blank line after a brace always makes sense, but not in other cases. It might sometimes make sense, but I would not force this rule.

It makes sense to put a blank line before "returning statements" about the same as after `}` to improve the code readability.

This is actually "the more used CS" already, with `\n\n *(exit|continue|break|throw|return)` regex I found:
- 2050 matches before this PR
- 2555 matches after this PR

With this PR, the CS is unified and asserted by CI.

mvorisek:
@alecpl can this PR be merged please? I would like to finish other PRs based on it.

alecpl:
Sorry, it looks like I edited your comment instead of replying to it. It must have been late. So, again

I'm not convinced. Especially when there is a single line of code in a `case` or `if` body that extra line makes it worse.

mvorisek:
Let's merge this PR, I have reverted the `blank_line_before_statement` changes, let's discuss them in another PR and unblock this one.

PR #9299, opened 2024-01-01
oauth: refactorize request_access_token() returning a simple boolean

EdouardVanbelle:
Hello please find this small refactorization on OAuth module (remove unused parameters) as they are now stored during login_phase

### DIFF
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -365,7 +365,6 @@ class rcmail_oauth
         return $this->last_error;
     }
 
-
     /**
      * Callback for `loginform_content` hook
      *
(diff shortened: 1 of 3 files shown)


tests/Rcmail/Oauth.php
```diff
@@ -10,6 +10,11 @@ public function forge_login_phase($data)
     {
         $this->login_phase = $data;
     }
+
+    public function get_login_phase()
```
alecpl, line 14: We can use Reflection instead. See `getProperty()` and `setProperty()` in `tests/bootstrap.php`.

program/include/rcmail_oauth.php
```diff
@@ -707,7 +702,7 @@ public function request_access_token($auth_code, $state = null)
                 // store crypted code_verifier because session is going to be killed
                 $this->login_phase['code_verifier'] = $_SESSION['oauth_code_verifier'];
             }
-            return $this->login_phase;
+            return true;
```
alecpl, line 705: Let's move the other two `return false` statements to the end of the function.

alecpl (changes requested):

PR #9302, opened 2024-01-04
Fix "phpdoc_no_empty_return" CS

mvorisek:
The rule enforces developers to use native `void` return type.

In tests, this is easy to do, in other places, I reviewed the fixes to be back compatible as much as possible.

### DIFF
--- a/plugins/example_addressbook/example_addressbook_backend.php
+++ b/plugins/example_addressbook/example_addressbook_backend.php
@@ -74,7 +74,7 @@ class example_addressbook_backend extends rcube_addressbook
         return $this->name;
     }
 
-    public function set_search_set($filter)
+    public function set_search_set($filter): void
     {
         $this->filter = $filter;
     }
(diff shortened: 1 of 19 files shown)


plugins/example_addressbook/example_addressbook_backend.php
```diff
@@ -74,7 +74,7 @@ public function get_name()
         return $this->name;
     }
 
-    public function set_search_set($filter)
+    public function set_search_set($filter): void
```
mvorisek, line 77: this is small BC breaking change, it should be fine for master

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -127,10 +127,8 @@ abstract public function get_name();
      * operation.
      *
      * @param mixed $filter Search params to use in listing method, obtained by get_search_set()
-     *
-     * @return void
      */
-    abstract public function set_search_set($filter);
+    abstract public function set_search_set($filter): void;
```
mvorisek, line 131: this is small BC breaking change, it should be fine for master

PR #9303, opened 2024-01-04
Fix more CS

mvorisek:


mvorisek:
> * I don't like "php_unit_test_case_static_method_calls". I don't see a value.

Sooner or later is should be done as static calls should be called statically otherwise phpstan will warn about it. repro: https://phpstan.org/r/3c8d1087-b72c-41be-acf8-a2f72d4d788e

>     * With "php_unit_data_provider_name" change some method names look bad now, e.g. `provideParse_attrib_stringCases`. Either we change all method names (including test* ones) to camel-case or revert that.

It is managed by CS fixer thus even if the provider methods does not look best now, the CS is enforced and once the test methods are renamed to camel case, they will be updated automatically.

alecpl:
> > I don't like "php_unit_test_case_static_method_calls". I don't see a value.
> 
> Sooner or later is should be done as static calls should be called statically otherwise phpstan will warn about it. repro: https://phpstan.org/r/3c8d1087-b72c-41be-acf8-a2f72d4d788e

Only in "strict rules" mode. So, I'm not convinced.
 
> > With "php_unit_data_provider_name" change some method names look bad now, e.g. `provideParse_attrib_stringCases`. Either we change all method names (including test* ones) to camel-case or revert that.
>
> It is managed by CS fixer thus even if the provider methods does not look best now, the CS is enforced and once the test methods are renamed to camel case, they will be updated automatically.

But I don't want it in this incomplete state.

mvorisek:
I have reworked this PR.

`php_unit_test_case_static_method_calls` - the projects uses snake_case, thus yes, snake_case is more consistent

`php_unit_test_case_static_method_calls` - let's land this, we have already a lot of changes by other CS changes, so the best time to land this is now - calling static methods statically is definitely more wanted

mvorisek:
@alecpl can you merge this PR? I would like to fix other CS based on it.

alecpl:
1. I still don't like "php_unit_test_case_static_method_calls". I don't see a value.
2. I'm not sure about all these empty lines removed. Definitely sometimes inside a switch they make the code cleaner.

mvorisek:
I have reverted "php_unit_test_case_static_method_calls" changes. I stand by them, but let's discuss them in a separate PR to unblock this one.

PR #9310, opened 2024-01-18
Don't render quota if it's unlimited

### LINKED ISSUE #8994: Hide information about quota, when there is no quota
I have no limit of my mailbox.  Roundcube displays at the bottom of the mailbox an indication, that I have used 0% of my quota.

Even when I set `$config['quota_zero_as_unlimited'] = true;` the information is still there, with a tooltip “Without limitations”

When there is no quota, Roundcube shall not tell the user that he uses 0% of it, as this information unnecessary consumes space.

pabzm:
In case an accounts's quota is zero and the config option `quota_zero_as_unlimited` is true, then the account has no quota and should not be shown a meaningless (or even wrong) quota-display.

This introduces a new function to avoid repeating identical code in two places.

If a template doesn't check for the env-flag it still can render quota details.

This PR closes #8994.

One question and two notes:
* This doesn't include a Changelog-entry yet – should I add one or do you rather compile those yourselves?
* Also I didn't touch the templates, so the `.footer`-element of the column remains even if it is blank (because the `if`-clauses in the templates are inside that element) – I'm not sure if that needs to stay like that but didn't feel comfortable changing the template without knowing why it was written like this. And I think it looks okay, event with an empty element. If you want the footer gone, please let me know.
* And since there's no test case for this feature I didn't change any. Adding a fresh one is a little out of my time budget currently, because I'd need to dive into the setup first.

### DIFF
--- a/program/include/rcmail_action.php
+++ b/program/include/rcmail_action.php
@@ -165,6 +165,10 @@ abstract class rcmail_action
         // need to do that manually (the return value is `false` if the server
         // doesn't support it).
         $quota = $rcmail->storage->get_quota();
+        // Return early if the server doesn't support quotas anyway.
+        if ($quota === false) {
+            $rcmail->output->set_env('quota', false);
+        }
         $zero_as_unlimited = (bool) $rcmail->config->get('quota_zero_as_unlimited');
         // Show the quota display only if it's not unlimited.
         $show_quota = is_array($quota) && !($quota['total'] === 0 && $zero_as_unlimited);
--- a/program/include/rcmail_action.php
+++ b/program/include/rcmail_action.php
@@ -168,6 +168,7 @@ abstract class rcmail_action
         // Return early if the server doesn't support quotas anyway.
         if ($quota === false) {
             $rcmail->output->set_env('quota', false);
+            return;
         }
         $zero_as_unlimited = (bool) $rcmail->config->get('quota_zero_as_unlimited');
         // Show the quota display only if it's not unlimited.
--- a/program/include/rcmail_action.php
+++ b/program/include/rcmail_action.php
@@ -152,6 +152,26 @@ abstract class rcmail_action
     }
 
     /**
+     * Set env-flag for quota. It is used in templates to determine if quota
+     * data should be displayed.
+     * In case the quota is zero and the config option
+     * `quota_zero_as_unlimited` is true, then the account has no quota and
+     * should not be shown a meaningless (or even wrong) quota-display.
+     */
+    public static function quota_set_env()
+    {
+        $rcmail = rcmail::get_instance();
+        // This call also checks the QUOTA capability of the server, we don't
+        // need to do that manually (the return value is `false` if the server
+        // doesn't support it).
+        $quota = $rcmail->storage->get_quota();
+        $zero_as_unlimited = (bool) $rcmail->config->get("quota_zero_as_unlimited");
+        // Show the quota display only if it's not unlimited.
+        $show_quota = is_array($quota) && !($quota['total'] === 0 && $zero_as_unlimited);
+        $rcmail->output->set_env('quota', $show_quota);
+    }
+    
+    /**
      * Return HTML for quota indicator object
      *
      * @param array $attrib Named parameters
--- a/program/include/rcmail_action.php
+++ b/program/include/rcmail_action.php
@@ -162,22 +162,13 @@ abstract class rcmail_action
     {
         $rcmail = rcmail::get_instance();
         // This call also checks the QUOTA capability of the server, we don't
-        // need to do that manually (the return value is `false` if the server
-        // doesn't support it).
+        // need to do that manually.
         $quota = $rcmail->storage->get_quota();
-        // Return early if the server doesn't support quotas anyway.
-        if ($quota === false) {
-            $rcmail->output->set_env('quota', false);
-            return;
-        }
-        $zero_as_unlimited = (bool) $rcmail->config->get('quota_zero_as_unlimited');
-        
-        if($zero_as_unlimited === false) {
-        	$rcmail->output->set_env('quota', true);
-        	return;
-        }
-        // Show the quota display only if it's not unlimited.
-        $show_quota = is_array($quota) && !($quota['total'] === 0 && $zero_as_unlimited);
+        // $quota being `false` means that the server doesn't support quotas,
+        // an empty array means "unlimited", a `total` of zero means
+        // "unlimited" or "unknown"). In all these cases we don't want to show
+        // the element.
+        $show_quota = is_array($quota) && array_key_exists('total', $quota) && $quota['total'] > 0;
         $rcmail->output->set_env('quota', $show_quota);
     }
 
@@ -276,8 +267,11 @@ abstract class rcmail_action
                 $quota_result['table'] = $table->show();
             }
         } else {
-            $unlimited               = $rcmail->config->get('quota_zero_as_unlimited');
-            $quota_result['title']   = $rcmail->gettext($unlimited ? 'unlimited' : 'unknown');
+            // This shouldn't happen because unsupported, unlimited or unknown
+            // quota shouldn't be shown (see `quota_set_env()`), but we better
+            // not risk breaking third-party templates that didn't check
+            // `env:quota` before calling this method.
+            $quota_result['title']   = $rcmail->gettext('unknown');
             $quota_result['percent'] = 0;
         }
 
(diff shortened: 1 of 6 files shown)


pabzm:
Hm, I'll have a look at those tests

pabzm:
The remaining failures from the CI are not related to my changes, as far as I can see. Is there anything I can do about them?

pabzm:
@miaulalala, @ChristophWurst I only now realized that I didn't yet notify you of this pull request, sorry for that! I can't assign you as reviewers, though. Would you have a look?

program/include/rcmail_action.php
```diff
@@ -151,6 +151,26 @@ public static function table_output($attrib, $table_data, $show_cols, $id_col)
...
+    public static function quota_set_env()
+    {
+        $rcmail = rcmail::get_instance();
+        // This call also checks the QUOTA capability of the server, we don't
+        // need to do that manually (the return value is `false` if the server
+        // doesn't support it).
+        $quota = $rcmail->storage->get_quota();
+        $zero_as_unlimited = (bool) $rcmail->config->get('quota_zero_as_unlimited');
+        // Show the quota display only if it's not unlimited.
+        $show_quota = is_array($quota) && !($quota['total'] === 0 && $zero_as_unlimited);
```
miaulalala, line 170: It's not working for me since the $quota array is empty in my case. 

Something similar to this could work.

```suggestion
        $show_quota = empty($quota) || (is_array($quota) && !($quota['total'] === 0 && $zero_as_unlimited);
```

program/include/rcmail_action.php
```diff
@@ -151,6 +151,26 @@ public static function table_output($attrib, $table_data, $show_cols, $id_col)
...
+     * should not be shown a meaningless (or even wrong) quota-display.
+     */
+    public static function quota_set_env()
+    {
+        $rcmail = rcmail::get_instance();
+        // This call also checks the QUOTA capability of the server, we don't
+        // need to do that manually (the return value is `false` if the server
+        // doesn't support it).
+        $quota = $rcmail->storage->get_quota();
+        $zero_as_unlimited = (bool) $rcmail->config->get('quota_zero_as_unlimited');
```
miaulalala, line 168: since you have a config param here, you could check that first and early return if that config param is set to false.

```suggestion
        $zero_as_unlimited = (bool) $rcmail->config->get('quota_zero_as_unlimited');
        
        if($zero_as_unlimited === false) {
        	$rcmail->output->set_env('quota', true);
        	return;
        }
```

miaulalala (dismissed):

program/include/rcmail_action.php
pabzm: I had considered that we could also skip displaying the quota-element if there's no quota data to show. That goes a little beyond the original issue's content but seemed right to me, because currently in this case (empty quota array) the element is shown as if no quota were used at all and its title is "unknown", which to me feels like a rather bad fallback.

(At least that's the case in my setup. Is it the same in yours?)

I'll happily use your suggestion if you still think it's right, just wanted to explain my change.

program/include/rcmail_action.php
pabzm: Thank you for the hint! I added a second early check, which catches the case that the server doesn't event support quotas.

program/include/rcmail_action.php
miaulalala: It's empty for me - it still works and shows "unlimited".

miaulalala:
Isn't the config option a bit misleading? If I have an unlimited quota and explicitly set that option, imho the intention is to show "unlimited" in the quota display. Or am I misunderstanding this?

miaulalala:
> I think the config option is correct, because it only specifies that a quota of zero should mean unlimited.
> 
> But the consequence to then not show the element surprised me, too (only didn't dare to question the original decision).
> 
> Maybe showing the quota-icon and the text "unlimited" (or ♾️) instead of the bar and percentage would be an option?

sounds like a good idea. So cases are:

Quota is not supported - don't show anything

Quota is supported
- Unlimited with config option true - show infinity sign instead of bar / %
- Unlimited with config option false - don't show anything
- Not unlimited - keep current display

pabzm:
> Unlimited with config option true - show infinity sign instead of bar / %
> Unlimited with config option false - don't show anything

Why not show the infinity sign in all cases of unlimited quota?

Also, maybe there's still a misunderstanding: I think the config option tells that `quota["total"] === 0` should mean "unlimited", not "zero byte allowed". So it doesn't govern what to do with the info "unlimited" but only how to determine it in a special case.

Thus I'd suggest these cases:

- Quota is not supported - don't show anything
- Quota is unlimited - show infinity sign instead of bar / %
- Quota is not unlimited - keep current display

What do you think?

pabzm:
Note to self: an empty quota-list and thus empty `$quota`-array is not an error but means "all mailboxes unlimited" (according to [RFC 2087](https://datatracker.ietf.org/doc/html/rfc2087.html#section-5)).

pabzm:
> Also, maybe there's still a misunderstanding: I think the config option tells that $quota["total"] === 0 should mean "unlimited", not "zero byte allowed". So it doesn't govern what to do with the info "unlimited" but only how to determine it in a special case.

No, you're right: it does more than that, [it also governs if unlimited quota is titled as "unlimited" or "unknown"](https://github.com/roundcube/roundcubemail/blob/eeac27ea81cc27abbab88314c6dd5dc3440d38c7/program/include/rcmail_action.php#L249-L250), regardless of how the unlimited quota is determined. In that way it's actually also misleading, because you need to set it to get the title "unlimited" even if your IMAP-server doesn't send a quota-value of zero to mean "unlimited".

pabzm:
But the suggested changes would (silently) fix that misleading usage without hurting nor breaking anything, I think.

pabzm:
The coding style check fails doe to code we might throw away in the next iteration, and the failing test is unrelated (AFAICS), so I'll deal with those later if they still fail.

miaulalala:
> > Unlimited with config option true - show infinity sign instead of bar / %
> > Unlimited with config option false - don't show anything
> 
> Why not show the infinity sign in all cases of unlimited quota?
> 
> Also, maybe there's still a misunderstanding: I think the config option tells that `$quota["total"] === 0` should mean "unlimited", not "zero byte allowed". So it doesn't govern what to do with the info "unlimited" but only how to determine it in a special case.
> 
> Thus I'd suggest these cases:
> 
> * Quota is not supported - don't show anything
> * Quota is unlimited - show infinity sign instead of bar / %
> * Quota is not unlimited - keep current display
> 
> What do you think?

makes sense!

alecpl:
I don't like the idea of "infinity sign". Also, while I see  an unknown quota indication might makes some sense, I think that we should just hide the widget whenever the quota is unsupported, unknown or unlimited.

pabzm:
@alecpl I think it's helpful to let people know about their quota, also if it is unlimited. Not showing it looks more like a broken feature to me, especially if I think about users that might work with different accounts or servers, and then sometimes get to see the information and sometimes it's missing.

We could also use the translation of "unlimited" (as it currently is used for the `title`-attribute) instead of the infinity sign?

pabzm:
So now if quota is unsupported, unlimited or "unknown" it won't get shown.

This drops using the config option `quota_zero_as_unlimited`, because unlimited and "unknown" are handled identically.

As far as I can tell the Changelog is written by @alecpl, so I didn't touch it. Please let me know if you want a change there.

pabzm:
I could also squash the branch to get rid of the previous commits if you're ok with that.

alecpl:
Zero is not unknown, but I'm for dropping the option, unless someone has a good use case for quota=0.

miaulalala (approved):

pabzm:
Is there anything left for me to do in this PR?

(I don't mean to push, just to avoid misunderstandings.)

miaulalala:
> Is there anything left for me to do in this PR?
> 
> (I don't mean to push, just to avoid misunderstandings.)
Not from my side.

@alecpl , anything from your side?

alecpl:
- The diff conflict needs to be resolved
- This change has a small performance impact (extra `GETQUOTAROOT` commands). An alternative approach could be to make the quota element hidden on page load and only when set_quota command is being send to the client-side it could be shown (if needed).

pabzm:
> An alternative approach could be to make the quota element hidden on page load

Do you mean hidden by JavaScript, or by markup in the template?

The latter would be more robust but requires cooperation of the template, so admins providing different skins would need to change code, I guess – that might be a problem?

alecpl:
Yeah, from the template or skin css (I'm not sure which one is better), and then un-hidden in `setquota` event handler that skins use. So, it would require some changes in skins anyway. Also hiding the parent element would make sens too, this part definitely should be done in the skin itself only.

pabzm:
I made a new PR to keep things clean: https://github.com/roundcube/roundcubemail/pull/9366

pabzm:
Superseded by #9366

PR #9314, opened 2024-01-21
Fix "php_unit_test_case_static_method_calls" CS

mvorisek:
originally discussed in https://github.com/roundcube/roundcubemail/pull/9303#issuecomment-1879380051

This is non-functional change only, but I belive calling `TestCase::assertXxx` static methods as non-static like `$this->assertXxx()` should be replaced with `self::assertXxx()` because static methods should be simply called statically.

Later, we should require this using PHPStan across the whole project, for now, let's require this using PHP CS Fixer for (known) TestCase methods only.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -90,7 +90,6 @@ return (new Config())
             'suffix' => '_cases',
         ],
         'php_unit_method_casing' => false,
-        'php_unit_test_case_static_method_calls' => false,
         'psr_autoloading' => false,
         'strict_comparison' => false,
 
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -16,8 +16,7 @@ parameters:
         - '~^Variable variables are not allowed.~'
         - '~^Variable .* might not be defined\.~'
         - '~Call to function array_filter\(\) requires parameter #2 to be passed to avoid loose comparison semantics.~'
-        # relax strict rules - move to phpstan baseline once almost all l6 errors are fixed
-        - '~^Dynamic call to static method .+\.~' # TODO in https://github.com/roundcube/roundcubemail/pull/9314
+        - '~^Dynamic call to static method .+\.~'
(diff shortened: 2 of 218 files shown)


alecpl:
I generally don't like this idea. Whenever I have an object instance I prefer to use it instead of a static call. E.g.
```
$var1 = $object->method();
$var2 = $object->staticMethod();
```
looks better than
```
$var1 = $object->method();
$var2 = SomeClass::staticMethod();
```
Maybe if a class has static methods only... So, probably TestCase is such a case, but everything inheriting from it should be using static methods (and properties) only too.

mvorisek:
I would say you are used to that, but it is definitely less readable, as one call needs an instance, the other not. With `->method()`, you cannot tell just by looking at the code. Also, PHP CS Fixer needs `::` to be used to correctly make anonyous functions static which are optimized better as they does not need to be rebound before a call is made.

mvorisek:
PR is done and fixes most of the `^Dynamic call to static method .+\.` phpstan errors.

mvorisek:
PR is done and it fixes most of the `^Dynamic call to static method .+\.` phpstan errors.

mvorisek:
@alecpl can we merge this PR?

alecpl:
I don't like it. And I don't want any "risky" rules. Sorry.

mvorisek:
@alecpl might I know if we can find a common view on this?

When a method is a static, it has nothing to do with instance, thus `ClassName::methodName()` is the preffered way to call static methods. TBH `$instance->methodName()` is working because of historical support.

This PR is not even about https://github.com/roundcube/roundcubemail/pull/9314#issuecomment-1910757849 case you mentioned, this PR is about calling PHPUnit static method that are contained withing test cases using `self`.

`self` is semantically better and also allows the developer to read the code better as it communicates explicitly the code does not need any instance variables/properties.

Does this make sense to you and would you like me to rebase this PR? If yes, please reopen this PR and will update it.

Thank you and I hope you are doing well in the summer!

PR #9316, opened 2024-01-21
Fix CS - imports

mvorisek:


mvorisek:
Is the E2E CI failure related?

alecpl:
"Cannot declare class Tests\Browser\Browser because the name is already in use." in tests/Browser/Browser.php

mvorisek:
Thank you, I will fix the CS Fixer soon.

mvorisek:
Thank you. I did the first 2 commits in the past using older PHP CS Fixer verison... Now the PR is passing and I checked the changes againts latest fixer applied on master.

PR #9318, opened 2024-01-22
Fix more CS whitespace

mvorisek:


alecpl:
I don't like it. Also, I think it was a mistake to include `@PhpCsFixer` rules from the start. We should have started with `@PSR1` + `@PSR12` and then introduce more strict rules.

mvorisek:
What part do you not like?

no useless else - this improves code by reducing indentation, this is natural and mostly done already, newly, the remaining code is fixed and asserted by CI

blank line before statement - I know you do no like it much, so I skipped this for `continue` and `break` but for `return`, `throw` etc. it is good and make again the code more readable, as these are code flow exiting/returning statements. Again, mostly done naturally, now remaining code is fixed and asserted by CI.

> Also, I think it was a mistake to include `@PhpCsFixer` rules from the start. We should have started with `@PSR1` + `@PSR12` and then introduce more strict rules.

Why? `@PhpCsFixer` is "optimal ruleset" in sense of CS strictness and PHP CS Fixer possibilities. Of course, starting with this ruleset will be wrong, that is why I added it, but disabled all the rules, so be we land the changes in reviewable parts. So effectively, we do it this way.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -688,6 +688,7 @@ public static function compose_list_name($contact)
                     $result[] = $contact[$field];
                 }
             }
+
```
alecpl, line 691: You see, here it makes sense, i.e. after "}", but not in other places. We don't have to be that strict. I see no value in being that strict.

program/lib/Roundcube/rcube_config.php
```diff
@@ -576,12 +577,12 @@ public function header_delimiter()
             $delim = $this->prop['mail_header_delimiter'];
             if ($delim == "\n" || $delim == "\r\n") {
                 return $delim;
-            } else {
```
alecpl, line 579: no_useless_else=true is all right. I prefer that too.

program/lib/Roundcube/rcube_addressbook.php
mvorisek: Done - I patched BlankLineBeforeStatement fixer and applied it.

PR #9326, opened 2024-01-28
Add orphan inline images below the message body

bohwaz:
* In plaintext mode: all inline images will be displayed below the message body
* In HTML mode: all inline images that are **not** referenced in the message HTML body will be listed below the message body

This doesn't add more CPU usage as it is using the already called HTML filtering function to discover if an inline part is referenced or not.

This fixes issue #5051 and is a bit better than #9150

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -985,10 +985,10 @@ class rcmail_action_mail_index extends rcmail_action
      * Convert the given message part to proper HTML
      * which can be displayed the message view
      *
-     * @param string             $body Message part body
-     * @param rcube_message_part $part Message part
-     * @param array              $p    Display parameters array
-     * @param array  $used_cids    List of CIDs appearing in message body
+     * @param string             $body      Message part body
+     * @param rcube_message_part $part      Message part
+     * @param array              $p         Display parameters array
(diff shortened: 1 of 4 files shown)


bohwaz:
Not sure why CI is failing, but it's not on any code I changed in this PR.

alecpl:
I think this is wrong because you display the images even if `inline_images` is `false`. With this option disabled they should appear on the attachments list, not below the body.

bohwaz:
> I think this is wrong because you display the images even if `inline_images` is `false`. With this option disabled they should appear on the attachments list, not below the body.

@alecpl As far as I understand the code of Roundcube, it is not possible to do what you want currently as the message body is parsed *after* the display of the attachments, so there is no way to know if the message parts are not in the message body.

This is why I'm appending the parts below the message and not in the message list. Please suggest a code fix if I'm wrong as I'm not familiar with the internals of Roundcube.

bohwaz:
I think your PR solves this issue as well, but I have zero time to invest in this currently. I'll check this later maybe.

Neustradamus:
@bohwaz: Can you check your PR?

bohwaz:
Don't have time right now, sorry. Feel free to do so :)

PR #9328, opened 2024-01-30
Fix some basic JS CS

mvorisek:
no functional change


plugins/enigma/enigma.js
```diff
@@ -199,16 +200,19 @@ rcube_webmail.prototype.enigma_key_create_save = function () {
     // validate the form
     if (!password || !confirm) {
         this.alert_dialog(this.get_label('enigma.formerror'));
+
```
alecpl, line 203: We didn't do this for PHO and I don't like it here too.

plugins/enigma/enigma.js
```diff
@@ -304,28 +309,28 @@ rcube_webmail.prototype.enigma_export = function (selected) {
...
                 text: this.get_label('enigma.onlypubkeys'),
                 click: function (e) {
                     rcmail.enigma_export_submit(args);
                     $(this).remove();
-                }
+                },
             },
             {
-                'class': 'export',
+                class: 'export',
```
alecpl, line 320: I'm afraid that this will not work. 'class' is a keyword in javascript.

plugins/enigma/enigma.js
```diff
@@ -532,7 +543,7 @@ rcube_webmail.prototype.enigma_add_list_row = function (r) {
 
 
 /*********************************************************/
-/*********        Enigma Message methods         *********/
+/* ********        Enigma Message methods        *********/
```
alecpl, line 546: This looks wrong.

mvorisek: There is a rule like in php, it should be fine.

alecpl: That added space does not make sense. Look at the line above (with all asterisks). If we can't fix it we might consider removing these whole comment blocks.

mvorisek: that is https://github.com/roundcube/roundcubemail/commit/f9b3ba60e3ba19c4153fdf46e1a85e4bc73541b0

IMO the comments can be helpful to place a new methods more logically as the files itself are quite large.

We did the same with PHP code, we simply made the fixer happy.

Thus I prefer to merge this PR as is.

PR #9330, opened 2024-01-31
Fix binary operator spaces CS

mvorisek:


mvorisek:
There are two styles we can use.

```
'binary_operator_spaces' => [
    'default'   => 'single_space',
    'operators' => [
        '=>' => 'align_single_space_minimal',
        '='  => 'align_single_space_minimal',
    ],
],
```
or
```
'binary_operator_spaces' => [
    'default'   => 'single_space',
],
```

docs: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.48.0/doc/rules/operator/binary_operator_spaces.rst

This project use more the 1st style (1st/2nd commit), but it is not diff friendly, not refactoring friendly and small changes very often leads to many LoC changed. Also, the aligned indentation on unrelated assignments is quite confusing.

I personally prefer the 2nd style (3rd commit) for the reasons said above, it is also the default Symfony and PHP CS Fixer style.

@alecpl what style do you prefer?

alecpl:
I'm fine with single_space.

mvorisek:
PR is done with single_space.

PR #9353, opened 2024-02-13
Improve phpstan rules

mvorisek:
- enable phpstan bleading edge for better detection/explanation
- enable phpstan strict rules, but disable some rules that are too strict for now


alecpl:
I really don't like it. I don't want our code to be so strict. Also, I don't like the baseline approach. This moves the target far away, this is discouraging.

mvorisek:
in https://github.com/roundcube/roundcubemail/compare/c6c1b96e306040d1de38c7ada9628b7b42cc31d6...c50ff182f68a8a56ea6e890adbd11a4e6ca868da you can see why bleading edge is helpful to better understand the already baselined issues

strict rules - I disabled most of them, but some like undefined property, not covariant with parent type, implicit array creation, ... errors are helpful and should be fixed, what exactly do you not like?

Like with CS, we enable the rules but disable the most strict one for now and put the "to be fixed next" in a baseline, so they do not have to be fixed in one commit and new contributions are checked againts these rules.

mvorisek:
@alecpl I would like to find a commonly welcomed solution, did the informations in my post above explained the advantages of this PR or is there still some rule you do not like? I would be happy if we can merge this PR, as I would like to work on it and fix some issues.

alecpl:
I'd prefer to settle on level:4 and no strict rules.

Issue #6510, opened 2018-11-02
Added meta for webapp in layout.html


gloony:
Since 1.4 allows a mobile interface, it might be interesting to add web-app definitions in the elastic skin.

For Android and iOS this would be this (_on layout.html, between head tags_) : 
```
	<meta name="apple-mobile-web-app-capable" content="yes"/>
	<meta name="apple-mobile-web-app-status-bar-style" content="black"/>
	<meta name="apple-mobile-web-app-title" content="<roundcube:object name="productname" />"/>
	<link rel="apple-touch-icon" href="skins/images/elastic/logo.png"/>
	<link rel="icon" href="skins/images/elastic/logo.png" />
```

This need a png icon for that, while svg is not supported as an icon on most device.

alecpl:
That could be nice. Because it requires (good quality) png icon it should be optional. Also, for Chrome meta tags without apple- prefix should be added.

It would make sense with forcing Elastic skin on mobile. So, maybe this should become a part of the elastic4mobile plugin. @thomascube

alecpl:
See also https://realfavicongenerator.net/

alecpl:
Another related link https://git.kolab.org/diffusion/RPK/browse/dev%252Fpwa/plugins/pwa/

jonmoesli:
The last commit on kolab repository is about one year ago. Is a dead project now? :-(

LoneFenris:
If this were done as a PWA, it likely would offer a path towards doing system notifications as well, I think.

Daniel-Abrecht:
I would appreciate PWA support as well. I have a linux phone, and of all mail clients, roundcube so far works best.

PR #9014, opened 2023-05-18
Allow to add Roundcube as app to supported browsers using manifest.json

rstefko:

PR #7709, opened 2020-11-04
new option for configurable cookie lifetime

emtiu:
Implementing #5050 by introducing a new config option `cookie_lifetime`.

The default is the current behavior (i.e. set session cookies with lifetime 0).

PR #9354, opened 2024-02-14
pwa support for elastic skin

the0ne:
As progressive web app support is great nowadays on all major operating system platforms, roundcube can be used even more convenient if it supports this standard. Roundcube can look like it was installed as a 'real' app, have a proper launcher icon, etc.
Even notifications would be possible if the service worker inside sw.js is further enhanced in the future.

This PR should solve #6510.


skins/elastic/templates/includes/layout.html
```diff
@@ -39,6 +39,47 @@
 		} catch (e) { }
 		</script>
 	<roundcube:endif />
+	<roundcube:if condition="env:action != 'print' && !config:devel_mode" />
+		<link rel="manifest" href="/manifest.json">
```
mvorisek, line 43: one high-res icon should be enough, also it is much easier to replace it with a custom icon then

the0ne: You'd have to discuss this point with the guys that defined PWA I guess 😊

alecpl:
I don't think I'm gonna accept this. Maybe it should/could become a plugin.

People really need a way to replace the icon easily. So, maybe we could generate all icon sizes from a single icon when they are requested.

Also things like `"application-name": "HS Webmail",` can't be there.

the0ne:
Yeah, the application name clearly wasn't meant like this.
I've updated it to Roundcube Webmail but obviously you can put there whatever you like.
Regarding the images, is there some mechanism in RC that could be used for this?

alecpl:
`rcube_image::resize()`

the0ne:
> `rcube_image::resize()`

😂👍🏻

I don't think that's gonna cut it... 😉


Anyway, I've made this PR as a proposal on how this whole feature can easily be implemented.

I don't feel you're particularly interested anyway and I don't intend to play a game where you request random changes and expect me to search for solutions in your software.

So just let me know and I'll happily cancel the PR if that's your wish. Thanks! 🙂👌🏻

miaulalala:
Will need something link https://github.com/roundcube/roundcubemail/pull/7709

the0ne:
Wasn't an issue on the installation I host because anyway I got the 'persistent_login' plugin installed which takes care of this automatically.

skins/elastic/sw.js
```diff
@@ -0,0 +1,76 @@
...
+
+/*
+create a list (array) of urls to pre-cache for your application
+*/
+
+/*  Service Worker Event Handlers */
+
+self.addEventListener( "install", function ( event ) {
+
+    console.log( "Installing the service worker!" );
```
pabzm, line 14: Please remove the logging from this file (or make it helpful beyond debugging).

pabzm: Actually this sets up caching network data, but that data doesn't ever get used, right? So please either complete that feature or remove it.

skins/elastic/templates/includes/layout.html
pabzm: https://developer.apple.com/design/human-interface-guidelines/app-icons#App-icon-sizes says:
> You can let the system automatically scale down your 1024x1024 px app icon to produce all other sizes

For MS Windows and Android, at least 512x512 pixels are required. So to me, too, this sounds like way fewer icons are possible to make it work.

I guess for optimal icon displaying more sizes are better, but for this PR to be mergeable we need to find a solution that keeps changing the icon feasible. (Nothing stops anyone from adding more icons at all.)

@the0ne Would you be willing to change this to reach a possible compromise?

pabzm:
FTR: #9014 is "prior art", but much less elaborate.

skins/elastic/templates/includes/layout.html
the0ne: I'd be really happy if somebody would step in and take care of those changes. 
In the next several months (or even longer) I won't be able to take care about this.

pabzm: Ok, thanks for the feedback!

skins/elastic/sw.js
```diff
@@ -0,0 +1,76 @@
...
+
+            return;
+
+        } )
+
+    );
+
+} );
+
+self.addEventListener( "fetch", function ( event ) {
```
Daniel15, line 55: Roundcube doesn't use `fetch` in most of its code (it's older code that uses XMLHttpRequest), so this won't actually do anything.

Daniel15:
IMO the service worker should be removed from this PR and done in a separate one. Offline access is a major feature of PWAs, but it'd be non-trivial to implement.

Adding the metadata and icon would be beneficial. Maybe this could be split into three PRs:
1. Add the metadata
2. Add the icon
3. Implement offline support using the service worker

pabzm:
Since the original author [stated](https://github.com/roundcube/roundcubemail/pull/9354#discussion_r1618553269) that he(?) won't work on this further and no one stepped up, I'll close this.

If anyone is interested to resume working on it, please reopen.

the0ne:
Why would you close this?
Closing the pull feels like there's not the right spirit here.

You only have a few remarks, but to solve those I'd have to learn soooo much about all the internals of RC.
It would be very easy if we could work on this together:
* I've voluntarily taken over the part of the general technic behind
* you could easily adapt it with possibly very minimal effort, making this a huge improvement of RC

I really don't get why you'd close this PR - is this about "statistics only" for you?

pabzm:
It's not at all about statistics, it's about reducing the mental load of open threads. Having too many open pull requests that no one is working on drowns other PRs that would need the attention, I feel. Thus I closed it with the explicit remark that it can be re-opened – if someone steps up to work on it.

I don't quite understand if you want to work on it but just need better feedback than before, or if you just think someone else should do it. In the former case, please let me know, I'd be happy to re-open and help!

the0ne:
Hi @pabzm if I understood the above right, it's about few things only:
* there's some setting requested to add some icon that should be resized automatically to different sizes
* there's some setting requested to give an individual name to the PWA
* forget about the offline stuff, the service worker wasn't meant for that actually but can be used for possible pre-caching

It's clear that it would take very much effort for me to learn about all the internals of RC whereas the actual implementation would probably take only a few hours.
I didn't have much time in the past to implement stuff in RC, so I don't see what sense it would make for me to take 10-15 days learning RC when some experienced RC developer could add the few missing things in a matter of hours.

pabzm:
@the0ne Well you convinced me that you do care about this feature, so I'll re-open even if it's unclear who's going to work on this :)

michellhdz (approved):

PR #9356, opened 2024-02-15
invert the display_next flag if user holds the shift modifier key while deleting a message

the0ne:
When deleting a message, usually I don't want to go to the next message, but sometimes I want to.
Other users might want to usually go to the next message, but sometimes they possibly don't want to.

With this small addition, this can be easily achieved:

1. Set your preference of "display_next" as usual.
2. Whenever you want the opposite behaviour when deleting a mail, just hold the 'Shift' modifier key while doing so.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -3517,11 +3517,19 @@ function rcube_webmail() {
 
     // delete selected messages from the current mailbox
     this.delete_messages = function (event, uid) {
-        var list = this.message_list, trash = this.env.trash_mailbox;
+        var list = this.message_list, trash = this.env.trash_mailbox,
+            display_next = this.env.display_next;
+
+        // invert the display_next flag in case shift is pressed
+        if ((list && list.modkey == SHIFT_KEY) || (event && rcube_event.get_modifier(event) == SHIFT_KEY)) {
+            this.env.display_next = !display_next;
+        }
 
         // if config is set to flag for deletion
         if (this.env.flag_for_deletion) {
             this.mark_message('delete', uid);
+
+            this.env.display_next = display_next;
             return false;
         }
         // if there isn't a defined trash mailbox or we are in it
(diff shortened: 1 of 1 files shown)


alecpl:
It does not make sense to me. Shift key already has a meaning on delete.

the0ne:
> Shift key already has a meaning on delete.

Ah, yes - I see. The shift key is used for a specific case where it's about deleting without moving to trash. I didn't notice this as it's clearly not my use case, I always have 'flag_for_deletion' active.

> It does not make sense to me.

You probably don't use the mail list in the way I do. By keeping mail unread it's marked as 'open' with the meaning of 'not yet done'. Often I need to delete newer (server log) emails but keep the newest message unread. So either I'd have to turn display_next on, then I'd have to remember to every time manually set the newest message to unread again. Or I'd have to turn display_next off, then I'd have to manually go to the next server log email after deleting one of them. Both is not convenient. In horde, the behavior was exactly like it's with the patch now, which I like a lot as a very convenient solution.

Maybe I could change this so that the new code is only active if 'flag_for_deletion' is active - would you like that better?

alecpl:
If 'flag_for_deletion' is active 'display_next' does not apply as the message is not being removed from the list. So, it does not make much sense. In this mode you can just press `Del` key and then `Arrow-Down` if you wish or not.

That being said, maybe use `Ctrl` key instead of `Shift` and we're good. I don't know. This may come in handy with #9357 if we decide it's a good idea.

the0ne:
> It does not make sense to me. Shift key already has a meaning on delete.

What meaning is that?

alecpl:
As you've find out earlier it's "delete without moving to Trash".

the0ne:
@alecpl thanks for the reminder - I don't know where I had my head.
If you prefer to have this on [Ctrl] modifier key, of cause it would work for me as well 👍🏻

PR #9361, opened 2024-02-17
add github issue templates

johndoh:
It seems to be getting more common for end users to be opening github issues when they should be contacting their email provider. Perhaps we could use github issue templates to try and nudge people in the right direction?

You can see a more friendly preview of the templates [here](https://github.com/johndoh/roundcubemail/tree/issue_templates/.github/ISSUE_TEMPLATE)


alecpl:
We either do not assign any labels or introduce "Needs Triage" label.

johndoh:
i have removed the labels, since the templates are new i thought may be its better to start with none and then later if some would be useful they can be added

pabzm:
Looks good, thank you very much!

pabzm (approved):

PR #9362, opened 2024-02-18
elastic: expose message_list_hover_menu functions

johndoh:
much much simpler alternative to #9038. simply expose the JS methods needed for the hove menu so an extended skin or a plugin can override them


alecpl:
I like simple solutions, but maybe introducing some events instead would be better? We already do this in `resize()`.

johndoh:
yes thats even simpler, nice idea

PR #9366, opened 2024-02-21
Show quota only if sensible

pabzm:
If a quota is not supported, or its value is unknown or unlimited, don't show the info element.
This approach simply hides the element until a value is added via JavaScript. The change looks bigger than it actually is because of indention (whitespace) changes.

### DIFF
--- a/program/include/rcmail_action.php
+++ b/program/include/rcmail_action.php
@@ -190,65 +190,63 @@ abstract class rcmail_action
         $quota = $rcmail->storage->get_quota($folder);
         $quota = $rcmail->plugins->exec_hook('quota', $quota ?: []);
 
+        if (empty($quota['total']) || $quota['total'] <= 0) {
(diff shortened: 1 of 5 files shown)


pabzm:
Force-pushed to rebase onto the current `master`-branch and resolve conflicts.

pabzm:
Another clarification to avoid misunderstanding: this PR only changes that if a quota is set but it is zero, then the column footer looks as if no quota was set (i.e. it is a blank rectangle).

pabzm:
I think the failing CI tasks aren't caused by these changes.

miaulalala (approved):

pabzm:
@alecpl Do you have an opinion on this? I tried to meet your expectations from https://github.com/roundcube/roundcubemail/pull/9310#issuecomment-1944286187 but would like a confirmation.

PR #9389, opened 2024-03-18
Minor correction to Serbian translation.

imilos:


alecpl:
Be aware that if this is not done in Transifex it might get overwritten on the next localization sync.

imilos:
No, it is not. I will do it on Transifex. Thanks for the information.

Kind regards
Milos


нед, 24. мар 2024. у 08:02 Aleksander Machniak ***@***.***>
је написао/ла:

PR #9406, opened 2024-04-07
Fix "Implicit array creation" in plugin config files

mvorisek:
extracted from #9241

"Implicit array creation" PHPStan error is otherwise present once the plugins are installed.

Also require Roundcube for all the bundled plugins - https://github.com/roundcube/plugin-installer/pull/44.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "pear/net_sieve": "~1.4.5",
         "pear/net_smtp": "~1.10.0",
         "pear/pear-core-minimal": "~1.10.1",
-        "roundcube/plugin-installer": "~0.3.1",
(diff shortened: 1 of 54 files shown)


alecpl:
Looks good with one note. The plugins repository was left there for backward compat. with plugins that do not exist on Packagist yet. So, again, this part probably would be easier to accept if we do a bigger jump in version number (v2.0).

mvorisek:
I read somewhere the plugins repository is no longer in operation...

I have reverted https://github.com/roundcube/roundcubemail/commit/bd17e2ed97002c887f91d8d3df99c1b307b153b0. Can this PR be merged now?

PR #9407, opened 2024-04-07
Allow composer autoloading (to coexist)

mvorisek:
extracted from #9241

This PR is 100% BC and does not imply any functional change.

This PR allows composer autoloading to address the goals from #9241, but keeps the Roundcube autoloading intact.

Later, the transition to composer autoloading can be discussed in the original #9241 with minimal LoC needed.


mvorisek:
PR is done and does not intact the current autoloading in any way. However, Roundcube/plugins can be newly easily used using composer as deps 🎉🎉🎉!

alecpl:
Looks good. I wonder whether we could do something that does not look like a hack regarding that change in `rcmail_oauth`. How about skipping `rcmail_oauth` instantiation in `rcmail::startup()` if `PHP_SAPI == 'cli'`?

mvorisek:
> Looks good. I wonder whether we could do something that does not look like a hack regarding that change in `rcmail_oauth`. How about skipping `rcmail_oauth` instantiation in `rcmail::startup()` if `PHP_SAPI == 'cli'`?

I do not support this. The reason is I belive it should be initialized for every environment (web & CLI).

Here we solve the first composer install/update only. The reason is composer classmap autoloader has not discovered the classes yet (when the roundcube plugin installer composer plugin is called). For consequent run, the class is autoloadable/present, thus disabling this for CLI in general is not needed and IMO not wanted.

alecpl:
It would also break oauth tests. So, maybe the `ExtensionInstaller` should include `iniset.php` instead of `clisetup.php`. I.e. not initialize `rcmail` at all. I'm not sure whether this would just work or some more changes would be needed.

mvorisek:
Is it ok to merge this as is? With the current workaround, there is no behavioural change, so we should not actually try to fix this in this PR.

mvorisek:
@alecpl now this PR is complete and fully BC. Please merge it and submit this repo to Packagist. I will then take care of the remaining changes.

alecpl:
ps. `roundcube/roundcubemail` has been registered on Packagist.

Issue #9377, opened 2024-03-04
Provide interfaces for Plugins exposing an API


jaudriga:
[audriga/roundcube-jmap](https://github.com/audriga/roundcube-jmap/) is a plugin that exposes a JMAP API focusing on Contacts and Calendars within Roundcube. Currently, this involves duplicating some of Roundcube's index.php code, which is not ideal for maintenance and compatibility.

It would be helpful if roundcube could provide some interfaces or hooks for plugins that want to expose an API. Presently, our implementation involves:
* Including environment via program/include/iniset.php.
* Authentication by caling the authenticate hook and subsequently `rcmail->login()`.

Including the environment file might already be sufficient. However, passing basic-auth credentials to the authenticate hook and the subsequent login is currently not ideal. One requirement for us would be that this is also compatible with Roudcube installations in cPanel.

**Additional Information**:

All code that audriga/roundcube-jmap duplicated from the index.php can be seen in [bridge.php](https://github.com/audriga/roundcube-jmap/blob/878ea7e7a9d370980bbc8374d01d4bcba9637589/bridge.php).

This issue is related to the discussions in [#5918](https://github.com/roundcube/roundcubemail/issues/5918).

Another example of a plugin that provides an API is freebusy, which you can find here: https://git.kolab.org/source/freebusy/.

alecpl:
There's the `startup` hook. I'm not sure what else would you need. Provide a pull request.

jaudriga:
I opened up a #9418 . I hope this makes it clearer what I meant initially. The JMAP RC Plugin is not a plugin in a strict sense, it follows a similar approach as Kolab's FreeBusy plugin, but uses the `rcmail` class to reuse authentication logic.

PR #9418, opened 2024-04-17
Provide imap auth interface for plugins

jaudriga:
Addresses #9377 .

Main change: Allows to use the `rcmail::login()` function for merely logging into IMAP by introducing a new flag `$just_connect`. This allows applications using the Roundcube Framework to reuse IMAP authentication logic.

Background: It would be beneficial for applications like Roundcube JMAP (see https://github.com/audriga/roundcube-jmap/pull/6 ) or Kolab's FreeBusy to reuse large parts of the existing IMAP Authentcation logic of Roundcube. This is my try at providing an interface for reusing that.

Minor additional change: Adds a function in rcube to connect to IMAP.  I actually consider this optional as its only beneficial for FreeBusy for now. This would basically allow FreeBusy to just call `rcube::imap_connect()` instead of what they are doing right now (from [lib/Kolab/FreeBusy/SourceIMAP.php](https://git.kolab.org/source/freebusy/browse/master/lib/Kolab/FreeBusy/SourceIMAP.php$298)):

```
// enable proxy authentication
if (!empty($config['proxy_auth'])) {
	$imap->set_options(array('auth_cid' => $config['proxy_auth'], 'auth_pw' => $config['pass']));
}

// authenticate user in IMAP
if (!$imap->connect($host, $config['user'], $config['pass'], $port, $ssl)) {
	Logger::get('imap')->addWarning("Failed to connect to IMAP server: " . $imap->get_error_code(), $config);
	return false;
}
```

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -648,14 +648,15 @@ class rcmail extends rcube
      * Perform login to the mail server and to the webmail service.
      * This will also create a new user entry if auto_create_user is configured.
      *
-     * @param string $username    Mail storage (IMAP) user name
-     * @param string $password    Mail storage (IMAP) password
-     * @param string $host        Mail storage (IMAP) host
-     * @param bool   $cookiecheck Enables cookie check
+     * @param string $username     Mail storage (IMAP) user name
+     * @param string $password     Mail storage (IMAP) password
+     * @param string $host         Mail storage (IMAP) host
+     * @param bool   $cookiecheck  Enables cookie check
+     * @param bool   $just_connect Breaks after successful connect
      *
      * @return bool True on success, False on failure
      */
-    public function login($username, $password, $host = null, $cookiecheck = false)
+    public function login($username, $password, $host = null, $cookiecheck = false, $just_connect = false)
     {
         $this->login_error = null;
 
@@ -773,6 +774,15 @@ class rcmail extends rcube
             sleep(1);
             return false;
         }
+
+        // Only set user if just wanting to connect
+        if ($just_connect) {
+            if (is_object($user)) {
+                $this->set_user($user);
+            }
+            return true;
+        }
+
 
         // user already registered -> update user's record
         if (is_object($user)) {
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -765,24 +765,18 @@ class rcmail extends rcube
         $storage = $this->get_storage();
 
         // try to log in
-        if (!$storage->connect($host, $username, $password, $port, $ssl)) {
-            if ($user) {
-                $user->failed_login();
-            }
-
-            // Wait a second to slow down brute-force attacks (#1490549)
-            sleep(1);
+        if (!$this->imap_connect($storage, $host, $username, $password, $port, $ssl, $user)) {
             return false;
         }
 
-        // Only set user if just wanting to connect
+        // Only set user if just wanting to connect.
+        // Note that for other scenarios user will also be set after successful login.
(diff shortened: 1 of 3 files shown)


program/include/rcmail.php
```diff
@@ -763,16 +764,18 @@ public function login($username, $password, $host = null, $cookiecheck = false)
...
-                $user->failed_login();
-            }
-
-            // Wait a second to slow down brute-force attacks (#1490549)
-            sleep(1);
+        if (!$this->imap_connect($storage, $host, $username, $password, $port, $ssl, $user)) {
             return false;
         }
 
+        // Only set user if just wanting to connect
```
pabzm, line 771: Could you maybe extend this comment a little bit?

I stumbled over it, because of course we also want to set user in other cases, too – and it's done, just a little later. Whenever I find myself stumbling like this I imagine it will happen to others as well, or to me again at a later point in time, and would like to increase the clarity.

pabzm (changes requested):
Besides the one request about the comment I find this well done and would merge it (which I don't decide on my own), thanks a lot for the thoughful and careful way to introduce a new feature!

pabzm:
@jaudriga After reading the corresponding ticket I'm wondering you want this actually merged or if it's more of a proof of concept to illustrate your broader request?

program/include/rcmail.php
jaudriga: I hope its a bit clearer now.

pabzm (approved):

jaudriga:
I rebased on newest master as on of the CI jobs failed: https://github.com/roundcube/roundcubemail/actions/runs/9059543070/job/24905660338

pabzm:
The CI will be fixed with #9449, please rebase again after that was merged.

jaudriga:
I rebased again ;)

pabzm:
@alecpl Any thoughts on this?

pabzm:
@alecpl Please have a look at this and provide feedback, or let us know that you're fine with us proceeding without your involvement.

alecpl:
I don't like it as it is.
1. It should be `storage_connect` not `imap_connect`. The existing `rcmail::storage_connect()` would need to be renamed (or moved to `rcmail_action`).
2. I think this new `storage_connect()` should contain more code from `login()`, i.e. lines from 751-809. Possibly also `set_user()` and `set_storage_prop()` calls.
3. Then that method could return `rcube_user` on success.

jaudriga:
> I don't like it as it is.

Thanks @alecpl for taking a look at the PR. I came around to addressing your feedback today.

Your first suggestion was relatively straight-forward to address IMHO.

I intentionally left out returning an `rcube_user` object in the first iteration of the PR to avoid as much refactoring as possible. However, I agree that its more well-designed to return `rcube_user` in the new `storage_connect`.

The refactoring that I now did is moving the error codes from the `rcmail` into the `rcube` parent class. I chose this refactoring, because the `$user` is checked for brute-force prevention directly after querying for the username. This potentially sets the error to `ERROR_RATE_LIMIT`. Alternative to this refactoring I had the ideas of either throwing a novel Exception within `storage_connect` to signal `ERROR_RATE_LIMIT` or duplicating the `rcube_user::query()` call.

Please take a look if the newest changes are fine for you.

alecpl (changes requested):

thibaultmol:
https://drewdevault.com/2021/10/26/stalebot.html

jaudriga:
@pabzm I am still interested in this, but sadly do not have the time to work on it right now.

PR #9424, opened 2024-04-22
Keep phpstan strict rules testing

mvorisek:
Keep phpstan strict rules for CI.

Too strict rules with many matches should be ignored by regex instead, but overall the quiality should be asserted, the rules are great!


mvorisek:
@alecpl can this PR be merged?

alecpl:
I appreciate your commitment, but I'm against this change.

mvorisek:
Given all the detected issues are valid, I kindly ask if this PR can be merged in order to maintain the project in unarguably a better shape.

In my opinion, the few needed `@phpstan-ignore` comments (of real issues to be fixed someday) are worth and the few ignores of too strict errors in phpstan config does not hurt either.

pabzm:
@alecpl Can you explain why you're against this?
I don't know if I agree with every rule, but in general I would prefer to have advanced automated syntax/style checking, as it helps contributors to conform to the wanted rules and thus requires a little less attention from you (or the team).

mvorisek:
@alecpl Can you explain why you're against this?

In https://github.com/roundcube/roundcubemail/pull/9424#issuecomment-2144897599 I have summarized the reasons why the changes in this PR are good and help to maintain this project in better shape. It is fine to ignore some too strict rules, but not all of them, some are legit.

mvorisek:
@alecpl I have rebased this PR, let's merge it.

alecpl:
Ok, let's do it. After the conflict is resolved.

mvorisek:
@alecpl conflicts are resolved

PR #9430, opened 2024-05-01
contacts: add more gender options

johndoh:
add `non-binary` and `other` to the existing `male` and `female` options for the gender field in contacts

### DIFF
--- a/program/actions/contacts/index.php
+++ b/program/actions/contacts/index.php
@@ -120,6 +120,8 @@ class rcmail_action_contacts_index extends rcmail_action
             'options' => [
                 'male' => 'male',
                 'female' => 'female',
+                'non-binary' => 'nonbinary',
+                'other' => 'other',
             ],
         ],
         'maidenname' => [
--- a/program/actions/contacts/index.php
+++ b/program/actions/contacts/index.php
@@ -121,7 +121,7 @@ class rcmail_action_contacts_index extends rcmail_action
                 'male' => 'male',
                 'female' => 'female',
                 'non-binary' => 'nonbinary',
-                'other' => 'other',
+                'other' => 'othergender',
             ],
         ],
         'maidenname' => [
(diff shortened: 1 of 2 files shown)


pabzm:
Thank you for the contribution! Would you maybe include another label for "other"? The current value for that string in `en_US` is capitalized, and other languages might have different words for the gender-specific meaning of "other".

miaulalala (approved):

alecpl:
While "other" is mentioned in https://www.rfc-editor.org/rfc/rfc6350.html#section-6.2.7, I don't see "non-binary" there. There is "none or not applicable" which I don't think can be used for the same purpose. So, I'm asking for reasoning about "non-binary" option. It's important to keep it standard-compliant/interoperable.

johndoh:
I included 'non-binary' because I thought it was an option I have seen other places. I looked at few articles like [stack exchang](https://softwareengineering.stackexchange.com/questions/381628/is-there-an-industry-standard-for-gender-model-other-than-male-and-female), [survey monkey](https://uk.surveymonkey.com/curiosity/ask-survey-questions-sexual-orientation-gender-identity/) or [this](https://www.csusm.edu/ipa/surveys/inclusive-language-guidelines.html).

You raise a great point about the vCard 4 standard. Roundcube currently uses vCard 3 and the custom `X-GENDER` field. One day Roundcube might update from 3 to 4 so does it make sense to add something which breaks that? I have no problem with only adding 'other' and not 'non-binary'. I agree there is no need for a 'none or not applicable' as the field is not required and can be left blank.

pabzm:
I think "non-binary" does exist as a self-description of some people in current societies (regardless of it being desired or legal), so I'd like to include it.

Currently with vcard v3 there's no technical reason not to include it, too, because (as @johndoh said) it's a custom field anyway.

RFC6350 for vcard v4 has the mentioned categories for the "sex component" part of the attribute, which has no direct representation of "non-binary", but additionally there's a free-form text part for expressing gender options. It literally includes "GENDER:;it's complicated" as an example, so I suppose "non-binary" could fit in there well.

Issue #9419, opened 2024-04-18
Chose from all available contact fields on CSV import


twisterbr:
Continuing #9393

@alecpl I applied the patch (commit ed47e11), there is no blank option anymore, however, I still can't set a mapping to email:work, this option is still missing from the select tag list from address import tool.

I want to have a mapping selection just as it shows when I edit a contact, like this:
![image](https://github.com/roundcube/roundcubemail/assets/4530260/d8af4d47-6828-41d3-94f1-aff6c5081b7a)

**import tool**
![image](https://github.com/roundcube/roundcubemail/assets/4530260/bd0285d7-95ee-49c6-9988-890b19961f01)

alecpl:
It displays field labels known to the csv2vcard converter. I agree that it would be better to list fields available in the Roundcube UI. However, it might be more work than you'd expect.

Maybe @johndoh would like to work on this.

alecpl:
Implemented

PR #9431, opened 2024-05-01
Contact import improvements

johndoh:
For #9419

Some of the mappings in the current code work (pass the tests) but the result is different to what you get if you create the same contact manually through the UI and then export it as a vCard, for example Roundcube uses `phone:mobile` not `phone:cell` and there is no `email:pref` field. I have changed the mapping so it matches the fields in the UI. I added some more aliases `rcube_vcard` to show existing vCards with mismatched fields nicely.

I have removed the subtype `other` from the IM field because that does not exist in `rcube_vcard` which means if when you hit save in the UI the value simply disappears, it’s not saved.

I have removed the variable `rcube_csv2vcard::$label_map` because it’s a duplicate of contents of `/program/localization/en_US/csv2vcard.inc`and just a bit confusing currently `/program/localization/en_US/csv2vcard.inc` is not used by the code and this PR changes that.

I have another reason for wanting to use `/program/localization/en_US/csv2vcard.inc` over `rcube_csv2vcard::$label_map` and if this PR gets merged then I will propose 1 additional change which will remove the `rcube_csv2vcard::$csv2vcard_map` variable and simplify the mapping process by storing both the source and destination for the map in `/program/localization/en_US/csv2vcard.inc` and allowing multiple sources (including multiple languages) for the same destination. This is a slightly larger and may be more controversial change hence the separate PR.


program/actions/contacts/import.php
```diff
@@ -369,16 +367,46 @@ public static function import_map($attrib)
...
+            $available_fields['groups'] = $rcmail->gettext('groups');
         }
 
+        // add separate birthday date parts fields for thunderbird imports
+        $available_fields['birthday-d'] = $rcmail->gettext('birth_day');
+        $available_fields['birthday-m'] = $rcmail->gettext('birth_month');
+        $available_fields['birthday-y'] = $rcmail->gettext('birth_year');
+
+        // sort by label for easy use
+        asort($available_fields);
```
alecpl, line 405: We need should use SORT_LOCALE_STRING flag.

program/actions/contacts/import.php
```diff
@@ -369,16 +367,46 @@ public static function import_map($attrib)
...
+
+        // allow importing of group assignments
+        if (!empty($params['with_groups'])) {
+            $available_fields['groups'] = $rcmail->gettext('groups');
         }
 
+        // add separate birthday date parts fields for thunderbird imports
+        $available_fields['birthday-d'] = $rcmail->gettext('birth_day');
+        $available_fields['birthday-m'] = $rcmail->gettext('birth_month');
+        $available_fields['birthday-y'] = $rcmail->gettext('birth_year');
```
alecpl, line 402: Maybe these extra fields definition should be stored in rcube_csv2vcard class.

program/lib/Roundcube/rcube_csv2vcard.php
```diff
@@ -165,141 +165,33 @@ class rcube_csv2vcard
...
+        'email_home' => 'email:home',
+        'email_work' => 'email:work',
+        'email_other' => 'email:other',
+        'phone_video' => 'phone:video',
+        'maidenname' => 'maidenname',
+        'im_aim' => 'im:aim',
+        'im_icq' => 'im:icq',
+        'im_jabber' => 'im:jabber',
+        'im_msn' => 'im:msn',
+        'im_other' => 'im:other',
```
alecpl, line 179: Didn't you remove that field?

program/lib/Roundcube/rcube_vcard.php
```diff
@@ -51,6 +51,10 @@ class rcube_vcard
         'IPHONE' => 'mobile',
         'CELL' => 'mobile',
         'WORK,FAX' => 'workfax',
+        'WORK,CELL' => 'other',
+        'HOOME,FAX' => 'homefax',
```
alecpl, line 55: A typo.

program/localization/en_US/labels.inc
```diff
@@ -505,6 +505,11 @@ $labels['importmapdesc'] = 'Confirm the field mapping information below is corre
 $labels['fieldnotmapped'] = 'Field not mapped (do not import)';
 $labels['skipheader'] = 'Do not import first line (headers)';
 $labels['done'] = 'Done';
+$labels['contactfieldsubtype'] = '$field $subtype';
+$labels['contactaddressfieldtype'] = '$subtype $field';
```
alecpl, line 509: I'm not sure these two need to be here at all.

alecpl (changes requested):

program/lib/Roundcube/rcube_vcard.php
johndoh: are you suggesting `work,cell` should map to `mobile`? there is no `workmobile` field type.

program/lib/Roundcube/rcube_csv2vcard.php
johndoh: I have not removed anything

program/localization/en_US/labels.inc
johndoh: In the contact UI the fields are displayed in different sections e.g. home, work, other but on the import screen where there is only a simple select box i needed way to differentiate the types of field. I don't know if its written as "phone work" in all languages so to be safe i included the format in the localization file. There are 2 because for most fields like written like "phone home" to group all the phone fields together in the list but for address fields its like "home city" so the parts of the address are together.

program/lib/Roundcube/rcube_csv2vcard.php
alecpl: I was referring to im:other that you removed in program/actions/contacts/index.php

program/lib/Roundcube/rcube_vcard.php
alecpl: A typo in "HOOME".

program/localization/en_US/labels.inc
alecpl: I think it should/could be "city - home" "phone - home" (or "city (home)", "phone (home)"). Using localization for this is wrong and will not work for many languages.

alecpl (changes requested):

program/actions/contacts/import.php
johndoh: created new rcube_csv2vcard::list_fields()

johndoh:
I moved the `list_fields` function back to `rcmail_action_contacts_import` because it relies on the contact list from its parent class `rcmail_action_contacts`. I think its the best place for it.

Also I completed the second part of my suggestion to remove the hard coded mapping information entirely and use 1 to define the source and destination.

program/actions/contacts/index.php
```diff
@@ -201,7 +201,7 @@ class rcmail_action_contacts_index extends rcmail_action
             'size' => 40,
             'maxlength' => 128,
             'label' => 'instantmessenger',
-            'subtypes' => ['aim', 'icq', 'msn', 'yahoo', 'jabber', 'skype', 'other'],
+            'subtypes' => ['aim', 'icq', 'msn', 'yahoo', 'jabber', 'skype'],
```
alecpl, line 204: Why 'other' is removed here?

program/localization/de_CH/csv2vcard.inc
```diff
@@ -16,84 +16,84 @@
...
-$map['url'] = "URL";
-$map['work_city'] = "Work City";
-$map['work_company'] = "Work Company";
-$map['work_dept'] = "Work Dept";
-$map['work_fax'] = "Work Fax";
-$map['work_mobile'] = "Work Mobile";
-$map['work_state'] = "Work State";
-$map['work_title'] = "Work Title";
-$map['work_zip'] = "Work Zip";
+$map['anniversary'][] = "Jahrestag";
```
alecpl, line 19: I don't like this change. We should keep this file syntax simple.

alecpl (changes requested):

program/actions/contacts/index.php
johndoh: Because it's not supported by the rest of the code. It's not included in the vCard, so once you hit save the value is lost.

program/localization/de_CH/csv2vcard.inc
johndoh: What is not simple about the syntax? This way, the mapping process is clearer. There is only one map of CSV field to vCard field rather than the CSV to internal ID, then a hard coded map of internl ID to vCard used currebtly. It also adds support variations of field names in the CSV file.

alecpl: `$map['anniversary'][]` is not simple ;)

johndoh: The reason for the nested array is to allow multiple labels to map to the same vCard field, e.g. currently the CSV labels `first` and `given_name` both map to vCard field `firstname`. Currently this is done via the hard coded `rcube_csv2vcard::$csv2vcard_map`. For me this is not simple. Adding a new field into the map means a change in 2 files, you have to know the hard coded map is there, there is no info about it in `csv2vcard.inc`. Using a nested array may not be simple but for me it is simpler than the current approach as it means there is only 1 place to change when adding a new field.

pabzm: I understand your approach but also understand Alec's concern and share the latter here. To me, `csv2vcard.inc` is a config file, and so the priority should be readability for humans. Understanding the syntax is one thing (which might be hard for non-PHP-developers), another thing I don't understand without intimate knowledge of the code is why e.g. `email:other` gets three values:

```
$map['email:other'][] = "E-mail Address";
$map['email:other'][] = "E-mail 2 Address";
$map['email:other'][] = "E-mail 3 Address";
```

Can we find a way to specify the mapping while keeping your approach (which I do like!) to de-duplicate and strip indirection?

johndoh:
@pabzm any chance this could be part of 1.7?

pabzm:
From my point of view, yes. I don't really like the syntax in the `csv2vcard.inc`-files, but I'm happy about the cleanup and those files don't change often, so it's not a big problem and shouldn't block us, I think.

@alecpl What do you think about this?

johndoh:
the format of `csv2vcard.inc` is now similar to `labels.inc`/`messages.inc`

pabzm:
Much nicer, thank you! Can we make the format even more close by swapping keys and values in the inc-files, and run `array_flip()` on them after reading?

johndoh:
I put them this way round because multiple CSV fields are mapped to the same vCard field so, I think, its either:
```php
$map['email:other'][] = "E-mail Address";
$map['email:other'][] = "E-mail 2 Address";
$map['email:other'][] = "E-mail 3 Address";
```
or
```php
$map['E-mail Address'] = 'email:other';
$map['E-mail 2 Address'] = 'email:other';
$map['E-mail 3 Address'] = 'email:other';
```

pabzm:
Oh, alright! Then I prefer the second and current way.

Thank you again!

pabzm (approved):

alecpl (approved):

PR #9440, opened 2024-05-06
include information about base skin when current skin is an extended one

johndoh:
when an extended skin is being the `skin_extends` array contains the name(s) of the skins the current skin is based on

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -31,6 +31,7 @@ class rcmail_output_html extends rcmail_output
     protected $js_labels = [];
     protected $js_commands = [];
     protected $skin_paths = [];
+    protected $skin_extends = [];
     protected $skin_name = '';
     protected $scripts_path = '';
     protected $script_files = [];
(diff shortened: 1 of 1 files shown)


miaulalala:
Can you add some context why this is useful / important?

johndoh:
example use case: you want to add something to the UI using a plugin, there are a couple of ways of doing already but this addition means in JS you can find out not only what the current skin is but if its extended so you can (probably) use the same code.

pabzm:
@johndoh I wouldn't oppose this change because it's small enough, but to be honest I still don't understand what it's actually useful for, which makes me hesitate. Could you maybe describe a concrete example?

johndoh:
A user of an extended skin I created had a question about support for a specific plugin.

It should have worked out of the box as the plugin already supports the base skin and if you use skins code on the server side this case is handled automatically.

The plugin dev was not doing the skin work on the server side but on the client side using JS.

I agree this is an edge case but I could not see a way for a plugin dev to handle an extended skin in JS like they can in PHP and my aim was to create this opportunity.

pabzm (approved):
Alright, it might help, doesn't hurt, and is small and clear. 👍

alecpl:
Looks good, but has a conflict.

PR #9441, opened 2024-05-06
add skin info to about dialog

johndoh:
mostly for the benefit of 3rd party skins this PR adds an `Interface Skin` section to the `About` dialog.

Untitled-1

The difference to the `Settings` section is that this includes version info and a download link (when available).

Also `meta.json` is read in a couple of different places so this PR unifies that into a single function.


pabzm (approved):
This looks good, thank you for the contribution!

PR #9449, opened 2024-05-14
Fix Code Spell check by disabling `String.raw` rule

pabzm:
The eslint package "eslint-plugin-unicorn" introduced a new rule in version 53.0.0 that suggests to use `String.raw` in order to avoid the need for escaping backslashes.

While this certainly is a good idea we cannot follow that rule as long as we support Internet Explorer 11+, which doesn't support template literals (and thus `String.raw`).


miaulalala (approved):
Makes sense!

PR #9455, opened 2024-05-19
Update .gitignore

gerbert:
Added JetBrains PhpStorm project folder (.idea)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ skins/elastic/styles/_variables.less
 # ignore IDE files
 .vscode
 .DS_Store
+.idea


gerbert:
One minor note I noticed in .gitignore, lines 25-27. Shouldn't the pass be relative?

```
# eslint dependencies                                                            
/node_modules
/package.json
/package-lock.json
```

pabzm:
> Shouldn't the path be relative?

The paths are all relative to the repo's root directory, see <https://git-scm.com/docs/gitignore#_pattern_format>:
> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

pabzm:
@alecpl I allowed myself to merge without further review because the change is so tiny and simpel.

gerbert:
> > Shouldn't the path be relative?
> 
> The paths are all relative to the repo's root directory, see https://git-scm.com/docs/gitignore#_pattern_format:
> 
> > If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

Well, thanks for pointing out, never knew, actually. Might be handy in the future.

PR #9467, opened 2024-05-25
Notification focus on click in Chrome

DannyDaemonic:
For Windows, it seems at some point Chrome stopped focusing on the window/tab that sent the notification when you click it. Firefox still behaves correctly, but to get this to work again for Chrome, I had to call `window.parent.focus()`.

Edit: I did test this change in other browsers. This fixes the same issue in Edge (because it's basically just Chrome now) and it doesn't break anything in Firefox.

### DIFF
--- a/plugins/newmail_notifier/newmail_notifier.js
+++ b/plugins/newmail_notifier/newmail_notifier.js
@@ -108,6 +108,9 @@ function newmail_notifier_desktop(body, disabled_callback) {
                 icon: icon,
             });
             popup.onclick = function () {
+                if (window.parent) {
+                    window.parent.focus();
+                }
                 this.close();
             };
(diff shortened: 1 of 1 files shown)


pabzm:
@DannyDaemonic Thank you for the contribution! Did you by chance test this change on other OS like Linux or macos?

I'm wondering how this behaves on those.

DannyDaemonic:
I tested it with Firefox on Linux and no issues there. I just installed Chrome and can confirm the same issue exists under the Linux builds of Chrome and the same `focus` fix resolves it.

Unfortunately, I don't currently have an easy way to test this on macOS.

pabzm (approved):
Thank you very much! That's good enough for me 👍

Issue #9465, opened 2024-05-24
Isolate inline message parts in individual, sandboxed iframes


pabzm:
I'd like to change how Roundcube renders messages: Let's fetch message- and mime-part-headers as JSON, render their details from Javascript, and insert an iframe for each message part that is to be shown, which then fetches its rendered contents individually.

This would:
* mitigate redressing attacks (like #6450, but also overlaying Roundcube-buttons, injecting new buttons into the action bar etc.),
* block a lot of XSS and similar attacks (due to strict sandboxing),
* prevent any Javascript execution at all (due to strict sandboxing),
* reduce inline Javascript (which brings us one small step closer to enable stricter CSPs).

I know this sounds like a big change, but I built a simple proof-of-concept that doesn't even muddle up too much existing code, so I think it's doable (if you're curious: https://github.com/pabzm/roundcubemail/tree/message-render-json-iframe). It's still probably much work due to a lot of edge cases with "interesting" messages, but I'd like to tackle it. 

Of course the edge case handling currently implemented in `show.php` would need to be moved, but that'd be good for testing anyway.

This is an itch that bugs me since quite a while, and I think it would be a huge step forward.

@roundcube/webmail-developers What do you think?

alecpl:
I think it's harder than you think, not mentioning it's a big BC break. Performance wise also might be not great.

pabzm:
I know it's ambitious, but it would solve or help so many issues. No scripting issues from HTML messages anymore! Way less room for redressing attacks! :)

I've come around to reduce the attempt to one-iframe-per-part, without loading-data-as-json, which is in branch pabzm/roundcube@message-render-iframe and contains even less changes.

>  Performance wise also might be not great.

Why do you think so? Because more HTTP requests are necessary? I'd still consider the security advantage a win over that.

pabzm:
@alecpl Which things would you expect to be the biggest problems? I made the iframed rendering work with HTML-emails, that contain remote objects, and with signed/encrypted emails, too. In both cases the necessary hints and buttons are displayed and are working, even though they are not rendered inside the iframe.

PR #9472, opened 2024-05-30
Filter "real" attachments by being referenced

### LINKED ISSUE #9443: Cannot download attachment for some multipart/mixed messages with base64-encoded jpg images
Hello,

We have an issue where pictures sent from a phone (presumably iPhone) to an e-mail address by Telenor DK will not be downloadable from Roundcube. The webmail will show the paperclip icon, indicating there is an attachment, but it won't be displayed when opening the message, nor will it allow the user to download it. See Example 1.
The same message opened in Thunderbird will show correctly and the attachment will be downloadable.

We have produced a similar example (Example 2) of a picture sent from an iPhone to an e-mail address using Telenor Norway, where Roundcube does show the attachment correctly.

Is this a bug in Roundcube?

1. **Example 1**: Not OK - Cannot download attachment
```text
Message-Id: <XXXXXXXXXXXXX@mx01.lytzenitmail.dk>
Date:Mon, 06 May 2024 15:24:47 +0200
From: <+45XXXXXXXX@telenor.dk>
To: XXXXXX@jubii.dk
Subject: Not OK
Mime-Version:1.0
Content-Type:multipart/mixed;boundary="--------------------------------------------=_NextPart_0_24856"

this is a multi-part message in MIME format.

----------------------------------------------=_NextPart_0_24856
Content-Type:image/jpeg; name="Resized_20240427_200026(1).jpeg"
Content-Transfer-Encoding:base64
Content-Location:Resized_20240427_200026(1).jpeg
Content-ID:<Resized_20240427_200026(1)>

iVBORw0KGgoAAAANSUhEUgAAAQAAAAEACAIAAADTED8xAAADMElEQVR4nOzVwQnAIBQFQYXff81R
UkQCOyDj1YOPnbXWPmeTRef+/3O/OyBjzh3CD95BfqICMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMO0TAAD//2An
hf4QtqobAAAAAElFTkSuQmCC

----------------------------------------------=_NextPart_0_24856--
```

2. **Example 2**: OK - Can download attachment
```text
Content-Type: multipart/mixed; boundary=1_663B4185_194FAEF8
Date: Wed, 08 May 2024 09:10:29 GMT
To: XXXXXXXXXX@corp.inbox.com
Subject: OK
From: 47XXXXXXXX@mms.mnc001.mcc242.gprs
Mime-Version: 1.0

--1_663B4185_194FAEF8
Content-Type: image/jpeg;Name=IMG_0639.jpg
Content-Disposition: Attachment;Filename=IMG_0639.jpg
Content-ID: 0
Content-Location: IMG_0639.jpg
Content-Transfer-Encoding: base64

iVBORw0KGgoAAAANSUhEUgAAAQAAAAEACAIAAADTED8xAAADMElEQVR4nOzVwQnAIBQFQYXff81R
UkQCOyDj1YOPnbXWPmeTRef+/3O/OyBjzh3CD95BfqICMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0C
MK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMK0CMO0TAAD//2An
hf4QtqobAAAAAElFTkSuQmCC
--1_663B4185_194FAEF8--
```

pabzm:
This changes the way in which attachments are determined to be shown as downloadable ("standalone"), or not ("inline").
In theory this should be determined by their `Content-Disposition`, but in reality this often doesn't work.

Previously, on top of the default standard handling, edge cases have been fixed by checking for their concrete details.

Now the code checks if the `Content-ID` or `Content-Location` of the attachment is actually being used in other parts of the message. If not, the attachment is considered to be "standalone". This might make some images be shown as downloadable, which have a "inline"-disposition – but it'll happen only in cases where the message is broken, or they wouldn't have been shown at all previously. So I consider this a progression while staying a comprehensible and review-able changeset. 

This fixes #9443. 

I didn't find an email that it breaks. I also don't see how it could.

It helps #9465, too.

And it (unintendedly) also helps #9326, because now the information, which image is actually referenced in an HTML-part, ist available already after message parsing.

### DIFF
--- a/program/lib/Roundcube/rcube_message.php
+++ b/program/lib/Roundcube/rcube_message.php
@@ -50,6 +50,14 @@ class rcube_message
     protected $got_html_part = false;
     protected $tnef_decode = false;
 
+    /**
+     * This holds a list of Content-IDs and Content-Locations by which parts of
+     * this message are referenced (e.g. in HTML parts).
+     *
+     * @var array
+     */
+    protected $replacement_references;
+
     public $uid;
     public $folder;
     public $headers;
@@ -564,6 +572,54 @@ class rcube_message
                     return true;
                 }
             }
+        }
+
+        return false;
+    }
+
+    /**
+     * Get a cached list of replacement references, which are collected during
+     * parsing from Content-Id and Content-Location headers of mime-parts.
+     */
+    protected function get_replacement_references(): array
+    {
+        if ($this->replacement_references === null) {
+            $this->replacement_references = [];
+            foreach ($this->mime_parts as $mime_part) {
+                foreach ($mime_part->replaces as $key => $value) {
+                    $this->replacement_references[] = preg_replace('/^cid:/', '', $key);
+                }
+            }
+        }
+        return $this->replacement_references;
+    }
+
+    /**
+     * Checks if a given message part is referred to from another message part.
+     * Usually this happens if an HTML-part includes images to show inline, but
+     * technically there can be other cases, too.
+     * In any case, an attachment that is *not* referred to, shall be shown to
+     * the users (either in/after the message body or as downloadable file).
+     *
+     * @param rcube_message_part $part Message part
+     *
+     * @return bool True if the part is an attachment part
+     */
+    public function is_referred_attachment(rcube_message_part $part): bool
+    {
+        $references = $this->get_replacement_references();
+
+        // This code is intentionally verbose to keep it comprehensible.
+        // Filter out attachments that are reference by their Content-ID in
+        // another mime-part.
+        if (!empty($part->content_id) && in_array($part->content_id, $references)) {
+            return true;
+        }
+
+        // Filter out attachments that are reference by their
+        // Content-Location in another mime-part.
+        if (!empty($part->content_location) && in_array($part->content_location, $references)) {
+            return true;
         }
 
         return false;
--- a/program/actions/mail/show.php
+++ b/program/actions/mail/show.php
@@ -621,7 +621,10 @@ class rcmail_action_mail_show extends rcmail_action_mail_index
      */
     public static function message_body($attrib)
     {
-        if (empty(self::$MESSAGE) || (empty(self::$MESSAGE->parts) && empty(self::$MESSAGE->body))) {
+        // Exit early if there's no content to be shown anyway.
+        // `mime_parts` also includes a message's body, even if it originally
+        // was the only part of the message.
+        if (empty(self::$MESSAGE) || empty(self::$MESSAGE->mime_parts)) {
             return '';
         }
 
--- a/program/actions/mail/show.php
+++ b/program/actions/mail/show.php
@@ -204,7 +204,7 @@ class rcmail_action_mail_show extends rcmail_action_mail_index
             }
 
             // Skip inline images
-            if (strpos($mimetype, 'image/') === 0 && !self::is_attachment(self::$MESSAGE, $attach_prop)) {
+            if (strpos($mimetype, 'image/') === 0 && self::$MESSAGE->is_referred_attachment($attach_prop)) {
                 continue;
             }
 
@@ -745,7 +745,7 @@ class rcmail_action_mail_show extends rcmail_action_mail_index
                 // Content-Type: image/*...
                 if ($mimetype = self::part_image_type($attach_prop)) {
                     // Skip inline images
-                    if (!self::is_attachment(self::$MESSAGE, $attach_prop)) {
+                    if (self::$MESSAGE->is_referred_attachment($attach_prop)) {
                         continue;
                     }
 
@@ -888,31 +888,5 @@ class rcmail_action_mail_show extends rcmail_action_mail_index
                 $rcmail->output->set_env('mdn_request', true);
             }
         }
-    }
-
-    /**
-     * Check whether the message part is a normal attachment
-     *
-     * @param rcube_message      $message Message object
-     * @param rcube_message_part $part    Message part
-     *
-     * @return bool
-     */
-    protected static function is_attachment($message, $part)
-    {
-        // Inline attachment with Content-Id specified
-        if (!empty($part->content_id) && $part->disposition == 'inline') {
-            return false;
-        }
-
-        // Any image attached to multipart/related message (#7184)
-        $parent_id = preg_replace('/\.[0-9]+$/', '', $part->mime_id);
-        $parent = $message->mime_parts[$parent_id] ?? null;
-
-        if ($parent && $parent->mimetype == 'multipart/related') {
-            return false;
-        }
-
-        return true;
     }
 }


program/lib/Roundcube/rcube_message.php
```diff
@@ -569,6 +577,44 @@ public function is_attachment($part)
...
+    }
+
+    /**
+     * Checks if part of the message is an attachment (or part of it)
+     *
+     * @param rcube_message_part $part Message part
+     *
+     * @return bool True if the part is an attachment part
+     */
+    public function is_standalone_attachment(rcube_message_part $part): bool
```
alecpl, line 600: Word "standalone" is not needed here. Also what "(or part of it)" means?

program/lib/Roundcube/rcube_message.php
```diff
@@ -569,6 +577,44 @@ public function is_attachment($part)
...
+    {
+        $references = $this->get_replacement_references();
+        // This code is intentionally verbose to keep it comprehensible.
```
alecpl, line 603: I prefer an empty line in here.

program/lib/Roundcube/rcube_message.php
```diff
@@ -569,6 +577,44 @@ public function is_attachment($part)
...
+        // Filter out attachments that are reference by their Content-ID in
+        // another mime-part.
+        if (!empty($part->content_id) && in_array($part->content_id, $references)) {
+            return false;
+        }
+        // Filter out attachments that are reference by their
```
alecpl, line 609: I prefer an empty line in here.

program/lib/Roundcube/rcube_message.php
```diff
@@ -569,6 +577,44 @@ public function is_attachment($part)
         return false;
     }
 
+    public function get_replacement_references(): array
```
alecpl, line 580: Does it needs to be public? Also, the name is not great, but I have no better idea now. Finally, this should probably take `context` property into account. I.e. testing this with forwarded messages might be good idea.

alecpl (changes requested):

program/lib/Roundcube/rcube_message.php
pabzm: I didn't understand that description anymore, either, so I fixed that. 🤔

I chose "standalone" to stress the meaning, because asking if an attachment "is_attachment()" feels confusing to me.

If we use this code to distinctly decide whether a mime-part is an attachment *(x)or* an inline-part (see #9473), then I'd agree that we don't use the extra word (and probably don't even need this method anymore). How about that?

program/lib/Roundcube/rcube_message.php
pabzm: Technically this could be `protected`, but it could be useful to solve #9326 as long as it's public.

program/lib/Roundcube/rcube_message.php
alecpl: I just noticed that we already have `is_attachments()` method. Anyway, maybe the method name should be `is_referred_attachment()`.

pabzm: I like that name, thank you for the suggestion! It allows to get rid of the negation in the calling code, too.

program/lib/Roundcube/rcube_message.php
pabzm: I realized that to solve #9326 one would want to call `is_referred_attachment()`, too. So I changed the method's visibility.

alecpl:
Looks to me that it fixes #9443 only partially. With `inline_images=true` I expect the image listed on attachments list to be displayed not only on the list, but also below the body (here a non-existent body).

I think for `inline_images` feature we'll really need to look into the body and display the image when: 1) there's no text parts at all, 2) there's no HTML part, 3) there's HTML part and it does not have a reference to the image part.

pabzm:
> Looks to me that it fixes #9443 only partially

That's because the message was considered empty and thus no body was rendered. An issue that is not exactly in scope of this PR, but because it is related to handling mime-parts I fixed it nonetheless in here.

Now the image is shown inline as expected.

I'd be happy to clean up the commit history before merging if you'd agree.

pabzm:
Wait a moment, I'll have to rebase to "master" anyway, and then I can clean the history a little, too.

pabzm:
@alecpl I rebased onto the latest master. Could you merge this or let me know what's blocking it, please?

pabzm:
@alecpl Just a note: I would prefer if you wouldn't use squash-commits when merging, because it makes rebasing much harder (git can't automatically identify the commits that were used before). If you don't like a list of commits please just ask to clean it up.

PR #9477, opened 2024-06-03
Rename test files for PHPUnit 10+

mvorisek:
Rename test files as PHPUnit 10 (or higher) require - each test class name must match its file name (/wo `.php` part of course).

Support for PHPUnit 10/11 to be landed in follow up PR as few additional changes are needed.


mvorisek:
E2E CI job is failing because it uses PHPUnit 11, probably because it uses some API that is no longer present or was renamed in PHPUnit 10/11, can someone advise?

mvorisek:
@alecpl can you please have a look - it is big refactoring PR and I would be happy if it can be merged asap to prevent merge conflicts.

alecpl:
I'd suggest to change the namespace from Roundcube\Mail\Tests to Roundcube\Tests, and it should be good to go.

mvorisek:
`Roundcube\Tests\...` can potentially collide with other Roundcube packages.

What about `Roundcube\Roundcubemail\Tests\...`?

alecpl:
I'd say the collision is unlikely, let's do Roundcube\Tests\..., please.

mvorisek:
Sure, you are the one who decide these :)

alecpl:
I also am not sure whether `Roundcube\Plugins\Tests` shouldn't become `Roundcube\Tests\Plugins`, but it probably does not matter.

mvorisek:
It should be more `Roundcube\XxxPlugin\Tests` as more ot less the plugins are separate composer packages...

PR #9479, opened 2024-06-04
Make dependabot check dependencies for NPM and composer

pabzm:
This makes dependabot check dependencies of composer in the root-directory, and NPM in both, the root-directory and in `skins/elastic`.

It will create a pull request for newer versions of dependencies, so we don't get behind on updates, or even miss important ones.

Roundcube's dependencies are not that many, but we can still make our life a little easier with this automation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ roundcubemail-git: buildtools
 	(cd roundcubemail-git/bin; rm -f transifexpull.sh package2composer.sh)
 	(cd roundcubemail-git; find . -name '.gitignore' | xargs rm -f)
 	(cd roundcubemail-git; find . -name '.travis.yml' | xargs rm -f)
-	(cd roundcubemail-git; rm -rf tests plugins/*/tests .git* .tx* .ci* .editorconfig* index-test.php Dockerfile Makefile)
+	(cd roundcubemail-git; rm -rf tests plugins/*/tests .git* .tx* .ci* .editorconfig* index-test.php Dockerfile Makefile package.json package-lock.json node_modules)
 	(cd roundcubemail-git; $(SEDI) 's/1.7-git/$(VERSION)/' index.php public_html/index.php installer/index.php program/include/iniset.php program/lib/Roundcube/bootstrap.php)
 	(cd roundcubemail-git; $(SEDI) 's/# Unreleased/# Release $(VERSION)'/ CHANGELOG.md)
 
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ roundcubemail-git: buildtools
 	(cd roundcubemail-git; $(SEDI) 's/# Unreleased/# Release $(VERSION)'/ CHANGELOG.md)
 
 buildtools: /tmp/composer.phar
-	npm install --include=dev
+	npm install --include=dev --omit=optional
 
 /tmp/composer.phar:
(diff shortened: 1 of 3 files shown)


alecpl:
We don't have npm dependencies per-se, so is this gonna do anything?

pabzm:
elastic depends on `lessc`, so I find it sensible to set that up for this monitoring. And then we can also supervise our dev-dependencies, I figured.

alecpl:
We don't have package.json so I don't know how dependapot would know what we depend on. But I have no idea about dependabot, so maybe it would just work.

alecpl:
I suppose we could add package.json in the root folder with all npm dependencies we use in the skin and bin scripts. They are mentioned in https://github.com/roundcube/roundcubemail/blob/d8817d39b9cc78915aaefa284c039a048e8b8215/Makefile#L74. Then I suppose this part of the makefile could be simplified, as well as some install instructions.

I have no idea whether we can/should change situation of dependencies from jsdeps.json.

pabzm:
So I added the dev-requirements and changed the Makefile accordingly.

The prod-dependencies in `jsdeps.json` are different, because we want the built packages, not the source package as we get them from NPM. We probably could also fetch the source packages and build a production package ourself, but we'd have to check with each package if their build steps are fully automatic, etc.

For now I added the jsdeps that are available via NPM (all but "publickey") as "optional" dependencies, which should make dependabot check them, but we can skip their installation with `--omit=optional` (as used in the Makefile).

I think this is the best compromise in our situation for now and would like to see what dependabot makes of it.

pabzm:
(I don't see how my changes could cause the failing Static Analysis check, if you have an idea please let me know.)

Makefile
```diff
@@ -61,9 +61,9 @@ roundcubemail-git: buildtools
...
 	(cd roundcubemail-git/skins/elastic; \
-		lessc --clean-css="--s1 --advanced" styles/styles.less > styles/styles.min.css; \
-		lessc --clean-css="--s1 --advanced" styles/print.less > styles/print.min.css; \
-		lessc --clean-css="--s1 --advanced" styles/embed.less > styles/embed.min.css)
+		../../node_modules/.bin/lessc --clean-css="--s1 --advanced" styles/styles.less > styles/styles.min.css; \
+		../../node_modules/.bin/lessc --clean-css="--s1 --advanced" styles/print.less > styles/print.min.css; \
+		../../node_modules/.bin/lessc --clean-css="--s1 --advanced" styles/embed.less > styles/embed.min.css)
 	(cd roundcubemail-git/bin; rm -f transifexpull.sh package2composer.sh)
 	(cd roundcubemail-git; find . -name '.gitignore' | xargs rm -f)
 	(cd roundcubemail-git; find . -name '.travis.yml' | xargs rm -f)
```
alecpl, line 69: Here we'd have to remove node_modules dir and the lock file. We don't want to make it part of the release.

alecpl:
The static analisis error is probably caused by the most recent phpstan version. Last commit in master was probably being checked with older version. So, it has nothing to do with this PR.

Makefile
pabzm: Good point! Do you want to keep `package.json`? E.g. `jsdeps.json` gets removed (in another target), too.

alecpl: Right, let's remove it too.

alecpl:
I think we should remove the `optionalDependencies`, but let it settle for a while.

PR #9480, opened 2024-06-05
Add PHPUnit 10.x and 11.x support

mvorisek:
follow up for #9477


alecpl:
Can we do something with these phpunit deprecation warnings when using phpunit >= 10?

mvorisek:
The phpunit deprecation warnings are gone, good catch!

mvorisek:
@alecpl this PR is done

I wonder also why such phpunit upgrade was not tried by Dependabot.

alecpl:
Maybe because of our minimum PHP version? I have no idea.

mvorisek:
@alecpl can it be merged?

PR #9484, opened 2024-06-07
Fix double scrollbar in compose window with big body

### LINKED ISSUE #7760: Double scroll bar when answering to a mail with many lines
I'm on 1.4.9 and elastic - but problem is still (or again?) the same.

_Originally posted by @tw- in https://github.com/roundcube/roundcubemail/issues/7500#issuecomment-736522029_

pabzm:
If editing a draft, that has a lot of text content already, previously sometimes a double scrollbar was shown, because the callback function that calculates the height for the textare was triggered too early.

This change sets the first height calculation behind a 100ms timeout. (A 0ms timeout might be feasible, too, just to put the callback execution on the call stack – it also works locally, but I'm not sure about real world behaviour, so better be safe than sorry.)

This closes #7760

### DIFF
--- a/skins/elastic/ui.js
+++ b/skins/elastic/ui.js
@@ -4052,7 +4052,8 @@ function rcube_elastic_ui() {
                 }
             };
 
-        $(textarea).on('input', resize).trigger('input');
+        $(textarea).on('input', resize);
+        setTimeout(resize, 100);
     }
 
     // Initializes smart list input


pabzm:
@alecpl  Should I add a line to the Changelog or do you prefer to do it yourself?

alecpl:
> @alecpl Should I add a line to the Changelog or do you prefer to do it yourself?

Adding the line in PR creates conflicts, so I prefer to add it after PR is merged.

Issue #8587, opened 2022-06-22
newmail_notifier desktop-notification


asaage:
It would be nice to see the subject and sender in the desktop notification (as requested in https://github.com/roundcube/roundcubemail/issues/8360).

Other improvements like an updated logo and sound select-options would also be nice (https://github.com/roundcube/roundcubemail/issues/7826).

alecpl:
So, why a new ticket?

asaage:
> So,` why a new ticket?

- one is labeled as bug and you didn't say anything about point 3.
- the second one: i just found that while writing
- so i can work on it if i have some time to spend

alecpl:
Added to the list in #7826.

Issue #8360, opened 2021-12-11
"newmail_notifier" plugin issues


sdondley:
I'm on mac, using Safari, running RC version 1.5.1. I'm noticing weirdness with the email notification plugin:

1) The "Test" for browser notifications does not work at all.

I have turned on notifications in Safari under Preferences -> Websites -> notifications. Pop-up windows are also unblocked.

2) I only get an audio notification when roundcube is open in my browser tab. It's silent if I'm in another tab. 

3) It would be nice to see the subject and sender in the desktop notification and perhaps the first line or two of the notification.

sdondley:
Regarding number 2 above, see https://stackoverflow.com/a/38482050/1641112

The solution there worked for me.

alecpl:
1. The notification test works for me in Firefox and Chrome on Linux. When I click "Test" I see a browser dialog asking for permission, if I accept the notification is disaplayed.
2. Works in Firefox. Does not work in Chrome (the sound is played as soon as I switch back to the Roundcube tab).

So, it looks like 1. might be a Safari only bug (or your setup bug), 2. seems to be a more common issue.

Issue #7826, opened 2021-01-12
Update favicon for newmail_notifier with new Roundcube logo


thomascube:
Subject says it all...

Bonus points for adding a new type to the `skin_logo` config option (e.g. favicon:newmail)

PR #9492, opened 2024-06-10
newmail_notifier: Include sender and subject in desktop notification

Daniel15:
Displays the sender and subject in desktop notifications:
image

References #8360 #7826 #8587


plugins/newmail_notifier/newmail_notifier.js
```diff
@@ -37,7 +37,19 @@ function newmail_notifier_run(prop) {
         newmail_notifier_sound();
     }
     if (prop.desktop) {
-        newmail_notifier_desktop(rcmail.get_label('body', 'newmail_notifier'));
+        var title = prop.from || rcmail.get_label('title', 'newmail_notifier');
```
Daniel15, line 40: The JavaScript in Roundcube seems to not be using newer features, so I'm avoiding them in my changes (`var` instead of `const`/`let`  here, and string concatenation instead of template literals for the strings on line 48 and 50)

plugins/newmail_notifier/newmail_notifier.js
```diff
@@ -37,7 +37,19 @@ function newmail_notifier_run(prop) {
...
+        var body = prop.subject || rcmail.get_label('body', 'newmail_notifier');
+
+        // If there's more than one unread email, show the number of other emails
+        var otherCount = prop.count - 1;
+        if (otherCount >= 1) {
+            title += prop.from
+                // Have sender name, e.g. "Daniel L +3"
+                ? ' +' +otherCount
+                // No sender name, e.g. "New Mail! (3)
+                : '('+otherCount + ')';
```
pabzm, line 50: Please add spaces around the plus-signs. Or maybe you could rework these title-related lines a little, including line 40, to be a little less terse but more readable?

plugins/newmail_notifier/newmail_notifier.js
pabzm: That's thoughtful, thanks!

pabzm (changes requested):
Thank you for the contribution, I like it!

From my side there's only the minor nitpick, the rest looks good!

plugins/newmail_notifier/newmail_notifier.php
```diff
@@ -203,12 +203,31 @@ public function notify($args)
...
             $this->notified = true;
 
+            $from = null;
+            $subject = null;
+            // Attempt to fetch headers for the latest unseen message
+            $latest_uid = $unseen->max();
+            if ($latest_uid !== null) {
+                $headers = $storage->fetch_headers($mbox, [$latest_uid]);
+                // fetch_headers returns an array, but we only care about the first one.
+                $headers = array_shift($headers);
```
alecpl, line 213: Not a big deal, but here you could use `array_first()`.

plugins/newmail_notifier/newmail_notifier.php
```diff
@@ -203,12 +203,31 @@ public function notify($args)
...
+                if ($headers !== null) {
+                    $from = $headers->from;
+                    $subject = $headers->subject;
```
alecpl, line 216: I wonder what sanity checks we should apply here. We probably should not use too long subjects. Also we should parse the From header content. We definitely should decode the content, not pass as-is.

See program/actions/mail/show.php for hints on what we normally do with subject and from header.

plugins/newmail_notifier/newmail_notifier.js
```diff
@@ -37,7 +37,19 @@ function newmail_notifier_run(prop) {
...
     if (prop.desktop) {
-        newmail_notifier_desktop(rcmail.get_label('body', 'newmail_notifier'));
+        var title = prop.from || rcmail.get_label('title', 'newmail_notifier');
+        var body = prop.subject || rcmail.get_label('body', 'newmail_notifier');
+
+        // If there's more than one unread email, show the number of other emails
+        var otherCount = prop.count - 1;
+        if (otherCount >= 1) {
+            title += prop.from
+                // Have sender name, e.g. "Daniel L +3"
```
alecpl, line 47: I don't like this whole thing. Neither "Daniel L +3" looks good nor "New Mail! (3)".

How about we show the sender/subject only when there's a single new mail message? And "(n) New Mail!" in other cases?

alecpl (changes requested):

Daniel15:
I haven't had free time to revise this yet, but I'll get back to it soon.

plugins/newmail_notifier/newmail_notifier.php
Daniel15: Thanks, I'll take a look. I wasn't sure whether to limit the subject length here or on the client side.

Daniel15: > We probably should not use too long subjects.

I tried with a long subject, and KDE makes them scrollable:

image

I haven't tested with Windows yet, but I think it truncates them.

> We definitely should decode the content, not pass as-is. See program/actions/mail/show.php for hints on what we normally do with subject and from header.

Thanks for the pointer. I thought the headers would already be decoded here.

Daniel15:
Updated to properly decode the from and subject headers. I updated it to just show the sender name rather than their email address too, to make the notification a bit more concise.

Tested with a multi-byte (Chinese) subject:
Screenshot_20240707_175454

Tested with a Chinese sender name:
Screenshot_20240707_175550

plugins/newmail_notifier/newmail_notifier.php
```diff
@@ -203,12 +203,36 @@ public function notify($args)
...
+                // fetch_headers returns an array, but we only care about the first one.
+                $headers = array_first($headers);
+                if ($headers !== null) {
+                    $subject = trim(
+                        rcube_mime::decode_header($headers->subject, $headers->charset)
+                    );
+                    $from_details = array_first(
+                        rcube_mime::decode_address_list($headers->from, 1, true, $headers->charset)
+                    );
+                    $from = $from_details['name'];
```
alecpl, line 221: Name can be empty. In such case let's fallback to the email address. Also, in some corner cases `From` header might not exist at all.

pabzm:
@Daniel15 Are you still interested in this? If so, please have a look at Alec's latest comments from July. If not, please close.

Neustradamus:
@Daniel15: Any progress on it?

It will be included in 1.7?

pabzm:
@Daniel15 I'm interested in bringing this forward and would offer to take over in about three weeks, if you can't find time for it.

PR #9494, opened 2024-06-11
Add a note about 'unsafe-inline' etc. required in CSP

### LINKED ISSUE #9468: Install Requirements: unsafe-inline and unsafe-eval in script-src if CSP is in use
Rouncube documentation in Github Wiki has a chapter dedicated to "Install Requirements".

An important information lacks to avoid many site administrators fail on deploying this software:
If **Content-Security-Policy** is in use at webserver, `script-src` parameter must include `'unsafe-inline' 'unsafe-eval'` so Rouncube works.
(as of version 1.6.5)

I hope in the future these specific configurations are not necessary so a Roundcube site can get a score of **A+** on web security observatories.

pabzm:
Closes #9468 

I can also edit https://github.com/roundcube/roundcubemail/wiki/Installation, but I was wondering why we have two sources of very similar but not identical information. How about merging both into our `INSTALL` file and link that from the wiki?

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -189,6 +189,13 @@ It is also recommended to change the document root to <install path>/public_html
 after installation if Roundcube runs at root of a dedicated virtual host. This
 will automatically keep sensitive files out of reach for http requests.
 
+CONTENT-SECURITY-POLICY
+-----------------------
+
+If you use a Content-Security-Policy, please note that Roundcube *requires* the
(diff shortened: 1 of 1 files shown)


miaulalala (approved):

alecpl:
The wiki is/was meant to be more extensive, but we in general not very good on documenting things.

pabzm:
@alecpl Then I'll take the liberty and update or remove (and replace with links) outdated information from the wiki (just FYI)

PR #9500, opened 2024-06-13
Mailvelope: Facilitate adding custom domains

### LINKED ISSUE #9498: Mailvelope: Facilitate adding custom domains
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have [searched](https://plugins.roundcube.net/) for plugins that provide already provide the feature

### Proposal

Mailvelope [allows to trigger adding the current domain as "Authorized Domain" including API-access](https://github.com/mailvelope/mailvelope/wiki/Guidelines-for-email-providers/4f9b42af90f97ebeb14032fe1c5ee6e2c08379b6#trigger-inclusion-in-the-authorized-domain-list-from-your-application).

Roundcube should add such a triggering button into its settings. There's already a button with a warning message if mailvelope is not present or accessible. That message could be extended to provide that button.

### Motivation and context

This is a big help for users and lowers the bar for using Mailvelope (properly, i.e. with API-access) a lot.

pabzm:
This facilitates using Mailvelope (properly, i.e. with API-access) a
lot, because people don't have to manually add their domain into
Mailvelope's options but just click a button in Roundcube's settings. This will make Mailvelope open a popup to confirm this, and after that it will Just Work™️.

---
Screenshot:

image

---

This includes a new way to allow server code to specify events on attributes without using
inline event listeners (which require a very lax CSP and should be avoided). I'm curious for comments on that, too.

Closes #9498

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -3982,6 +3982,19 @@ function rcube_webmail() {
         }
     };
 
+    /**
+     * Triggger Mailvelope to add the current domain to the list of authorized
+     * domains (with API access).
+     */
+    this.mailvelope_enable = function () {
+        // Remove warning and enabling button if mailvelope was enabled.
+        window.addEventListener('mailvelope', function (ev) {
+            $('#mailvelope-warning').hide();
+        });
+        // Trigger Mailvelope.
+        $('body').append('<iframe style="display: none;" src="https://api.mailvelope.com/authorize-domain/?api=true" />');
+    };
+
     // Load Mailvelope functionality (and initialize keyring if needed)
     this.mailvelope_load = function (action) {
         var keyring = this.env.mailvelope_main_keyring ? undefined : this.env.user_id,
--- a/program/localization/en_US/labels.inc
+++ b/program/localization/en_US/labels.inc
@@ -681,6 +681,7 @@ $labels['encryption'] = 'Encryption';
 $labels['mailvelopeoptions'] = 'Mailvelope options';
 $labels['mailvelopemainkeyring'] = 'Use Mailvelope main keyring';
 $labels['mailvelopenotfound'] = 'Mailvelope extension is not enabled/installed in your browser.';
+$labels['mailvelopeenable'] = 'Enable';
 
 $labels['sortby'] = 'Sort by';
 $labels['sortasc'] = 'Sort ascending';
--- a/program/actions/settings/index.php
+++ b/program/actions/settings/index.php
@@ -1478,6 +1478,7 @@ class rcmail_action_settings_index extends rcmail_action
 
                         $field_id = 'rcmfd_mailvelope_main_keyring';
                         $input = new html_checkbox(['name' => '_mailvelope_main_keyring', 'id' => $field_id, 'value' => 1]);
+                        $mailvelope_enable_button = new html_button(['type' => 'button', 'class' => 'btn btn-secondary', 'onclick' => rcmail_output::JS_OBJECT_NAME . '.mailvelope_enable()']);
 
                         $blocks['mailvelope']['options']['mailvelope_status'] = [
                             'content' => html::div(
@@ -1487,6 +1488,7 @@ class rcmail_action_settings_index extends rcmail_action
                                     rcube::Q($rcmail->gettext('mailvelopenotfound'))
                                 )
                                 . html::script([], "if (!parent.mailvelope) \$('#mailvelope-warning').show()")
+                                . $mailvelope_enable_button->show(rcube::Q($rcmail->gettext('mailvelopeenable')))
                             ),
                         ];
 


pabzm:
Force-pushed because the commits were badly cut

alecpl:
> @alecpl Is it correct to commit the change to `en_US/labels.inc`? I'm not sure about the accustomed workflow there.

Yes.

program/js/app.js
```diff
@@ -10637,6 +10647,36 @@ function rcube_webmail() {
...
+     * <a data-rc-event-listener='click:aMethodToCall,dblclick:anotherMethodToCall'>something</a>
+     * A click will cause calling `this[aMethodToCall]`, and a double click
+     * will cause calling this[anotherMethodToCall].
+     * All methods receive the event as only argument.
+     * If you need data inside your method, store it e.g. in other
+     * data-attributes (JSON-serialized, if necessary).
+     */
+    this.addEventListenersFromElements = function () {
+        $('[data-rc-event-listener]').each((_idx, elem) => {
+            elem.dataset.rcEventListener.split(',').forEach((string) => {
```
alecpl, line 10666: We use a lot of inline code, that is not necessarily events and they require arguments. So, I'm not sure this is the solution.

Plugins might want to use methods of other object than `rcmail`.

program/localization/en_US/labels.inc
```diff
@@ -681,6 +681,7 @@ $labels['encryption'] = 'Encryption';
 $labels['mailvelopeoptions'] = 'Mailvelope options';
 $labels['mailvelopemainkeyring'] = 'Use Mailvelope main keyring';
 $labels['mailvelopenotfound'] = 'Mailvelope extension is not enabled/installed in your browser.';
+$labels['mailvelopeauthorizecurrentdomain'] = 'Enable Mailvelope for the current domain.';
```
alecpl, line 684: Maybe 'Enable Mailvelope for this application.' would be more clear for an end-user. I don't like "current domain". Also, I would use shorter label e.g. 'mailvelopeenable'.

program/js/app.js
```diff
@@ -3982,6 +3984,14 @@ function rcube_webmail() {
         }
     };
 
+    /**
+     * Triggger Mailvelope to add the current domain to the list of authorized
+     * domains (with API access).
+     */
+    this.mailvelope_trigger_authorize_domain = function () {
```
alecpl, line 3991: `mailvelope_register` might be better shorter name.

program/js/app.js
pabzm: > We use a lot of inline code, that is not necessarily events

Extending this to support also direct invocation of code is pretty easy – if this is our way to go.

> they require arguments

Relevant data can be stored in other data-attributes and then read from there by the invoked method.

I just built it in a way that suits the current purpose and only a little more, because I believe in developing code along the lines of actual necessity. I'd be happy to extend it to more use cases if you agree it makes sense. 

Perspectively I'd very much like to try to get rid of inline code, in order to enable more secure CSPs, iframe-sandboxing, etc. There are other ways to do this, but all of them require to separate JS-code from HTML-Elements (AFAIK).

program/localization/en_US/labels.inc
pabzm: Technically the domain is authorized, not an application, so in case someone has another Webapp on that domain, Mailvelope will be active in that application, too.

But I see the point, some people might be irritated by the wording. I'll change it.

program/js/app.js
pabzm: I'd like to keep our wording consistent, so if we use "enable" in the labels key, I'd use that here, too. I'll change it to `mailvelope_enable`, in the hope you agree.

program/js/app.js
pabzm: > Plugins might want to use methods of other object than rcmail

Good point, that would have to be implemented differently. Should be doable, too, though. We just have to think of a way to express it in a string that is unambigously parseable without using `eval`.

pabzm:
I noticed that the mailvelope-popup is opened twice, have to investigate that

pabzm:
> I noticed that the mailvelope-popup is opened twice, have to investigate that

That problem appears to be related to Firefox, and is reproducible on other sites, so I suppose it's not our problem. (If https://github.com/mailvelope/mailvelope/issues/841 leads to any contrary info, I'll take it back to us.)

alecpl:
1. I would not introduce here new ways to handle scripts, especially when this new way is incomplete and we should think about some solution for all cases.
2. After some more thought I think the button label should be just "Enable".
3. After the mailvelope dialog is closed (and user Accepted not Cancelled) we should hide the button/whole warning box.
4. That duplicate dialog issue is unfortunate, maybe we should wait and find a solution before throwing this on users.

pabzm:
1. I think it's pretty generic, just not for plugins (yet). But OK, I create a standalone ticket for that feature: #9508 – but please let's find a way forward, we must start somewhere to get rid of all the inline JS.
2. OK 👍 
3. OK 👍 
4. It happens only to Firefox-users, and not on servers with a proper CSP. How many Mailvelope-Users that are freshly onboarding will that be relevant for? I'd guess that's a pretty low number. We also can't find a solution for it, it's the extensions (or Firefox's) fault and responsibility. I'd prefer to not have this blocked from their issue.

(I force-pushed to clean the git-history and to rebase, I hope you don't mind.)

pabzm:
Mailvelope has published a new version, which includes a fix for the issue with duplicated popups (https://github.com/mailvelope/mailvelope/issues/841), so that problem is gone.

Issue #9499, opened 2024-06-13
Plugin hook for message_delete and message_move


JE4GLE:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have [searched](https://plugins.roundcube.net/) for plugins that provide already provide the feature

### Proposal

A plugin hook, which allows to get notified on message_delete and message_mvoe

### Motivation and context

I want to write a plugin, which logs these actions for legal reasons.

alecpl:
I'm not sure an IMAP client is the proper tool for that, I think such things should be done on the IMAP server.

That being said, we have already 'message_saved' hook, so adding 'message_delete' and 'message_move' could be added too, I suppose. Provide a pull request, please.

alecpl:
Done.

PR #9501, opened 2024-06-14
Added PluginAPI hooks for message_move and message_delete

JE4GLE:
This pull request implements the requested PluginAPI Hooks from ticket #9499.

_The wiki needs to be updated, once the change is released._

### message_delete

Triggered when a message is deleted.

_Arguments:_
  * uids: _The messages uids_
  * mailbox: _The mailbox the message is deleted from_

_Return values:_
  * abort: set to true to skip the delete and fail

### message_move

Triggered when a message is moved from one folder to another.
The message is identified by the `uid`. The `mailbox` and `target` value may be equal, if it is moved to the same folder.

_Arguments:_
  * uids: _The messages uids_
  * source_folder: _The mailbox the message is moved from_
  * target_folder: _The mailbox the message is moved to_

_Return values:_
  * abort: set to true to skip the move and fail

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -2589,16 +2589,10 @@ class rcube_imap extends rcube_storage
             return false;
         }
 
-        $abortedUids = [];
-        $uidsArray = explode(',', $uids);
-        foreach ($uidsArray as $uid) {
-            $plugin = $this->plugins->exec_hook('message_move', ['mailbox' => $from_mbox, 'target' => $to_mbox, 'uid' => $uid]);
-            if (isset($plugin['abort']) && $plugin['abort']) {
-                $abortedUids[] = $uid;
-            }
+        $plugin = $this->plugins->exec_hook('message_move', ['mailbox' => $from_mbox, 'target' => $to_mbox, 'uids' => $uids]);
+        if (isset($plugin['abort']) && $plugin['abort']) {
+            return false;
         }
-        $uidsArray = array_diff($uidsArray, $abortedUids);
-        $uids = implode(',', $uidsArray);
 
         $config = rcube::get_instance()->config;
         $to_trash = $to_mbox == $config->get('trash_mbox');
@@ -2710,16 +2704,10 @@ class rcube_imap extends rcube_storage
             return false;
         }
 
-        $abortedUids = [];
-        $uidsArray = explode(',', $uids);
-        foreach ($uidsArray as $uid) {
-            $plugin = $this->plugins->exec_hook('message_delete', ['mailbox' => $folder, 'uid' => $uid]);
-            if (isset($plugin['abort']) && $plugin['abort']) {
-                $abortedUids[] = $uid;
-            }
+        $plugin = $this->plugins->exec_hook('message_delete', ['mailbox' => $folder, 'uids' => $uids]);
+        if (isset($plugin['abort']) && $plugin['abort']) {
+            return false;
         }
-        $uidsArray = array_diff($uidsArray, $abortedUids);
-        $uids = implode(',', $uidsArray);
 
         $deleted = $this->conn->flag($folder, $uids, 'DELETED');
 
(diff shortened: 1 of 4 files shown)


alecpl:
I think better place to put these will be `rcube_imap`. Also, for your use-case this might be enough, but maybe we should take the `abort` flag into consideration.

alecpl:
I guess it depends what you want to do with it. For example `message_saved` is in `rcube_imap`. It's obvious why it makes sense to keep these closer to the storage class. `rcube_imap::move_message()` might be called from various places, e.g. `folder_purge` action uses it too. So, you can handle different move message actions from a single place.

Putting these in actions would make sense if you wanted to interfere in any way with the UI, I suppose.

JE4GLE:
Thank you, I adjusted the code accordingly!

program/lib/Roundcube/rcube_imap.php
```diff
@@ -2589,6 +2589,15 @@ public function move_message($uids, $to_mbox, $from_mbox = '')
             return false;
         }
 
+        $abortedUids = [];
+        foreach ($uids as $uid) {
```
alecpl, line 2593: `$uids` is a string here.

program/lib/Roundcube/rcube_imap.php
```diff
@@ -2590,13 +2590,15 @@ public function move_message($uids, $to_mbox, $from_mbox = '')
         }
 
         $abortedUids = [];
-        foreach ($uids as $uid) {
+        $uidsArray = explode(',', $uids);
```
alecpl, line 2593: Nope. $uids can be "*" or "1:*" or "1-2". I suggest to skip the loop and pass $uids as is to the hook. No aborted uids. Just return (false or []) on abort.

program/lib/Roundcube/rcube_imap.php
```diff
@@ -2699,6 +2704,11 @@ public function delete_message($uids, $folder = null)
             return false;
         }
 
+        $plugin = $this->plugins->exec_hook('message_delete', ['mailbox' => $folder, 'uids' => $uids]);
```
alecpl, line 2707: I know for example `message_read` hook has 'mailbox' argument, but I prefer 'folder' that is used in other hooks.

program/lib/Roundcube/rcube_imap.php
```diff
@@ -2589,6 +2589,11 @@ public function move_message($uids, $to_mbox, $from_mbox = '')
             return false;
         }
 
+        $plugin = $this->plugins->exec_hook('message_move', ['mailbox' => $from_mbox, 'target' => $to_mbox, 'uids' => $uids]);
+        if (isset($plugin['abort']) && $plugin['abort']) {
```
alecpl, line 2593: No need to use isset(), 'abort' is always set.

program/lib/Roundcube/rcube_imap.php
```diff
@@ -2589,6 +2589,11 @@ public function move_message($uids, $to_mbox, $from_mbox = '')
             return false;
         }
 
+        $plugin = $this->plugins->exec_hook('message_move', ['mailbox' => $from_mbox, 'target' => $to_mbox, 'uids' => $uids]);
```
alecpl, line 2592: How about 'from' and 'to' or 'source_folder' and 'target_folder'?

alecpl (changes requested):

program/lib/Roundcube/rcube_imap.php
```diff
@@ -2589,6 +2589,11 @@ public function move_message($uids, $to_mbox, $from_mbox = '')
             return false;
         }
 
+        $plugin = $this->plugins->exec_hook('message_move', ['from_folder' => $from_mbox, 'target_folder' => $to_mbox, 'uids' => $uids]);
```
alecpl, line 2592: All right, we're almost there. If we use 'target_folder' then we should use 'source_folder' or just 'folder' and 'target'. Or even just 'from' and 'to' like in the variable names.

JE4GLE:
Thank you for all your helpful comments! I hope the PR can be merged quickly :)

Issue #9508, opened 2024-06-18
Build foundation to replace inline JS code


pabzm:
Roundcube should try to get rid of inline Javascript code to allow more secure Content-Security-Policies.

Currently a lot of inline code is used by server code to set event listeners and run other short code snippets on DOM elements.

We need to find a way to either allow the server to reference existing Javascript code that should be called.

Or re-think how server code and client code interact.

pabzm:
In order to renovate Roundcube's core code and get rid of inline Javascript (as opposed to radically replace it with a new framework or paradigm, which I would rule out as too much work currently), we must put all JS-code in JS-files.

There's three questions with this:
1. Where to put the JS-code?
2. How to trigger it being executed?
3. How to provide metadata (`set_env()`)?

For **1.**, I would suggest to put the extracted code into new files, so we don't clutter `app.js` even more. We might want use one JS-file per `task` or `action`, so it we don't cause too many HTTP-requests (even though that's less of an issue with growing support for HTTP/2), but let's not put it all into one long file either.

Regarding **2.** I see two options: Specify which event or method to call on which interaction in HTML attributes, or let the JS-code attach itself to the relevant elements, selected via ID-attributes. I would prefer the latter, because the first idea will either work only for a subset of cases, or will get very complex quickly.
Letting JS-code attach itself means to not have the JS-logic right inside the HTML markup, as it is now, but I think one can get used to that, too.

(I proposed a solution with HTML attributes in #9500 at first because I figured it might be more acceptable to you, because it keeps the trigger inside the markup. But I actually prefer a clearer separation of JS and HTML.)

As for the metadata (**3.**) I would suggest to transport those in HTML attributes and read them from there in dedicated code in `app.js`. Complex data can be JSON-stringified and parsed. (In another project that approach worked fine and didn't cause (notable) performance problems, so I wouldn't expect those here.)

All these approaches work for plugins, too. Of course they would need some work, too, but we can announce the goal early and give plugin authors a lot of time to adjust. And in the end we're not setting the CSP, we just enable to use a stricter one – if a provider wants a plugin that conflicts with that because it didn't adapt, it still their decision.

@alecpl What do you think?

I can also build a small example implementation for one page, if that helps.

(If you want other people's opinion on this, please tag them.)

alecpl:
For event handlers we'll need the approach you proposed in #9500, but improved to work with plugins, etc. I don't see a better solution for this part of the problem.

[...]

PR #9513, opened 2024-06-21
Remove inline Javascript, part I

### LINKED ISSUE #6281: Support CPS nonces in message frames so that unsafe-inline isn't required.
Trying for a much more narrowly scoped version of #6202.

The message preview and message content frames each have one big inline script at the top. Adding a `nonce` attribute to this, and a `Content-Security-Policy` header with said nonce should be fairly straightforward.

The tricky bit is that there are event handlers added in-line to various elements on the page, mainly via `<roundcube:button command="whatever" ... />`, though there's an `onmouseover` defined for the subject line in the default theme.

Perhaps the inline event handlers could be replaced with `data-*` attributes that an otherwise allowed script can use to make the required `addEventListener` calls?

pabzm:
As part of #9508 here's a change that removes `add_script()` and transports `js_commands` in JSON, which then gets interpreted as callbacks to call with given arguments, instead of `eval`ing strings. 

This does not tackle the many inline event handlers, in order to keep it manageable.


pabzm:
@alecpl I don't know yet why the tests fail, will have a look at that on monday. But if you have the time maybe you can already have a look if you (roughly) agree?

pabzm:
@alecpl Could you have a look at this, please? I'll go forward and work on extracting inline event handlers, but I'd like to know as early as possible about vetos from your side.

program/actions/settings/index.php
```diff
@@ -371,8 +371,7 @@ public static function user_prefs($current = null)
 
                     if ($current) {
                         $product_name = $rcmail->config->get('product_name', 'Roundcube Webmail');
-                        $rcmail->output->add_script(sprintf("%s.check_protocol_handler('%s', '#mailtoprotohandler');",
-                            rcmail_output::JS_OBJECT_NAME, rcube::JQ($product_name)), 'docready');
+                        $rcmail->output->command('check_protocol_handler', rcube::JQ($product_name), '#mailtoprotohandler');
```
alecpl, line 374: rcube::JQ() is redundant here.

program/include/rcmail_action.php
```diff
@@ -171,7 +171,7 @@ public static function quota_display($attrib)
         $quota = self::quota_content($attrib);
 
         $rcmail->output->add_gui_object('quotadisplay', $attrib['id']);
-        $rcmail->output->add_script('rcmail.set_quota(' . rcube_output::json_serialize($quota) . ');', 'docready');
+        $rcmail->output->command('set_quota', rcube_output::json_serialize($quota));
```
alecpl, line 174: json serialization is redundant here.

program/include/rcmail_output_html.php
```diff
@@ -2042,24 +1977,21 @@ protected function _write($output = '')
...
         $page_header .= $this->header . "\n";
 
         if (!empty($this->script_files['head_bottom'])) {
             $page_header .= array_reduce((array) $this->script_files['head_bottom'], $merge_script_files);
         }
 
-        if (!empty($this->script_files['foot'])) {
-            $page_footer .= array_reduce((array) $this->script_files['foot'], $merge_script_files);
-        }
+        $page_footer .= html::div(['id' => 'js-data', 'style' => 'display: none', 'hidden' => true], $this->get_js_commands());
```
alecpl, line 1986: We need to encode the div content here. E.g. to not contain `<` and `>`, etc. Then I think we should be able to do `.text()` on the client side.

alecpl, line 1986: I think there might be a need to have two separate blocks of js commands. Sometimes you might need to register a command before other core commands are registered, but to be executed after them. I'm not 100% sure, but I'm thinking of some plugins.

alecpl (changes requested):
Some preliminary comments. Overall not bad.

program/include/rcmail_output_html.php
pabzm: I would rather go with the current code, because it's safe enough (JSON-encoded), and encoding and then decoding everything adds a (maybe small but still) performance weight. Also, it works, why change it?

pabzm: That issue was solved in 486717f85 (load JS-"foot"-files after the JS-commands).

I played with a lot of features and didn't find any errors, so I'd say it works like it is.

(For future code I'd suggest to express dependencies through events, not loading order, because then we'd one day maybe be free to e.g. load scripts in parallel, too.)

program/js/googiespell_init.js
```diff
@@ -0,0 +1,17 @@
+rcube_webmail.prototype.googiespell_init = function (asseturl, baseurl, use_dict, lang_chck_spell, lang_rsm_edt, lang_close, lang_revert, lang_no_error_found, lang_learn_word, languages, currentLanguage) {
```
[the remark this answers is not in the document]
pabzm: I had to shuffle the code around a little to place it in `editor.js` (good suggestion!), please have a look at https://github.com/roundcube/roundcubemail/commit/3b669addb95bc9ee7fd26486131df4e132dab91c if that works for you.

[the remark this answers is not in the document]
pabzm: This has errors, please wait until I resolved them.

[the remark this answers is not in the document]
pabzm: Fixed.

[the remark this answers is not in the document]
pabzm: The changed related code is in https://github.com/roundcube/roundcubemail/commit/8d6104b2b055cd3a6f6937c5312e5e2eac175ec5?diff=unified&w=1 now

program/include/rcmail_output_html.php
pabzm: While doing the requested changes the data handling broke again, so I decided to use a data-attribute instead. That works for all cases.

pabzm:
I rebased onto the latest `master` to get rid of the unrelated phpstan errors, and then force-pushed. Sorry if that muddles up your review process.

program/actions/mail/compose.php
```diff
@@ -513,30 +513,17 @@ public static function spellchecker_init()
...
-            ), 'foot');
+            $rcmail->output->set_env('googiespell_asset_url', $rcmail->output->asset_url($rcmail->output->get_skin_path()));
+            $rcmail->output->set_env('googiespell_base_url', $rcmail->url(['_task' => 'utils', '_action' => 'spell', '_remote' => 1]));
+            $rcmail->output->set_env('googiespell_use_dict', !empty($dictionary) ? 'true' : 'false');
+            $rcmail->output->set_env('googiespell_lang_chck_spell', rcube::JQ(rcube::Q($rcmail->gettext('checkspelling'))));
+            $rcmail->output->set_env('googiespell_lang_rsm_edt', rcube::JQ(rcube::Q($rcmail->gettext('resumeediting'))));
+            $rcmail->output->set_env('googiespell_lang_close', rcube::JQ(rcube::Q($rcmail->gettext('close'))));
+            $rcmail->output->set_env('googiespell_lang_revert', rcube::JQ(rcube::Q($rcmail->gettext('revertto'))));
+            $rcmail->output->set_env('googiespell_lang_no_error_found', rcube::JQ(rcube::Q($rcmail->gettext('nospellerrors'))));
+            $rcmail->output->set_env('googiespell_lang_learn_word', rcube::JQ(rcube::Q($rcmail->gettext('addtodict'))));
```
alecpl, line 524: I believe `rcube::JQ()` is redundant for values passed to `set_env()`. Also, while we're breaking BC anyway, we could make all these vars a single value.

program/js/editor.js
```diff
@@ -77,145 +77,152 @@ function rcube_text_editor(config, id) {
...
-        tinymce.registered_request_token = true;
-        tinymce.util.XHR.on('beforeSend', function (e) {
-            // secure spellchecker requests with Roundcube token
-            e.xhr.setRequestHeader('X-Roundcube-Request', rcmail.env.request_token);
-            // A hacky way of setting spellchecker language (there's no API for this in Tiny)
-            if (e.settings && e.settings.data) {
-                e.settings.data = e.settings.data.replace(/^(method=[a-zA-Z]+&lang=)([^&]+)/, '$1' + rcmail.env.spell_lang);
+    this.init = function () {
+        this.googiespell_init();
+        if (window.googie) {
```
alecpl, line 82: I suppose we should get rid of the global state (window.googie) and keep the reference inside the editor where it belongs.

program/include/rcmail_output_html.php
```diff
@@ -1929,21 +1887,6 @@ public function include_script($file, $position = 'head', $add_path = true)
         }
     }
 
-    /**
-     * Add inline javascript code
-     *
-     * @param string $script   JS code snippet
-     * @param string $position Target position [head|head_top|foot|docready]
-     */
-    public function add_script($script, $position = 'head')
```
alecpl, line 1938: Instead of removing the API we should consider using the `nonce` for these. It would make the plugin developers life easier. However, there will be problem with execution order, I suppose, so I'm not sure.

pabzm: I would prefer to not offer that option, because it's cleaner without it, and I guess most plugin authors wouldn't bother to clean up their code if it still works, which factually will degrade the CSP-related security of most Roundcube installations.
As you see in this branch, converting code to use external script files isn't really hard. It's just some chores to modernise ones code.

pabzm: Also the potential problems with code order you mentioned might well result in a steady stream of issues, which we would need to take care of – that's a work load I would very much like to avoid, especially since Roundcube only loses security advancements from it.

alecpl:
Another case to figure out is `rcmail_html_page` class. It's used to display attachment contents, i.e. without the `rcmail` client context, but contains onclick handler. I think it might be a case for `nonce` based solution. On the other hand it's an iframe some maybe it could be handled by the parent in onload.

pabzm:
This isn't meant to completely remove all inline-JS at once, that would be too big a change, I figured. It's only the first part, as the title says.

I already made some progress on further changes to take care of event-handlers (which led to #9544), but I'd like this to be merged before I open another PR to avoid overload.

alecpl:
Well, this is one big BC break. I'd rather see it completed before we make the decision to merge. There will be no return.

pabzm:
Alright, I'll push forward.

alecpl:
I don't see a reason to change `command()` method name. Also, you included changes from other PR here.

program/actions/mail/compose.php
```diff
@@ -1503,7 +1503,7 @@ public static function editor_selector($attrib)
             $attrib['name'] = 'editorSelect';
         }
 
-        $attrib['onchange'] = "return rcmail.command('toggle-editor', {id: '" . $attrib['editorid'] . "', html: this.value == 'html'}, '', event)";
+        $attrib['data-onchange'] = json_encode(['toggle_html_editor_by_value', '__EVENT__']);
```
alecpl, line 1506: There's a difference between `rcmail.toggle_editor` and `rcmail.command('toggle-editor')`. I'm not sure we can freely get rid of that distinction everywhere. Maybe introduce 'command:<command-name>' syntax.

program/actions/mail/index.php
```diff
@@ -1563,7 +1563,7 @@ public static function search_filter($attrib)
         }
 
         if (!self::get_bool_attr($attrib, 'noevent')) {
-            $attrib['onchange'] = rcmail_output::JS_OBJECT_NAME . '.filter_mailbox(this.value)';
+            $attrib['data-onchange'] = json_encode(['filter_mailbox_with_this_value', '__EVENT__']);
```
alecpl, line 1566: Maybe we should implement some magic in `html` class so we don't have to use json_encode() everywhere, just pass an array.

alecpl: I would even suggest that we stop using `data-` prefix here (in the code defining the event) and in `html` class we just check if an event attribute value is an array. It would be the most clean solution.

pabzm: I like the first idea 👍 

I wouldn't drop the prefix, though:
1. I'm a fan of expectable behaviour, and changing the handling of known attributes is not really expectable.
2. In templates we have to write the prefix anyways, because we can't specify arrays as arguments and thus must write JSON-encoded data there.
3. If we drop the prefix in the source-files, we're definitely breaking foreign code that might still want to define inline event handlers. I'm still playing with the idea to use a config option for strict CSP, which would throw errors if someone calls `output->add_script`, but which also could be turned off to preserver compatibility with old code.

program/actions/mail/compose.php
pabzm: You're right, I might have missed this distinction in a couple of places, I'll go over them again when cleaning up in August.

pabzm:
Many tests fail because they check for the former Javascript-code being present. I'll fix those after my absence.

But besides that I think I might have tackled most of the problems successfully! The UI works mostly as expected, and the browser console doesn't  show any errors despite the strict CSP. As far as I can tell the only feature not working is the page showing attachments, which you mentioned earlier. So the rest is hopefully only chores.

@alecpl Please have a look at this at a convenient occasion and let me know if you see fundamental problems or big bugs.

program/actions/mail/compose.php
alecpl: Also there's a difference between "function()" and "return function()".

alecpl:
1. I don't think we need both `command()` and `add_js_call()` methods. I think we can accept a BC break here, which probably is not that big. And this makes the patch needlessly bigger.
2. I think we should add magic to template objects and html classes so we don't have to add `data-` everywhere. If the attribute value is an array add `date-`. This way the "allowed attributes" feature will not change.
3. I want to keep the patch as small as possible as this will be a good reference for plugin authors to see what changes to their code will be needed. We probably will have to document this in wiki anyway, but still.
4. There's a difference between `onclick = "return rcmail.some_method()"` and `onclick = "rcmail.some_method()"` and `onclick = "rcmail.some_method(); return false"`. I see you covered the last case, but not the second one. Also, I think that assumption that the last argument of a command, if an object, is always an options object is wrong.
Maybe a better approach would be to prefix command name with `=` for the first case and `!` for the last one. Or something like that.
5. I see you use ES6 syntax in your code. Maybe it's time, but we might need to update `bin/jsshrink.sh` for that. I'm guessing here.
6. There's one more `eval()` call in `app.js`.
7. Question: Is using `elem.addEventListener('click', ...)` instead of `elem.onclick = ...` in javascript code needed for strict CSP? What's the reason for these changes?
8. Question: Do we need `makeNodesFromInput()` if we have jQuery?

I may have some more in another review, I haven't finished with it.

pabzm:
At the end of last week I came to the conclusion that the current approach to attach event handlers gets too complicated after adapting it for all the cases (your point No 4). I started to change the code to use a more direct way, which I didn't present to you yet, because it's still a little messy. I'll push some of it in the next days, please don't spend more time on this until then.

pabzm:
@alecpl Sorry I didn't tell you earlier, I got sick and forgot!

pabzm:
## 1. + 2. Backwards compatibility
I kept `command()` for backwards compatibility of older plugins. I would also suggest to not remove the `on*`-attributes from the list of allowed attributes, and to make the strict CSP configurable by a config option. Then people could still use older, "unsafe" plugins if they really need to, without patches. And we would ship a "soft" breaking change, that can be switched off.

## 3. + 4. Event handlers
I'm also aiming for comprehensible changes and code style! Also I found the same problem as you did, that sometimes the last argument is already an object in some code parts, and thus can't be taken for config options. Thinking about adding even more "magic" (e.g. using prefixes, like you suggested) made me realize that this is going to be hard to comprehend for plugin authors, future contributes or even us in a few years. Thus I decided to roll back some of the "magic" and use a more direct approach: 

My suggestion for the event-handlers is to use an attribute-value for identification and optionally more for arguments, and then use jQuery to attach an event handler. That changes the code flow a little, but it's way less "magical" than the previous solution and much easier to comprehend. Meanwhile the actual working code stays pretty unchanged thanks to jQuery interpreting the return values of the handler callbacks identically to the way inline JS return values are interpreted.

An example diff:

```diff
--- a/plugins/enigma/enigma.js
+++ b/plugins/enigma/enigma.js
@@ -567,14 +567,15 @@ rcube_webmail.prototype.enigma_compose_handler = function (props) {
 };
 
 // Import attached keys/certs file
-rcube_webmail.prototype.enigma_import_attachment = function (mime_id) {
+$('[data-event-handle="enigma_import_attachment"]').on('click', function (event) {
+    var mime_id = event.target.dataset.part;
     var lock = this.set_busy(true, 'loading'),
         post = { _uid: this.env.uid, _mbox: this.env.mailbox, _part: mime_id };
 
     this.http_post('plugin.enigmaimport', post, lock);
 
     return false;
-};
+});
 

--- a/plugins/enigma/lib/enigma_ui.php
+++ b/plugins/enigma/lib/enigma_ui.php
@@ -1107,7 +1107,8 @@ class enigma_ui
             $p['content'] = html::p(['class' => 'enigmaattachment boxinformation aligned-buttons'],
                 html::span(null, rcube::Q($this->enigma->gettext('keyattfound'))) .
                 html::tag('button', [
-                        'data-onclick' => ['enigma_import_attachment', $part],
+                        'data-event-handle' => 'enigma_import_attachment',
+                        'data-part' => $part,
                         'title' => $this->enigma->gettext('keyattimport'),
                         'class' => 'import btn-sm',
                     ], rcube::Q($this->rc->gettext('import'))
```

I've been using a `data`-attribute for identification to not get in the way of `id`s or `class`es that might change or be overwritten in other code parts, but technically it could be any attribute.

This is obviously not finished in all code parts, but please let me know what you think about this approach.

(I'm not sure yet if I'd keep the code around `add_js_call()` as it is, since compared to this it looks quite "magical", too.)



## 5. ES6

I'm using only a few parts of ES6, which are compatible with IE11 (according to <https://caniuse.com/es6>). Those parts are surely supported by uglifyjs: "uglify-js supports JavaScript and most language features in ECMAScript." <https://github.com/mishoo/UglifyJS#note>.


## 6. There's one more eval() call in app.js.

Yes, I'm not finished. But thank you for the hint!


## 7. onclick

`elem.onclick` requires `unsafe-inline` in a CSP, thus I'm replacing all those assignments.


## 8. makeNodesFromInput()

jQuery uses assignments to `innerHTML` in its `html()` method, which is another open door to XSS attacks, thus I want to avoid it. So: technically for a strict CSP it's not relevant, but for better XSS protection it is.

alecpl:
> ## 1. + 2. Backwards compatibility
> 
> I kept `command()` for backwards compatibility of older plugins. I would also suggest to not remove the `on*`-attributes from the list of allowed attributes, and to make the strict CSP configurable by a config option. Then people could still use older, "unsafe" plugins if they really need to, without patches. And we would ship a "soft" breaking change, that can be switched off.

The difference between add_js_call() and command() is negligible. Wait... command() now just calls add_js_call(). This makes no sense. You talk about BC, but at the same time you removed add_script(). I'm not against BC breaks, but let's not make the change bigger than required.

> ## 3. + 4. Event handlers
>
> My suggestion for the event-handlers is to use an attribute-value for identification and optionally more for arguments, and then use jQuery to attach an event handler. That changes the code flow a little, but it's way less "magical" than the previous solution and much easier to comprehend. Meanwhile the actual working code stays pretty unchanged thanks to jQuery interpreting the return values of the handler callbacks identically to the way inline JS return values are interpreted.
> 
> An example diff:
> 
> ```diff
> --- a/plugins/enigma/enigma.js
> +++ b/plugins/enigma/enigma.js
> @@ -567,14 +567,15 @@ rcube_webmail.prototype.enigma_compose_handler = function (props) {
>  };
>  
>  // Import attached keys/certs file
> -rcube_webmail.prototype.enigma_import_attachment = function (mime_id) {
> +$('[data-event-handle="enigma_import_attachment"]').on('click', function (event) {
> +    var mime_id = event.target.dataset.part;
>      var lock = this.set_busy(true, 'loading'),
>          post = { _uid: this.env.uid, _mbox: this.env.mailbox, _part: mime_id };
>  
>      this.http_post('plugin.enigmaimport', post, lock);
>  
>      return false;
> -};
> +});
>```

I don't like it at all. If you replace method with an event you won't be able to call it from other places. So, I guess the steps will be to keep the method and add an event handler, but then that event handler is execution-order sensitive and may require other changes. Also, passing arguments via data- attributes? Another extra code in the handler.

>```diff
> --- a/plugins/enigma/lib/enigma_ui.php
> +++ b/plugins/enigma/lib/enigma_ui.php
> @@ -1107,7 +1107,8 @@ class enigma_ui
>              $p['content'] = html::p(['class' => 'enigmaattachment boxinformation aligned-buttons'],
>                  html::span(null, rcube::Q($this->enigma->gettext('keyattfound'))) .
>                  html::tag('button', [
> -                        'data-onclick' => ['enigma_import_attachment', $part],
> +                        'data-event-handle' => 'enigma_import_attachment',
> +                        'data-part' => $part,
>                          'title' => $this->enigma->gettext('keyattimport'),
>                          'class' => 'import btn-sm',
>                      ], rcube::Q($this->rc->gettext('import'))
> ```

How about something like this:
```
'onclick' => ['enigma_import_attachment', $part],
// Mode optional, also EVENT_MODE_STOP or in future EVENT_MODE_ONCE
'onclick-mode' => rcube_output::EVENT_MODE_RETURN,
```
Then we send the mode as the last (or first) argument (int), that is added always no matter it was defined by the caller or not. Or we can use "<mode>:<command>" syntax and not mess with arguments. What we do here is secondary, as it's internal, though it might be tricky in `html` classes. But would be easy and clean on the calling code side.

> ## 7. onclick
> 
> `elem.onclick` requires `unsafe-inline` in a CSP, thus I'm replacing all those assignments.

I suppose I'd prefer the jQuery equivalent code to addEventListener(), for consistency. Also, maybe it should be a separate PR.

> ## 8. makeNodesFromInput()
> 
> jQuery uses assignments to `innerHTML` in its `html()` method, which is another open door to XSS attacks, thus I want to avoid it. So: technically for a strict CSP it's not relevant, but for better XSS protection it is.

So, a separate PR please.

program/include/rcmail_output_html.php
```diff
@@ -728,6 +718,7 @@ public function page_headers()
                 $this->header('X-Frame-Options: sameorigin', true);
             }
         }
+        $this->header("Content-Security-Policy: script-src 'self'", true);
```
alecpl, line 721: It should be `false` not `true`. Multiple CSP headers are allowed.

program/include/rcmail_html_page.php
```diff
@@ -74,8 +74,7 @@ public function register_inline_warning($text, $button_label = null, $button_url
         $text = html::span(null, $text);
 
         if ($button_label) {
-            $onclick = "location.href = '{$button_url}'";
-            $button = html::tag('button', ['onclick' => $onclick], rcube::Q($button_label));
+            $button = html::tag('button', ['data-event-handle' => 'call_redirect', 'data-url' => $button_url], rcube::Q($button_label));
```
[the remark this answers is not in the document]
pabzm: Your suggestion looks like the calling code adds an `onclick`-attribute, because in all other cases attributes are just inserted as given – but the result will be something very different. That breaks comprehensibility, which I consider a code smell and a potential problem for future contributors and maintainers.

If you insist on hiding the actual logic then let's use attributes that don't exist in HTML. E.g. `click-handler` or something similar, that obviously is custom.

pabzm:
> This makes no sense. You talk about BC, but at the same time you removed add_script()

As I said: I would suggest to try and make this not a breaking change, a pivot from the initial approach. I didn't find the time to catch up with the implementation due to other duties. Also you're blocking some of my ideas, so I'm talking about them before fully implementing them.


> If you replace method with an event you won't be able to call it from other places.

All methods I converted into event handlers were not called from anywhere else. This is true for many event-handling methods, if not all of them.

At the same time my approach is just obvious code to every reasonably experienced web developer out there, while using internal, opaque, non-obvious implementations makes it way harder for contributors to understand what's going on. I consider that a problem.


> but then that event handler is execution-order sensitive

How is that different from calling code from an `onclick`-handler?


> Also, passing arguments via data- attributes? Another extra code in the handler.

How is that different from your suggestion to use a `onclick-mode` attribute?

---

I understand the preference for reduced changesets, but that shouldn't come at the cost of harder comprehensibility. Keeping the calling code untouched but changing its effect isn't clean but convoluted in my eyes.

Please, let's try to find a comprehesible, modern solution that eases the mental load for us (and future) developers, instead of increasing it.

program/include/rcmail_html_page.php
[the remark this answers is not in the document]
alecpl: I insist on hiding the logic. We tell developers "no raw javascript", so they should not expect events to be special.  We can even log a warning/error when it's a string and not an array (or a specific object if we really want).

If you insist on a "custom" name then `<event>-handler` maybe could work, but I really don't like that.

Then there are templates... the json-array value looks more suited for templates, keeping its compactness in comparison to multi-attributes solution.

ps. the `html` class(es) (and `rcube_output`) probably should be moved into `program/include`. Output handling should never have been part of the Framework. Not that it changes anything, but for the record.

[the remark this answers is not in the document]
pabzm: > We tell developers "no raw javascript"

Where do we tell them that? Actually I think using more "vanilla" javascript that is (mostly) independent of app.js would be an improvement.

Or do you mean inline Javascript? But how would using explicit, obvious code be "special" in that meaning?

> Then there are templates

Good point. So if we hide the logic in the `html` class, there would be a very different way to specify the same JS-call from PHP-code to a template. I find that confusing and hard to understand, also for people who look at the code and its changes to find out e.g. how to change their own plugin code to align with a strict CSP.

> I really don't like that

I hear you, but I'd like to talk about technical advantages and disadvantages of different approaches and our goals with these. Personal preference is of course important, too, but could you let me know *why* you don't like the changes and ideas you reject?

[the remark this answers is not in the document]
alecpl: > Or do you mean inline Javascript? But how would using explicit, obvious code be "special" in that meaning?

Yes, we WILL tell them "no inline javascript because of CSP". I meant that if you know inline code is "forbidden" then you expect handler to do some magic, i.e. maybe the input is not inline/raw code string. Besides on PHP side we have `output->command()` with an input that is not an inline code.

> > Then there are templates
> 
> Good point. So if we hide the logic in the `html` class, there would be a very different way to specify the same JS-call from PHP-code to a template. 

To make it even more problematic we have event handlers on template objects e.g. `roundcube:button` and on normal HTML elements. We do not have control over the latter (unless you want to rewrite template parser). I'm not sure we have to have a unified API to handle both.

> > I really don't like that
> 
> I hear you, but I'd like to talk about technical advantages and disadvantages of different approaches and our goals with these. Personal preference is of course important, too, but could you let me know _why_ you don't like the changes and ideas you reject?

Because developers know "onclick" not "some custom name". And I really don't see a difference between learning "onclick" input API over "some custom name" input API.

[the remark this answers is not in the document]
pabzm: > I meant that if you know inline code is "forbidden" then you expect handler to do some magic

The whole point of my latest suggestion (using some identifying attribute and hook an jQuery-event-handler onto that) is to *avoid* magic, because I think it makes things unnecessarily complicated.

>  I'm not sure we have to have a unified API to handle both.

Using an identifying attribute and a event-handler function does work for both.

> Because developers know "onclick" not "some custom name"

Yes, but they know it to work in a certain way – but in Roundcube it would work differently. That's a potential for confusion I'd like to avoid.

---

I know this is not the way most Javascript code has worked in the past in Roundcube – but you said yourself it might be due to modernise the Javascript code. You referred to ES6-features there, but there's more to gain from modern Javascript and CSS(-selectors).

Another technical argument for the latest approach is performance: we'd completely drop JSON-encoding data, looping over all possible events, and decoding JSON-data. That's no doubt a win.

pabzm:
(Rebased to latest of "master")

pabzm:
@alecpl I would really appreciate if you would take the time to reply again on this topic. Just getting the ball dropped is a pretty frustrating collaboration experience.

PR #9518, opened 2024-06-25
Fix stable branch copyright

adrienbeau:
The copyright was only updated on the master branch.


alecpl:
We don't change copyright in stable branches.

adrienbeau:
All right. I'm a bit surprised, considering there have been updates to the stable code in 2023 and 2024.

PR #9519, opened 2024-06-25
Render each mime-part into an individual iframe

pabzm:
Implementing #9465 

TODO:
- [x] reading emails using mailvelope 
- [ ] check/change larry skin
- [x] Don't "wash" HTML content twice
- [x] Check reading emails with enigma
- [x] Fix hide_blockquote plugin


pabzm:
@alecpl I didn't find any problems, but still want to check some more plugins. Also one test is missing. But I think the general approach won't change anymore, so a review would already make sense.

pabzm:
@alecpl Do you think we should keep "larry" working for future releases?

I'm not sure yet how much work it would be to make larry fit for this change, but before I even try I want to be sure that you consider it necessary. (I would deprecate it and keep it maintained for v1.6.x, but not later versions.)

pabzm:
(Rebased to latest of "master".)

pabzm:
> @alecpl Do you think we should keep "larry" working for future releases?
> 
> I'm not sure yet how much work it would be to make larry fit for this change, but before I even try I want to be sure that you consider it necessary. (I would deprecate it and keep it maintained for v1.6.x, but not later versions.)

@alecpl Could you give me your opinion on this?

Please also review the code.

pabzm:
@alecpl Could you please let me know why you don't react on this PR? Are there technical concerns? Or just a lack of time? Or something different?

pabzm:
Would it maybe help to have renown community members review this to give a better standing ground and ease the burden?

pabzm:
@johndoh @mvorisek If you have the time and headspace, I'd appreciate a review or any comment on this PR.

alecpl:
> How about we decide to go forward with breaking or structural changes targeting versions 2.x, and at the same time commit to maintaining a version 1.x by providing bug fixes (and maybe relevant improvements) without an EOL?

This is exactly the plan, we do not backport new features (with a few exceptions) into old version branches. So, I'm not afraid of breaking stable versions. I'm afraid of supporting them being much harder with architectural changes like this. And I don't see that to be the main concern anyway.

I need more time for this PR. This one is the hardest one. It's not about the code quality of the PR nor the idea itself. There are obvious pros, but there are also cons.

pabzm:
I finished working on this, added a Changelog line, rebased onto the "master" branch, and cleaned up the git history. (From my point of view this shouldn't be squashed, all commits are made on purpose to stand out in the history in case someone is looking for certain changes.)

alecpl:
At least 6 commits here could be extracted and merged as they are not really part of the main feature.

pabzm:
@alecpl I extracted some commits into their own pull-requests as you see. The commits on top of the main feature don't apply cleanly without that "main commit", so I'd rather like to avoid splitting them off, too – unless you insist?

PR #9520, opened 2024-06-26
smime decrypt support added

VM-IPLNet:
written the functions the function decrypt in the /plugins/enigma/lib/enigma_driver_phpssl.php and written the function parse_smime_encrypted in /plugins/enigma/lib/enigma_engine.php
created the function smime_decrypt in /plugins/enigma/lib/enigma_engine.php to mimic the pgp_decrypt

Tested not with master, but with the 1.16.7 version we have in production, i saw that a verify function for smime is now commented, in may version is not and the example msg i have worked just fine.

its not a complete implementation but it allows user to se smime encrypted messages. (partially related to issue #4977)


VM-IPLNet:
it only worked in a cases that the e-mail was signed and the public key was sent in it. A mail from microsoft outlook worked...

PR #9525, opened 2024-06-28
Docker compose for testing

pabzm:
I couldn't get the browser tests to run green locally, and felt uneasy about running a downloaded binary with superuser-rights (the IMAP-server) on my machine anyway. So I decided to build a testing setup with docker compose, which depends on external containers for the browser and the IMAP-server and helps to make the environment reproducecable. Additionally it is faster for repeated runs, because it doesn't repeatedly downloads the server-jar, installs the webdriver, etc.

And because more people might find it useful I'm posting this pull request.

We could also rebuild the automated testing workflows to use services (external containers) to speed things up. But before I do that I wanted to know if this gets merged or provokes discussion in the first place.

The container image needs to be built locally, currently. If this gets merged I'd set it up to be built repeatedly on github and to be downloadable.


alecpl:
I don't like docker and I have no clue about docker, but...
1. I'm not against using it in CI, as long as possibility to use local/external IMAP server is still there.
2. There's  more or less official roundcube/roundcubemail-docker repository, so we should not add any docker files here, unless they reside in .ci folder.

pabzm:
Do I understand correctly that you would merge this if the image is built in https://github.com/roundcube/roundcubemail-docker and we drop the `.docker` directory here?

> I'm not against using it in CI, as long as possibility to use local/external IMAP server is still there.

How would you currently make the CI to use an external IMAP server?

alecpl:
> How would you currently make the CI to use an external IMAP server?

Let's use docker in CI. I want possibility to run tests (manually) against an external IMAP server (and local greenmail).

As for docker. I don't care what the image we use or build it is coming from. I don't want docker related stuff outside of .ci directory.

pabzm:
This setup works in the CI and in a container, and still allows to run the script locally as well as using an external IMAP-Server. Please review!

miaulalala (approved):
Nice!

.github/workflows/tests.yml
```diff
@@ -92,7 +77,7 @@ jobs:
         run: composer update --prefer-dist --no-interaction --no-progress
 
       - name: Roundcube configuration
-        run: cp .github/config-test.inc.php config/config-test.inc.php
+        run: cp .ci/config-test.inc.php config/config-test.inc.php
```
alecpl, line 80: You should call run_tests.sh on Windows too.

tests/Browser/TestCase.php
```diff
@@ -38,7 +38,6 @@ protected function newBrowser($driver)
     public static function prepare(): void
     {
         static::startWebServer();
-        static::startChromeDriver();
```
alecpl, line 41: Maybe `if (!getenv('WEBDRIVER_CONNECT_URL'))`...?

alecpl:
Looks good, almost there.

.github/workflows/tests.yml
pabzm: Good point, thank you!

But why can't? The script runs successfully on the Windows-runner, too.

alecpl:
Looks good, but can we move compose.yaml from root?

pabzm:
> can we move compose.yaml from root?

I moved it into `tests/`, is that ok?

alecpl:
`.ci` folder would be a better location.

alecpl:
Everything else is in .ci including docker images and run scripts.

pabzm:
I don't really agree, but don't find it very important either and want this merged, so here you go. Can you merge it now?

PR #9535, opened 2024-07-05
Test setup for local and CI using containers

pabzm:
This uses standalone containers for the greenmail IMAP server and the
standalone browser.
A testrunner image is built in the CI (for `linux/amd64` only, because
Github doesn't support multi-platform building on their default
runners and we don't have our own.)

This setup helps to run the tests (reproduceably) also locally.
Previously, on my machine, they produced varying results.
It also reduces the dependencies for running the browser test.
Local execution only depends on `docker compose`, no other tools
(previously it required `sudo`, `java`, and some more.)

The previous solution should still work, if you want it.

The scripts are stored in a directory called `.ci` to hide them a little
and avoid confusion with the container images from the
`roundcubemail-docker` repo.


pabzm:
(This was intended as a replacement for another PR, but only after opening it I realized that I don't need it.)

Issue #9549, opened 2024-07-15
better error messages when config.inc.php does not have the right ownership / is not readable by roundcube


JohnRDOrazio:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have [searched](https://plugins.roundcube.net/) for plugins that provide already provide the feature

### Proposal

Roundcube should check whether the `config.inc.php` file and / or `default.inc.php` file is / are readable, and if not should show a more specific error rather than just "Internal error".

### Motivation and context

I recently ran into this situation when uninstalling and reinstalling Roundcube as a Plesk component. I recreated the `config.inc.php` file but forgot to change ownership of the file. I was getting an "Internal error" screen when trying to access webmail, but couldn't for the life of me figure out what was causing the error. I wound up creating a ticket with Plesk support, they didn't identify the exact issue but when they mentioned that probably it was some setting in the `config.inc.php` file that was causing the issue, that at least led me to look closer at the configuration file and realize that it probably didn't have the right ownership. As soon as I made the file group readable by `roundcube_sysgroup` voilà I was again able to access Roundcube webmail.

alecpl:
Done.

PR #9550, opened 2024-07-15
trigger error when config files are not readable

JohnRDOrazio:
fixes issue #9549

I just tested on my own installation and it is working correctly, when any config file has the wrong ownership I am now seeing in the logs:
```bash
[15-Jul-2024 18:03:36 UTC] PHP Warning:  /usr/share/psa-roundcube/config/config.inc.php is not readable. Please check ownership of file in /usr/share/psa-roundcube/program/lib/Roundcube/rcube_config.php on line 312
```

### DIFF
--- a/program/lib/Roundcube/rcube_config.php
+++ b/program/lib/Roundcube/rcube_config.php
@@ -300,8 +300,8 @@ class rcube_config
 
         foreach ($this->resolve_paths($file) as $fpath) {
             if ($fpath && is_file($fpath)) {
-                if(false === is_readable($fpath)) {
-                    trigger_error($fpath . ' is not readable. Please check ownership of file', \E_USER_WARNING);
+                if (false === is_readable($fpath)) {
+                    trigger_error($fpath . ' is not readable', \E_USER_WARNING);
                 } else {
                     // use output buffering, we don't need any output here
                     ob_start();
--- a/program/lib/Roundcube/rcube_config.php
+++ b/program/lib/Roundcube/rcube_config.php
@@ -300,12 +300,12 @@ class rcube_config
 
         foreach ($this->resolve_paths($file) as $fpath) {
             if ($fpath && is_file($fpath)) {
-                if (false === is_readable($fpath)) {
+                if (is_readable($fpath) === false) {
                     trigger_error($fpath . ' is not readable', \E_USER_WARNING);
                 } else {
                     // use output buffering, we don't need any output here
                     ob_start();
-                    include($fpath);
+                    include $fpath;
                     ob_end_clean();
 
                     if (isset($config) && is_array($config)) {
--- a/program/lib/Roundcube/rcube_config.php
+++ b/program/lib/Roundcube/rcube_config.php
@@ -299,20 +299,25 @@ class rcube_config
         $success = false;
 
         foreach ($this->resolve_paths($file) as $fpath) {
-            if ($fpath && is_file($fpath) && is_readable($fpath)) {
-                // use output buffering, we don't need any output here
-                ob_start();
-                require $fpath;
-                ob_end_clean();
+            if ($fpath && is_file($fpath)) {
+                if(false === is_readable($fpath)) {
+                    trigger_error($fpath . ' is not readable. Please check ownership of file', \E_USER_WARNING);
+                } else {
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
Perhaps even better would be to indicate the exact process user name and group that are needed for ownership:
```php
                    $processUser = posix_getpwuid(posix_geteuid());
                    $groupid   = posix_getegid();
                    $groupinfo = posix_getgrgid($groupid);
                    trigger_error($fpath . ' is not readable. Please check ownership of file, it should be owned by user ' . $processUser['name'] . ' or at least by group ' . $groupinfo['name'], \E_USER_WARNING);
```
I can make another commit if reviewers think that it would be a correct enhancement.

program/lib/Roundcube/rcube_config.php
```diff
@@ -299,20 +299,25 @@ public function load_from_file($file)
...
-                if (isset($config) && is_array($config)) {
-                    $this->merge($config);
-                    $success = true;
-                }
-                // deprecated name of config variable
-                if (isset($rcmail_config) && is_array($rcmail_config)) {
-                    $this->merge($rcmail_config);
-                    $success = true;
+            if ($fpath && is_file($fpath)) {
+                if(false === is_readable($fpath)) {
```
pabzm, line 303: Please insert a space before the first bracket here.

program/lib/Roundcube/rcube_config.php
```diff
@@ -299,20 +299,25 @@ public function load_from_file($file)
...
+                    trigger_error($fpath . ' is not readable. Please check ownership of file', \E_USER_WARNING);
```
pabzm, line 304: The problem might be due to other access restrictions than ownership, too. Maybe the message should be a little more general (or use just the first sentence, it has already all the relevant information).

pabzm (commented):
Besides the two small nitpicks this looks good! Thank you for the contribution!

I wouldn't add more code to check for the owner and group names. If we do that it would have to work on different platforms and that's going to take more work, which is not really worth the effort, I think.

pabzm (changes requested):
(placeholder to make the review marked as having requested changes, github doesn't allow me to change that after posting...)

program/lib/Roundcube/rcube_config.php
JohnRDOrazio: added missing space

program/lib/Roundcube/rcube_config.php
JohnRDOrazio: cut the error message to the first sentence

JohnRDOrazio:
Does the second commit satisfy the requirements? Let me know if there are any other useful changes I can commit.

JohnRDOrazio:
Should I fix the remaining coding style complaints? I see there is a rule for non-yoda-style. I had been encouraged by a developer on stackoverflow to start implementing yoda-style comparisons to make for better code debugging, but if yoda-style is considered less readable I can fix it.
I didn't touch the include but I see it's also complaining about the include using parentheses. I have started implementing `phpcs` in a few projects of my own, using the base `PSR12` ruleset which doesn't care about parentheses for includes. I can go ahead and remove the parentheses if that's considered more readable.

alecpl:
Yes please. The rest looks good and it will be merged after the project's code style is used.

JohnRDOrazio:
Code style is now passing, I'm now seeing that in the `origin/master` branch `require` is used rather than `include`, I don't believe I had changed anything in that area of the code but if you prefer that it match `master` I can set it to `require`.

JohnRDOrazio:
I'm seeing that that change took place in commit https://github.com/roundcube/roundcubemail/commit/1e360999b261002ccdabb353bc951025179fb6e3, which goes back to April 11th. I guess my fork wasn't quite up to date...

JohnRDOrazio:
If I try setting `include` to `require` my VSCode is telling me:
```
File is being conditionally included; use "include" instead
(PEAR.Files.IncludingFile.UseInclude)
```
Perhaps we can leave `include`?

alecpl:
I prefer "require" here.

JohnRDOrazio:
> I prefer "require" here.

Ok done

PR #9551, opened 2024-07-16
Make pgpmime-support known after Mailvelope init

pabzm:
The support for pgpmime was only checked initially after the login. If a user e.g. authorizes a domain in Mailvelope only after a login, Mailvelope now instantly works, instead of (silently and undocumentedly) requireing a logout+login-cycle.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -3997,6 +3997,9 @@ function rcube_webmail() {
 
     // Load Mailvelope functionality (and initialize keyring if needed)
     this.mailvelope_load = function (action) {
+        // Make the server code aware that this browser now knows about
+        // PGP/MIME (would otherwise only be recognized after the next login.
+        this.env.browser_capabilities.pgpmime = 1;
         var keyring = this.env.mailvelope_main_keyring ? undefined : this.env.user_id,
             fn = function (kr) {
                 ref.mailvelope_keyring = kr;


alecpl:
This does not look like a best solution. If you take a look at how `mailvelope_load()` is invoked you'll see that  it is only when Mailvelope is enabled. There's no need to call `pgpmime_support_check()`. It will not set the `env` variable in some circumstances.

I see three solutions:
1. I think we should just set `rcmail.env.browser_capabilities.pgpmime = 1` in `mailvelope_load()`.
2. Unset `$_SESSION['browser_caps']['pgpmime']` when enabling Mailvelope.
3. Always check for pgpmime capability as it is not a heavy check (in `browser_capabilities_check()`).

2nd is not great as it will require an extra http request. From 1. and 3. I have no preference.

pabzm:
I chose the first alternative since it's the most straightforward. Thank you for the hint on my misunderstanding of the code!

program/js/app.js
```diff
@@ -3997,6 +3997,9 @@ function rcube_webmail() {
 
     // Load Mailvelope functionality (and initialize keyring if needed)
     this.mailvelope_load = function (action) {
+        // Make the server code aware that this browser now knows about
+        // PGP/MIME (would otherwise only be recognized after the next login.
+        this.env.browser_capabilities.pgpmime = 1
```
alecpl, line 4002: Missing semicolon.

alecpl (changes requested):

program/js/app.js
pabzm: Fixed 👍

Issue #9566, opened 2024-08-02
Options `'socket'` in `$config['imap_conn_options']` ignored in IMAP connector through `$this->prefs['socket_options']`


alexboss:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I can recreate the issue with all plugins disabled

### Describe the issue

Context: my server has IPv4 / IPv6 enabled. I use roundcube to fetch mail through IMAP on a server that only provides IPv4 (no AAAA record for the hostname `mail48.lwspanel.com`). Unfortunately, roundcube always times out when connecting. If I disable IPv6 on my server then roundcube works fine. But I don't want to disable IPv6 on my server and would prefer to force IPv4 when connecting to IMAP using sockets.

So, I need to pass socket options to the IMAP connector. I've set `config/config.inc.php` with the following:

```
// Force IPv4 (LWS mail mail48.lwspanel.com does not provide any IPv6 )
$config['imap_conn_options'] = [
        'socket' => [
                'bindto' => '0:0'
        ],
];
```

But unfortunately they are ignored in the IMAP connector (while they work fine for the SMTP connector when defined in `$config['imap_conn_options']`

I believe the error lies in `program/lib/Roundcube/rcube_imap_generic.php` line 1042:

```
$options = array_intersect_key($this->prefs['socket_options'], ['ssl' => 1]);
```

The `'socket'` options are actually stripped out. One solution would be to modify this line for:

```
$options  = array_intersect_key($this->prefs['socket_options'], ['ssl' => 1, 'socket' => []]);
```

What do you think about this approach ? 

Thank you and best regards.

### What version of PHP are you using?

8.1

### What version of Roundcube are you using?

1.6.7

alecpl:
Fixed in fdf19f3e7c023.

PR #9567, opened 2024-08-02
Use options 'socket' in $config['imap_conn_options'] for IMAP connector - Fix proposal for https://github.com/roundcube/roundcubemail/issues/9566

alexboss:
Fix proposal for https://github.com/roundcube/roundcubemail/issues/9566


alecpl:
Fixed in fdf19f3e7c023.

PR #9585, opened 2024-08-15
Check if attachment is actually(!) referred to

### LINKED ISSUE #9565: Attached picture is not shown if text-part is present in multipart/mixed and image-part as Content-ID
### Prerequisites

- [ ] I can recreate the issue with all plugins disabled

### Describe the issue

Hello

I've composed two similar mails from two public email interfaces (gmail and yandex-mail) and attached an jpeg 
тест
 w/ russian characters in filename (тест.jpg).

Here 

you can see email from gmail - there's no in-line shown attachment, but it detects that the mail has an attachment

And now you can see 

email from yandex-mail and the attachment shown as expected

I downloaded source text for these mails and here's the diff in content part: 
diff

My environment is Chromium ver. 126.0.6478.126 (official build), for Linux Mint (64 bit)

Hope that's enough info for eliminating the bug
WBR, Oleg

Other

8.2.21

1.7-git [GIT 20240728.1158]

### JavaScript errors

_No response_

### PHP errors

_No response_

pabzm:
This finishes what #9472 intended to do – but didn't actually do, as I found out.

The code now checks for each non-text mime-part in a multipart-part if its `Content-ID` or `Content-Location` is (probably) used in a sibling HTML-part, and only if that matches the respective mime-part is considered an "inline" attachment (that won't show up as downloadable or below the message content).

The second commit in this PR makes sure that for all image-parts, all mime-part-headers are loaded from the server, in order to actually get hands on the `Content-Location`-header (which isn't always fetched in the first place). It is limited to image-parts because those are the most common ones maybe having a `Content-Location`-header and I assumed that we shouldn't load the headers for every mime-part, so this seems like a workable real-world distinction for me.

One could probably change how the `BODYSTRUCTURE` response is fetched and parsed to ensure a `Content-Location`-header is always fetched in the first place, but I didn't dare to touch that code.

This fix closes #9565


program/lib/Roundcube/rcube_message.php
```diff
@@ -1006,13 +1006,37 @@ private function parse_structure($structure, $recursive = false)
...
+                        // text/html-part sibling (i.e. that is part of the
+                        // same `$structure`).
+                        $html_parts = $this->find_html_parts($structure);
+                        if (empty($html_parts)) {
+                            $this->add_part($mail_part, 'attachment');
+                            continue;
+                        }
+                        // Note: There might be more than one HTML part, thus
+                        // we use a callback and concatenate the results.
+                        $html_content = implode('', array_map(function ($part) { return $this->get_part_body($part->mime_id); }, $html_parts));
```
alecpl, line 1026: There's indeed my @TODO comment 70 lines below to get the HTML bodies and check for references. However, this has a performance impact, that for a message with many images and big HTML content might be noticeable. As of now I considered fetching part headers acceptable, but part body is another story.

We need some more considerations. For example, when loading an image attachment (`rcmail_attachment_handler`) maybe we could fetch the image without needing to parse the message structure and loading HTML parts again only to get the attachment part data.

Maybe checking for references must be done outside of `rcube_message`. So `rcube_message` is not that heavy (to not slow down all the cases where we deal with the message), but do not really need the full structure information, e.g. when viewing source or downloading the message, or when dealing with a single attachment.

Or maybe we need to use cache. It might not help much when dealing with parallel requests (loading image attachments) though. And caches are usually optional.

pabzm: I see your point and am working on it.

olegStreejak: FYI https://github.com/roundcube/roundcubemail/issues/9565#issuecomment-2327319524

pabzm: I changed the approach so the reference checking will be only done on demand. I'd like to test this, but I have a hard time figuring out how, since the entire class essentially depends on the response to an IMAP `bodystructure` command, which I'd like to avoid mocking.

pabzm: I made some progress and am using the approach in #9460 for testing, but this needs more tests to be ready. I'm mostly afk until the beginning of January, though.

pabzm:
I would like to integrate a lot more messages into the "Message Rendering" tests, but the rest of the code is ready as it is, and I can supplement additional tests later, too.

tests/MessageRendering/data/greenmail/test-message-rendering@localhost/INBOX/spam-with-image-data-uri.eml
```diff
@@ -0,0 +1,6734 @@
...
+<div style="text-align: center;"><a href="https://firebasestorage.googleapis.com/v0/b/project-6510355573021224549.appspot.com/o/08_08_24%2F8%2FFressnapf.html?alt=media&amp;token=b6a90819-bea0-40bd-9ee3-0c6587ec9a72"><img src="cid:2478d4f4b6b373ac9292cdbab380ee6bd4e4a16ee4feef5112ac3fe61b1303fe"/></a></div>
+
+<div class="signature">&nbsp;</div></div></body></html>
+--abmob-f9fb5c85-0819-4385-9568-6687985eef93
+Content-Type: image/png
+Content-Disposition: inline
+Content-ID: <2478d4f4b6b373ac9292cdbab380ee6bd4e4a16ee4feef5112ac3fe61b1303fe>
+Content-Transfer-Encoding: base64
+
+iVBORw0KGgoAAAANSUhEUgAAAlsAAAKYCAYAAABEuT1rAAAAAXNSR0IArs4c6QAAAARnQU1BAACx
```
alecpl, line 26: It would be good to keep the test data small. I.e. use some simple/small images.

alecpl (commented):
I'll need more time to properly test this.

alecpl:
I've tested a bit with various messages. I've found one regression: a forwarded message with eml attachment - there's a difference in displaying it with this patch and without. The attached message headers section is missing. Also if you open the eml attachment the headers will be missing.

ps. I noticed that we FETCH BODY.PEEK[2] for such messages, i.e. we fetch the whole attached message. It seems not optimal if it contains big attachments (I'd have to investigate that separately) - it's the same in release-1.6, so nothing new, I suppose.

tests/MessageRendering/data/greenmail/test-message-rendering@localhost/INBOX/spam-with-image-data-uri.eml
pabzm: I was using original emails, but I see the point, we can save quite some byte here 👍

program/lib/Roundcube/rcube_message.php
```diff
@@ -1111,27 +1150,28 @@ private function get_mime_numbers(&$part)
...
      */
-    private function add_part($part, $type = null)
+    private function add_attachment($part)
     {
         if ($this->check_context($part)) {
             // It may happen that we add the same part to the array many times
             // use part ID index to prevent from duplicates
-            switch ($type) {
-                case 'inline':
-                    $this->inline_parts[$part->mime_id] = $part;
```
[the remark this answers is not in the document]
pabzm: Good catch, thank you!

pabzm:
> I've found one regression: a forwarded message with eml attachment

Thank you for finding this! It's present in the "master" branch already, though. I'll try to find out which change is to blame and how to fix it.

alecpl:
> > I've found one regression: a forwarded message with eml attachment
> 
> Thank you for finding this! It's present in the "master" branch already, though. I'll try to find out which change is to blame and how to fix it.

It works fine in master.

pabzm:
> It works fine in master.

I double-checked and still see the wrong behaviour on the lastest state of the "master"-branch.

I'm testing with the attached message, can you please check if it's displayed properly in your setup?

[Fwd Lines.eml.zip](https://github.com/user-attachments/files/18424684/Fwd.Lines.eml.zip)

And can you provide me with your test message?

alecpl:
Ah, sorry. You're right. It works in release-1.6, but not in master. My bad.

pabzm:
This is what I see with Roundcubemail v1.6.9:

<kbd>
Screenshot RC 1.6.9
</kbd>

---------

And this is what I see with Roundcubemail "master" branch:

<kbd>
Screenshot RC 'master' branch
</kbd>

pabzm:
Actually my fix does fix the bug if I work thoroughly: https://github.com/roundcube/roundcubemail/pull/9753

pabzm:
Rebased onto the "master" branch to fix a conflict

PR #9600, opened 2024-08-22
Allow skins to override translations

DoobleD:
I'm not familiar with the codebase, but I haven't found a way for skins to override an existing translation, rather than just adding a new one. This PR allows overrides as well.


alecpl:
Sounds like a duplicate of #9076.

pabzm:
Closing this in favour of the earlier created PR

PR #9603, opened 2024-08-24
Use server date year as copyright end year

stefan-muc:
To not have to update it every year or getting it wrong as in 1.6.8 simply use the year of the server date as copyright end year

### DIFF
--- a/installer/index.php
+++ b/installer/index.php
@@ -164,7 +164,7 @@ include $include_steps[$RCI->step];
 </div>
 
 <div id="footer">
-  Installer by the Roundcube Dev Team. Copyright &copy; 2008-2021 – Published under the GNU Public License;&nbsp;
+  Installer Copyright &copy; by the Roundcube Dev Team – Published under the GNU Public License;&nbsp;
   Icons by <a href="http://famfamfam.com">famfamfam</a>
 </div>
 </body>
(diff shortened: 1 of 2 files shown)


alecpl:
I think we should just remove the year(s), as we did in source file headers some time ago. Copyright years do not provide much value. And the info can be derived from git history.

pabzm:
@stefan-muc Thank you for your contribution! As you see it invoked a thought to solve the topic differently. Would you be willing to remove the years instead of changing them? Then we could merge this!

stefan-muc:
Sure I'm willing! I just thought that simply removing it can be done by the team itself - but here you are with a changed pull request.

pabzm:
Of course we could have done it ourselves, but I didn't want to steal the opportunity of a contribution from you. :)

Thank you!

installer/index.php
```diff
@@ -164,7 +164,7 @@
 </div>
 
 <div id="footer">
-  Installer by the Roundcube Dev Team. Copyright &copy; 2008-2021 – Published under the GNU Public License;&nbsp;
+  Installer &copy; by The Roundcube Dev Team – Published under the GNU Public License;&nbsp;
```
pabzm, line 167: Could you please re-add the word "Copyright" (since is used in the other places as well in combination with the symbol).

pabzm (changes requested):

installer/index.php
stefan-muc: Sorry, somehow I missed this. Re-added now!

pabzm (approved):
Thank you! 👍

pabzm:
@alecpl I'll merge this on Monday unless you object. (The CI errors are unrelated, don't happen locally neither.)

Issue #9594, opened 2024-08-17
Installer hints to address issues which are not identified


Remmy:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I can recreate the issue with all plugins disabled

### Describe the issue

With a fresh install, the installer displays:
Sorry but your webserver does not meet the requirements for Roundcube!
Please install the missing modules or fix the php.ini settings according to the above check results.
Hint: only checks showing NOT OK need to be fixed.

However, no check displays "NOT OK".

The problem seems to be that I don't have any supported database extensions available. I think it should display a different warning not (just) mentioning NOT OK, but also ensuring at least one supported database type is available.

### What browser(s) are you seeing the problem on?

Other

### What version of PHP are you using?

8.2.20

### What version of Roundcube are you using?

v1.6.7

pabzm:
@Remmy Thank you very much for the report and welcome to the Roundcube project! Your idea is good, and as you can see @johndoh already converted it into a fix, which will (probably) be merged soon, and then you can claim for yourself to have made Roundcube's installer error message a little more helpful!

Remmy:
Thank you gentlemen, I appreciate the swift response and fix!

alecpl:
Done.

PR #9604, opened 2024-08-25
add a NOT OK message to installer when no database extensions available

johndoh:
for #9594

Untitled-1


pabzm (approved):
Looks good, thank you very much!

Issue #9591, opened 2024-08-17
Contacts import: Contacts with "invalid" e-mail address silently dropped


respiranto:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I can recreate the issue with all plugins disabled

### Describe the issue

When importing contacts from a vCard file, any vcard object that contains an "invalid" e-mail address (as per `rcube_utils::check_email()`, I believe), the vcard object is silently ignored.

Example (note that vCard mandates CRLF as newline sequence):
```
BEGIN:VCARD
VERSION:3.0
N:Doe;Jane;;;
FN:Jane Doe
EMAIL:invalid
END:VCARD
```

I would expect an error or warning message indicating that some contacts failed to import and why.  Optionally, the contact could be imported without the offending e-mail address value.

Note: Creating or modifying a contact via the Roundcube web UI to have an "invalid" e-mail address fails with the error message "Invalid email address: [...]".

The current situation is particularly problematic when `$config['email_dns_check'] = true`, and when previously existing e-mail addresses' domains no longer exist.  In such a case (which is how I found this bug), a contacts export followed by an import will silently fail to recreate the same addressbook.

---
Maybe related: #4198

---
Originally reported at the Debian BTS as [Bug 1078774](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078774), where I was asked to report the bug here instead.  The bug report has slightly changed.

### What browser(s) are you seeing the problem on?

Firefox

### What version of PHP are you using?

v8.2

### What version of Roundcube are you using?

master / 58721e3037671888f82d75fd8e91a8622405af45

alecpl:
Even if we import such a contact with removed email address (which is now possible) we should still display a notice to the user.

I think we should not import such contacts. We should add a list of invalid contacts in the import summary.

pabzm:
@respiranto https://github.com/roundcube/roundcubemail/pull/9605 introduces error messages in case of failed imports. Is that enough to close this issue in your opinion?

rpc-scandinavia:
What about adding a boolean option to the import dialog, like "Force import".
Then import all, and append un-importable data (like an malformed e-mail address)  to the notes, and perhaps prefix the display name with an "* " to indicate import issues.

It is very frustrating having a import stop half way, as a user I prefer to get as much data imported as possible - then I can delete unwanted data and edit the rest.

[...]

PR #9605, opened 2024-08-25
add invalid and error details to vcard import report

johndoh:
for #9591

currently there are 2 silent error cases on vcard import, invalid (fails roundcube contact validation) and error (its valid but i didnt insert for some reason e.g. plugin prevented it) this adds messages about those to the screen.

Untitled-1


program/actions/contacts/import.php
```diff
@@ -187,6 +189,7 @@ public function run($args = [])
                     // skip invalid (incomplete) entries
                     if (!$CONTACTS->validate($a_record, true)) {
                         self::$stats->invalid++;
+                        self::$stats->invalid_names[] = $vcard->displayname ?: $email;
```
pabzm, line 192: Could you insert a check to handle the case in which both, `displayname` and `$email` might be `null` or empty?

Also if you rebase onto the latest "master"-branch you should get rid of the failing tests with PHP8.4.

pabzm (changes requested):
Thank you for the contribution! I really appreciate your attention to issues related to "your" topic!

program/actions/contacts/import.php
johndoh: quite right. I changed it to use `rcube_addressbook::compose_display_name()` which will always return a value because to get this far there must be at least one of `displayname`, `surname`, `firstname` or `email` and this function checks all of those.

pabzm (approved):
Looks good to me, thank you!

program/localization/en_US/messages.inc
```diff
@@ -162,6 +162,8 @@ $messages['importwait'] = 'Importing, please wait...';
 $messages['importformaterror'] = 'Import failed! The uploaded file is not a valid import data file.';
 $messages['importconfirm'] = '<b>Successfully imported $inserted contacts</b>';
 $messages['importconfirmskipped'] = '<b>Skipped $skipped existing entries</b>';
+$messages['importconfirminvalid'] = '<b>Skipped $invalid invalid entries</b>';
+$messages['importconfirmerrors'] = '<b>Failed to import $errors valid contacts</b>';
```
alecpl, line 166: We should probably get rid of HTML tags from all these three messages.

program/actions/contacts/import.php
```diff
@@ -431,7 +435,7 @@ public static function import_confirm($attrib)
     {
         $rcmail = rcmail::get_instance();
         $vars = get_object_vars(self::$stats);
-        $vars['names'] = $vars['skipped_names'] = '';
+        $vars['names'] = $vars['skipped_names'] = $vars['invalid_names'] = $vars['error_names'] = '';
```
alecpl, line 438: This line is probably unnecessary. Also, we should probably not use "magic" `get_object_vars()` and assign the `$vars` elements explicitly.

alecpl (approved):
Approved with some comments.

program/localization/en_US/messages.inc
johndoh: I think removing the `<b>` tags would be a breaking change because the formatting should be moved to the skin to do this properly. I can work on this in a separate PR.

program/actions/contacts/import.php
johndoh: are you happy for me to replace `self::$stats = new stdClass();` with a new specific class like say `rcmail_action_contacts_import_stats` and then we can have something like a `get_vars()` method to initialize `$vars`.

alecpl: I'm fine with `StdClass`, I don't like filling needlessly `$vars` with all properties while we need one at a time. And we don't use *_names variants in localized messages. I would probably remove `$vars` completely.

johndoh: I have removed `$vars` and also the `nr` passed to `gettext` as that is meaningless.

pabzm:
@johndoh Would you add a line to the Changelog? (I can do it as well but would prefer to have it as part of the PR.)

johndoh:
I've added a changelog entry. If you don't like the text or its in the wrong place its no problem.

pabzm:
@johndoh Thanks again for your contribution! 🙏

PR #9640, opened 2024-09-17
Fix test by make sorting independent of locale and case

pabzm:
These tests have been failing locally repeatedly. This fixes them and also works in the CI.

### DIFF
--- a/.ci/docker-images/Dockerfile
+++ b/.ci/docker-images/Dockerfile
@@ -24,7 +24,11 @@ RUN apt-get update \
 		aspell-de \
 		hunspell-en-us \
 		git \
+		locales \
  && apt-get clean
+
+# Some tests require en_US.UTF-8 as locale.
+RUN sed -i 's/^# en_US.UTF-8 /en_US.UTF-8 /' /etc/locale.gen && locale-gen
 
 # TODO: Do we need the multiarch-args? What for?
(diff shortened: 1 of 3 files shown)


pabzm:
@alecpl Do you have a comment on this? I'm not really sure why you chose a case-sensitive comparing function in the first place.

alecpl:
We use `strcoll` because it is locale-aware.

pabzm:
But then we cannot compare its output with a fixed list, right? Because the output depends on the locale?

alecpl:
The solution is to set locale in the test.

pabzm:
I tried without success.

```
putenv('LANG=C');
putenv('LC_ALL=C');
$_ENV['LANG'] = 'C';
$_ENV['LC_ALL'] = 'C';
setlocale(\LC_ALL, 'C');
```

These all don't help, neither with `C.UTF-8` nor `en_US.UTF-8`. Which one is correct needle for this haystack?

I still don't understand what we gain from depending our test result on one specific locale, which we then have to pin in our test?

alecpl:
The test works with `setlocale(\LC_ALL, 'en_US.UTF-8')`, but you have to have this locale installed in the system. It fails with 'C' locale.

Feel free to modify the test to pass (e.g. use upper-case where needed), but I don't think we should change the sorting algorithm.

pabzm:
Does this work for you?

PR #9654, opened 2024-09-27
Fix getting IMAP vendor name

### LINKED ISSUE #9650: Undefined array key name encountered in get_vendor() function
### Prerequisites

- [X] I can recreate the issue with all plugins disabled

### Describe the issue

Undefined array key name encountered in get_vendor() function

_No response_

8.2.x

1.6.9

### JavaScript errors

_No response_

### PHP errors

On line 661 of file program/lib/Roundcube/rcube_imap.php the test `$vendor  = (string) (!empty($ident) ? $ident['name'] : '');` returns  Undefined array key "name"
Here are the datas received  in $ident :
```
array (
  'NAME' => 'Zimbra',
  'VERSION' => '9.0.0_GA_4612',
  'RELEASE' => '20240412104050',
  'USER' => 'courrier@...',
  'SERVER' => 'fd376667-0000-4282-ac83-09617f120abc',
) 
```

pabzm:
In some cases, the array's keys where upper case, and the previous code produced a warning and resulted in an empty string, even though the name was present.

Closes #9650

### DIFF
--- a/program/lib/Roundcube/rcube_imap.php
+++ b/program/lib/Roundcube/rcube_imap.php
@@ -655,7 +655,7 @@ class rcube_imap extends rcube_storage
             $ident = null;
         }
 
-        $vendor = (string) (!empty($ident) ? $ident['name'] : '');
+        $vendor = (string) ($ident['name'] ?? $ident['NAME'] ?? '');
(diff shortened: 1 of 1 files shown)


alecpl:
Maybe we should also normalize rcube_imap_generic::id() result, so it always returns keys in lower-case.

PR #9655, opened 2024-09-27
ManageSieve: make sieve_formattime() parse all time-formatters PHP does.

bennet0496:
Roundcube's `$config['time_format']` understands all PHP time and especially timezone format specifiers listed in the [PHP docs](https://www.php.net/manual/en/datetime.format.php); however, `sieve_formattime()` that used this value via `rcmail.env.time_format` does not.

Therefore if the config value e.g. is `$config['time_format'] = "H:i T"` to include the abbreviated timezone name (which works else where, like the message list), then forms using `sieve_formattime()`, the OOO form, will generate times like "6:00 T" and so on
image
This causes the actual time be set to "Tango Time", i.e. UTC-7
image
Which for users outside of UTC-7 is unexpected behavior.

Therefore I changed `sieve_formattime()` to understand any formatter that may reasonably (or unreasonably, like microseconds or Unix Time. But never say never) be configured by an admin or user. For the timezone values, I opted to pass these values from PHP to JS via `set_env()` as in JS these values are extraordinarily hard to reliably to obtain without additional libraries.
image

I also changed some thing at the beginning of `sieve_formattime()` that looked odd to me. Like per-declaring all variables. And actually used the value that `h` would have been in the `g` and `h` cases. But I'm also happy reverting these changes if they are undesired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 - Managesieve: Add List-ID to the list of headers for creating new sieve-filters (#8307)
 - Managesieve: Support an array in managesieve_host option (#9447)
 - Managesieve: Fix the frontend datetime picker not respecting the 12h format and apending a dangling 's' to the seconds (#9688)
+- Managesieve: Add parsing for all PHP time formatters from `time_format` config to frontend the time picker (#9655)
 - Password: Add `ldap_samba_ad` driver (#8525)
 - Password: Allow LDAP access using LDAP URI and SASL binding (#8402)
 - Password: Use Guzzle HTTP Client in the `pwned` driver
--- a/plugins/managesieve/managesieve.js
+++ b/plugins/managesieve/managesieve.js
@@ -1009,6 +1009,9 @@ rcube_webmail.prototype.managesieve_tip_register = function (tips) {
 function sieve_formattime(hour, minutes) {
     var i, c, h, time = '', format = rcmail.env.time_format || 'H:i';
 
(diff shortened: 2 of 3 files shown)


pabzm:
Thank you very much for your contribution!

Your solution is solid, but please don't include the drive-by improvements into the same PR (or at least not into the same commit). It makes it unnecessary hard to comprehend, which changes are related.

Could you correct that?

bennet0496:
I removed the drive-by changes to the 12h format. Do you want the fix for that here as another commit or do you want another PR for that?

Also I removed the Unix timestamp, because looking over it, it doesn't quite makes sense here, because of the implicitly included date.

pabzm:
Please open a new PR for the 12h-changes. Thank you!

pabzm:
Great, thanks! Now all that's missing from my point of view is an entry to the Changelog. Would you mind adding one? (Feel free to decline!)

@alecpl Any thoughts on this?

bennet0496:
Is the entry ok like this? If not I will revert it :)

pabzm:
Perfect, thank you! 👍

pabzm (approved):

pabzm:
@alecpl I'll merge this after one more week without a comment.

/remind me to merge in one week

pabzm:
@bennet0496 Can you rebase to resolve the conflict, please?

pabzm:
@bennet0496 Thanks again for your contribution! 🎊

PR #9665, opened 2024-10-07
Send a configurable CSP in every HTML response

### LINKED ISSUE #9638: Document a working Content-Security-Policy
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have [searched](https://plugins.roundcube.net/) for plugins that provide already provide the feature

### Proposal

We should document a Content-Security-Policy, that is a strict as possible without breaking any usage of Roundcube.

### Motivation and context

These days, CSPs are a relevant part of web security. We should help people with that. It might also reduce the number of issues in which people have problems due to a CSP, or having questions about it.

pabzm:
The CSP gets adapted to remote objects being allowed or not. It can be configured or disabled via the config option `content_security_policy` (and
`content_security_policy_add_allow_remote`).

The default is pretty weak in order to not introduce a breaking change, but still not useless (e.g. it adds another layer of defence against loading remote objects unless allowed.)

Closes #9638


pabzm:
@alecpl Same question for this: would reviews by other developers help you?

alecpl:
I'm not sure it would help me, but it would likely help the pull request move forward.

pabzm:
Are there other things we can do to ease your load? E.g. get more ticket work off your back? Or release/version/patch-management? I notice you're backporting many changes that go into the `master`-branch to one or two other branches, I guess that's a lot to juggle, too.

pabzm:
@johndoh, @mvorisek Do you feel competent to review this PR? It would help us a lot, even if you're not perfectly sure about all parts, just tell us about your opinion.

program/include/rcmail_output_html.php
```diff
@@ -2717,4 +2719,20 @@ protected function get_template_logo($type = null, $match = null)
...
+    /**
+     * Add the Content-Security-Policy to the HTTP response headers (unless it
+     * is disabled).
+     */
+    protected function add_csp_header(): void
+    {
+        $csp = $this->app->config->get('content_security_policy');
+        if (!in_array($csp, ['', false, 'false'])) {
+            $csp_header = "Content-Security-Policy: {$csp}";
+            if (isset($this->env['safemode']) && $this->env['safemode'] === true) {
```
mvorisek, line 2732: This needs E2E test as it is it can very easily break security if coded wrongly.

alecpl:
I'm not convinced that providing such configuration options is a good idea.

johndoh:
I know nothing about CSP headers. From a Roundcube code style point of view only I would say that 

` if (!in_array($csp, ['', false, 'false'])) {`

is wrong and it should be

`if ($csp = $this->app->config->get('content_security_policy', "default-src 'self' data: blob:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline';")) {`

this means, i think, in common with other config values `null` reverts to a hard coded default and `false` disables (e.g `imap_ns_*`).

I would also say the comment in `defaults.inc.php` needs to mention that `content_security_policy` must end in a `;` or the header will be invalid when concatenated with `content_security_policy_add_allow_remote`

pabzm:
> I'm not convinced that providing such configuration options is a good idea.

@alecpl Why? I thought some people might need to change the CSP (e.g. for plugins using external resources), and this way they wouldn't need to patch the code.

program/include/rcmail_output_html.php
pabzm: Testing is a good point 👍 

Do you have something specific in mind regarding breaking security?

pabzm:
> I know nothing about CSP headers.

@johndoh But your comments are helpful, thank you very much!

> this means, i think, in common with other config values null reverts to a hard coded default and false disables

For some values that's right, but `defaults.inc.php` has a lot of actual default values, too. E.g. most booleans, but also `oauth_user_create_map`, or `blankpage_url`.
I would like to avoid repeating the default value, because it's pretty complex in this case – but at the same time it needs to be visible outside of the actual code as a reference. Thus I find the other way more appropriate.
Is there a declared will or style guide towards one of the default value methods?


> `content_security_policy` must end in a ;

Thank you for that catch! I'l solve that in code so admins can't break the CSP so easily by mistake.

program/include/rcmail_output_html.php
mvorisek: I mean we need to be sure external resources are not fetched (otherwise spammers know what email addresses to spam even more) in every other case than explicitly configured so.

johndoh:
> I would like to avoid repeating the default value, 

I agree with this.

I think in the past there has been issues with `defaults.inc.php` not being properly updated in 3rd party packaged versions of Roundcube (which I acknowledge are not our problem) and people affected by that create issues here. The comment could suggest a value while the code has a reliable default so if `defaults.inc.php` was not updated there is no unpredictable behavior.

Personally I just don't like the `in_array` check, I think its unnecessary because it handles the edge case where someone set 'false' vs `false` and I think in general Roundcube does not do that so why do it for this one var.

program/include/rcmail_output_html.php
pabzm: I agree, but I think this change doesn't enlarge that danger. It only adds a Content-Security-Policy, which should reduce that danger. The `safe`-flag is not touched.

Do you see any way in which this change could be dangerous?

pabzm:
> I think in the past there has been issues with defaults.inc.php not being properly updated in 3rd party packaged versions of Roundcube

Oh, I wasn't aware of that. Thank you for that explanation! Then we maybe should have a hardcoded value for all defaults.

I changed the approach to specifying the default value in `defaults.inc.php`, *and* having a hardcoded default. And to check that these are in sync, I wrote a test. What do you think about this way?

(I'd be happy to extend this approach to more config options, it would reduce some special case handling in `rcube_config`, too.)

> Personally I just don't like the in_array check, I think its unnecessary because it handles the edge case

I also find it ugly, but I really like solid code, which isn't vulnerable to edge cases. But maybe checking against `'false'` is too much. Regarding readability I can turn the check around by using `is_string($csp) && strlen($csp) > 0`.

pabzm:
> I'm not convinced that providing such configuration options is a good idea.

@alecpl Why? Some people might need to change the CSP (e.g. for plugins using external resources).

alecpl:
Because it's easy to do wrong. Because CSP can be set on the http server and Roundcube cannot relax the rules, it can only make them more strict. Because the initial problem was lack of documentation, not lack of options. In other words this might be one of those power-user options that may bring more harm than value.

That being said, I don't really see a middle ground here. Either we set CSP (and need options and ways for plugins) or not.

pabzm:
From your tone and from the experience of other comment threads I assume that you're not interested in an open discussion of the points you listed. (Please prove me wrong.)

Which makes me wonder what you want from this PR. If you want to veto the change, please say so clearly. If not: please suggest a way forward.

alecpl:
I don't want it to sound like a veto, but I think we should not merge this. As other security-related headers that you can set, add the header sample to the `.htaccess` file or extend CONTENT-SECURITY-POLICY section in the INSTALL file (we'll get rid of the `.htaccess` file at some point).

It does not mean we shouldn't make Roundcube compatible with more strict rules in the future, but Roundcube should not set the header itself, imo. Of course, there's always exceptions, we use CSP header on file downloads.

pabzm:
This change increases security by introducing a CSP that is stricter if the `safe`-mode is off. How would you suggest to do that in a HTTP daemon config?

And it increases security of all installations by ensuring from code that the CSP is always sent, not only where an admin took the extra step to configure it.

And it doesn't break anything in a default installation.

Of course people can make themselves a problem with the config option, but that is true for a lot of the config options we already have. How is this one different?

pabzm:
@johndoh @mvorisek Do you have an opinion on this PRs approach vs. Alec's suggestion?

pabzm:
@alecpl Could you please answer to my questions from Nov 7th? I still think this is helpful and it doesn't break anything.

alecpl:
1. I see some warnings in firefox: `Content-Security-Policy: Couldn’t process unknown directive ‘font-src:’`, `Content-Security-Policy: Couldn’t process unknown directive ‘frame-src:’`. You have to remove the colons.

But more importantly:
2. We allow contacts with external photo URLs. This will block them.
3. We allow external images in responses, signatures and mail composer. This will block them.

I don't think we can just say we don't support this anymore.

pabzm:
I removed the wrong colons and I think I fixed the usage of the cases you mentioned, thank you for those hints!

I found another problem: If a mailvelope user wants to automatically lookup missing openpgp keys (to encrypt a message to), the HTTP request is sent from JavaScript directly to the configured keyserver. This would be blocked by a strict CSP.

I see two options:
* Include the configured keyserver as allowed source in the strict CSP, or
* Proxy the search requests through the server and keep the strict CSP as it is.

The latter would keep the security advantage of a stricter CSP, and also improve the user's privacy slightly because it won't allow for profiling based on the client IP address.
This would also be beneficial for a future feature, to lookup keys in a [WKD](https://datatracker.ietf.org/doc/draft-koch-openpgp-webkey-service/), for which hiding client IP addresses would be nice, too.

@alecpl Do you have an opinion on this?

alecpl:
That reminds me of a key search in the enigma plugin, which may have the same issue, I suppose.

I'm not sure we can fix the HTML composer this way, though. When you compose a reply/forward you deal with the original mail content that should get secured, don't you think? Marking the content safe does not look right.

As I said in the beginning this becomes complicated and problematic.

pabzm:
Enigma calls  the same JavaScript-function as the Mailvelope-integration.

Good point about the replies!

I've changed the code to not use the `safe` mode for triggering the lax CSP but a dedicated variable. (Actually I had this change set in place once before but discarded it to save a variable and jump onto `safemode`… should've trusted my initial idea!)

(I'll happily clean up the code before it would be merged, just don't want to stir around in the commits as long as they're continuously reviewed.)

pabzm:
FTR: this is abandoned from my side. Anyone who's interested would be invited to pick it up and push it further.

PR #9676, opened 2024-10-22
Changed character encoding for password to iconv()

jbweb:
To ensure that password for LDAP Samba AD is saved correctly, iconv() must be used instead of password::hash_password().


alecpl:
I don't see why hash_password() with 'ad' method would not work. It executes `rcube_charset::convert('"' . $password . '"', RCUBE_CHARSET, 'UTF-16LE');` which should do the same. Could you explain the problem?

jbweb:
I can't remember why and what was the reason, its a long time ago. Maybe its not necessary to use iconv() directly anymore. I'll close this pull request, and do some tests to see if any issues persists with the hash_password() method or not.

PR #9680, opened 2024-10-26
Enforce leading backslash even in non-yet-namespaced classes

mvorisek:
extracted from #9481

let's land this in a separate PR and minimize future diffs/conflicts


mvorisek:
@alecpl can this PR be merged asap to prevent merge conflicts?

mvorisek:
@alecpl I have rebased this PR, can it be merged to prevent further merge conflicts?

mvorisek:
@alecpl can I ask you to merge this PR?

mvorisek:
@pabzm can you review/merge? I need this one for #9481.

alecpl:
I think I would prefer a `use` statements instead. Or we should do something with the release-1.6 code. Backporting anything has become hard.

mvorisek:
@alecpl thank you for the idea using import to minimize conflicts!

So I reworked this PR:
a) all non-intra project usages are newly with leading `\` - this is how they should be imported once namespaced
b) all intra project usages in currently non-namespaced files are kept, but dummy `use X as X;` imports are added

After https://github.com/roundcube/roundcubemail/pull/9481 the dummy imports will be replaced with meaningful imports.

Code can stay as it is, so conflicts across release branches are minimal.

Leading `\` for non-instra project usages is asserted by CI.

mvorisek:
@alecpl can you please review?

plugins/enigma/lib/enigma_driver_gnupg.php
```diff
@@ -1,5 +1,15 @@
...
 
+use enigma_driver as enigma_driver;
+use enigma_error as enigma_error;
+use enigma_key as enigma_key;
+use enigma_signature as enigma_signature;
+use enigma_subkey as enigma_subkey;
+use enigma_userid as enigma_userid;
+use rcmail as rcmail;
+use rcube as rcube;
+use rcube_user as rcube_user;
```
alecpl, line 11: What's up with these. Why not just a `use` without `as`? It is temporary, I hope?

program/include/rcmail_sendmail.php
```diff
@@ -327,7 +343,7 @@ public function create_message($headers, $body, $isHtml = false, $attachments =
         }
 
         // create PEAR::Mail_mime instance
-        $MAIL_MIME = new Mail_mime("\r\n");
+        $MAIL_MIME = new \Mail_mime("\r\n");
```
alecpl, line 346: For cases like this, when we use an external class many times, I'd like to see a `use` statement too.

plugins/enigma/lib/enigma_driver_gnupg.php
mvorisek: We need `use` because of https://github.com/roundcube/roundcubemail/blob/5dcb791647/.php-cs-fixer.dist.php#L83, the config lines will be removed once the files and properly namespaced and the use changes to namespaced classes.

We need the `use X as X;` to suppress php warning - https://3v4l.org/dS4TX

program/include/rcmail_sendmail.php
mvorisek: The convection in php community is to use backslash /wo import if the classes are not namespaced.

It makes sense, as for ex. `catch (\Exception $e) ` as much more readable than `catch (Exception $e) ` - in the later case, you might think it catches only project's exceptions.

Is this ok to keep?

If you worry stictly about backport conflicts, I can make a followup PR into v1.6 branch.

pabzm:
> @pabzm can you review/merge? I need this one for #9481.

Didn't find the time until now, sorry. From my side all is fine! The offer to contribute to a backport-PR is very helpful, too!

alecpl:
I'm not happy with this change. Where does it lead to, is there an end goal? Are we going to make a big BC break, call it Roundcube 2.0, where e.g. `rcube_utils` becomes `Roundcube\Utils` etc.? I suppose I wouldn't be against it if someone has time to do all this. What about a timeline? Do you foresee any obstacles (plugins maybe)? The intermediate state is a mess.

mvorisek:
@alecpl good questions. We should namespace the classes sooner or later. As doing so will be imply BC break, I created #9481 which namespace the classes, but at the same time provide BCible layer for loading the classes as non-namespaced. So these PRs which might look like a mess are actually very precisely crafted to allow smooth transition.

As #9481 might be hard to review/discuss at once, I created this preparatory PR which focuses solely on enabling `leading_backslash_in_global_namespace` PHP CS Fixer syntax. This is needed for minimizing diffs a) for backporting, b) for reducing the size of #9481.

> I wouldn't be against it if someone has time to do all this. What about a timeline? Do you foresee any obstacles (plugins maybe)? The intermediate state is a mess.

In #9481 I have demonstrated it working incl. the BCible autoloading layer. Nothing is broken, everything works.

If you support this, I would be happy if this PR can be merged as is. This PR is safe and does not imply any BC breaks. It is pure CS update.

mvorisek:
@alecpl do I have your support here in order to namespace the files? I offer a controlled migration path to land most of the required changes asap in order to minimize the future diffs and still keep BC.

mvorisek:
I am going to rework this PR using custom PHP CS Fixer rules and I will open another PR for it.

PR #9684, opened 2024-10-30
Test config option rule: null -> hardcoded default, false -> disabled

pabzm:
An unspoken (as far as I know) rule of Roundcube config options is that setting an option to `null` makes Roundcube use the (hardcoded) default, while setting it to `false` disables it.

This PR introduces a test that checks that behaviour.

### DIFF
--- a/tests/Framework/ConfigTest.php
+++ b/tests/Framework/ConfigTest.php
@@ -48,6 +48,16 @@ class ConfigTest extends TestCase
 
         $this->assertSame(4190, $object->get('test_int'));
 
+        // If the configured value is `null`, expect the fallback value (which is `null` by default).
+        $object->set('test', null);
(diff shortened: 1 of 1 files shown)


pabzm:
@johndoh I'd be interested if you have any comments, too! (Can't request a formal review from you.)

alecpl:
I'm not sure you are not mixing two things. This is indeed how set() and get() methods work. However, loading options from a config file will override the defaults. Which also is as intended.

pabzm:
@alecpl I'd like your feedback on my last comment here

alecpl:
One could argue that using `array_key_exists()` would be better than `isset()`, but this is how it always been so I'd keep that. So, I agree this is correct behavior.

PR #9688, opened 2024-11-04
ManageSieve: Fix timeformat processing for frontend the time picker

bennet0496:
Additionally to the missing time formatters described in #9655, the function `sieve_formattime()` has also two additional bugs processing the existing time formats.

First the 12h format is not processed properly e.g. resulting in 24h format with am/pm specifiers in the timeformat is set to `h:i a`. This is the result of the calculated 12h value not actually being used to the string construction.
image

And second having seconds in the timeformat will append a dangling `s` to the number due the fall through of the last switch case in the function.
image

This PR fix both issues (in separate commits).


pabzm (approved):
Looks good to me, thank you very much!

Could you add a Changelog-entry for the fix?

bennet0496:
added it

pabzm:
@alecpl I'll merge this in one week unless you object.

PR #9690, opened 2024-11-05
Introduce security@roundcube.net as security contact

pabzm:
Using a dedicated email address with a dedicated PGP key allows to give multiple people access while still keeping things under wrap.

A single, private email address as security contact is such a huge bus factor, which we should avoid. Event just a holiday or illness could lead to escalation due to missing replies.

Also, in case of potentially severe security issues Nextcloud's security team must have access to all details and communication. This is already given for all issues reported via hackerone.com, and with this change is now also enabled for issues reported by email.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,7 +10,7 @@ If you found a security issue or vulnerability of the software, please report it
 
 Your report should include clear steps for reproduction and a classification of the found vulnerability.
 
-You can also send an encrypted email message to *alec[at]alec.pl*. You can find the PGP public key on the major public keyservers like [pgp.key-server.io](https://pgp.key-server.io).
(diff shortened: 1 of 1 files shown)


pabzm:
@alecpl I'll provide you with the IMAP credentials and the PGP private key as soon as this is merged. Please let us know what you think.

pabzm:
Related: https://github.com/roundcube/roundcube.github.com/pull/67

alecpl:
I'm fine with it.

pabzm:
> @alecpl I'll provide you with the IMAP credentials and the PGP private key as soon as this is merged. Please let us know what you think.

@alecpl This ist still on my list, currently blocked by an email system problem. Sorry!

PR #9693, opened 2024-11-06
fix ldap sort column usage from RC config

drlight17:
This will fix ldap sort column usage from config in actual RC 1.6.9


program/lib/Roundcube/rcube_ldap.php
```diff
@@ -561,7 +561,7 @@ public function set_pagesize($size)
     public function set_sort_order($sort_col = null, $sort_order = null)
     {
         if (!empty($this->coltypes[$sort_col]['attributes'])) {
-            $this->sort_col = $this->coltypes[$sort_col]['attributes'][0];
+            $this->sort_col = $this->prop['sort'];
```
alecpl, line 564: With this change the `if` condition does not page sense anymore. Anyway, it does not look right to me, but I didn't investigate closer.

PR #9708, opened 2024-11-20
Run CI tests with PHP8.4

pabzm:
This PR is to check which compatibility problems still exists. I'll rebase it repeatedly onto the latest "master" branch.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ["7.3", "7.4", "8.0", "8.1", "8.2", "8.3"]
+        php: ["7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4"]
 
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ["7.3", "7.4", "8.0", "8.1", "8.2", "8.3"]
+        php: ["7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4"]
```
mvorisek, line 18: should be added everywhere, incl. Windows jobs etc.

pabzm: Good point, we might as well go all in in this PR 👍

alecpl:
I fixed plugin-installer and re-run the jobs.

As for php-cs-fixer we have to run it with `PHP_CS_FIXER_IGNORE_ENV=1`, and it should work.

Issue #9719, opened 2024-11-28
Link to the official documentation


Cwpute:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have [searched](https://plugins.roundcube.net/) for plugins that provide already provide the feature

### Proposal

Instances of Roundcube do not come with a link to the official documentation. This is a shame, as it would certainly help users would not naturally search for "roundcube documentation" when asking for help on a subject regarding Roundcube.

### Motivation and context

As i wanted to clarify what the "Flag" action does to my emails, i searched for informations directly in Roundcube. I was suprised to see the About page didn't include something to help me answer my question, but also lacking other informations one would expect in such place (as suggested by issue #8395 ). I didn't find any Help page or documentation elsewhere on the webapp so i reverted back to querrying the internet for an answer (which i did find, along with the official documentation).

alecpl:
1. Our documentation is quite bad and outdated.
2. There is the Help plugin that does what you need.

So, overall I'm not sure it would be a good idea.

Cwpute:
Didn't know there was a plugin for that. Thanks for the heads-up, i'll talk about it to my instance maintainers if it's relevant.

> Our documentation is quite bad and outdated.
Would a call for action to contribute and update the documentation be relevant or useful ?

Issue #8395, opened 2022-01-12
Add relevant links in the About modal


nclm:
When clicking on “About” in Roundcube, the user gets a modal starting like this:

> ### Roundcube Webmail 1.5.1
> 
> Copyright © 2005-2021, The Roundcube Dev Team
> 
> This program is free software; you can redistribute it and/or modify it under the terms of the [GNU General Public License](http://www.gnu.org/licenses/gpl.html) as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
> Some [exceptions](https://roundcube.net/license) for skins & plugins apply.

I believe it really misses a link to Roundcube.

Other improvements could be writing a line about what Roundcube is (after all, if its the default solution provided by my e-mail host, I might except something more informative than technical when clicking on “About”), and possibly to the Github repo.

For instance it could be something like this:

> ### Roundcube Webmail 1.5.1
> 
> This webmail is powered by [Roundcube](https://roundcube.net), a fully-featured browser-based e-mail client.
> 
> Roundcube is free software. You can [contribute to making it better](https://roundcube.net/contribute/) and you can redistribute it and/or modify it under the terms of the [GNU General Public License](http://www.gnu.org/licenses/gpl.html) as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. Some [exceptions](https://roundcube.net/license) for skins & plugins apply. 
> 
> Copyright © 2005-2021, The Roundcube Dev Team

thomascube:
I agree that the wording of the about dialog could be improved and also help reducing support requests from users to us.

Cwpute:
Additionally, a link to the official documentation would be very welcome !

nclm:
I made a PR here: #9720

Things can be rephrased of course, and I would reorder things to have the paragraphs being actual paragraphs (instead of <br/>s) and have the copyright notice very last, but didn’t know how to do that.

nclm:
Great point, I updated :)

alecpl:
Adding a link to Roundcube will not "help reducing support requests from users to us", quite the opposite.

Cwpute:
Good point @alecpl , too many issues are not dependant on RC per se but on the specific installation supplied to end users.
Maybe then, remove the contribution link and only link the main RC website https://roundcube.net/ 
[...]

PR #9720, opened 2024-11-28
Add useful links in About

nclm:
Add links to Roundcube, its documentation and the contribute page to the About window.

Solves  #8395 and  #9719

### DIFF
--- a/program/actions/settings/about.php
+++ b/program/actions/settings/about.php
@@ -43,7 +43,7 @@ class rcmail_action_settings_about extends rcmail_action
             },
             'license' => static function () {
                 return 'This webmail is powered by <a href="https://roundcube.net" target="_blank">Roundcube</a>, a fully-featured browser-based e-mail client. '
-                     . 'You can find documentation in <a href="https://docs.roundcube.net" target="_blank">the user guide</a>.'
+                     . 'You can find help and further documentation in <a href="https://docs.roundcube.net" target="_blank">the user guide</a>.'
                      . '<br/><br/>'
                      . 'Roundcube is free software. You can <a href="https://roundcube.net/contribute/" target="_blank">contribute to improve it</a> '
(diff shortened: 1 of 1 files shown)


program/actions/settings/about.php
```diff
@@ -42,12 +42,13 @@ public function run($args = [])
...
-                    . 'Some <a href="https://roundcube.net/license" target="_blank">exceptions</a> '
-                    . 'for skins &amp; plugins apply.';
+                return 'This webmail is powered by <a href="https://roundcube.net" target="_blank">Roundcube</a>, a fully-featured browser-based e-mail client. '
+                     . 'You can find help and further documentation in <a href="https://docs.roundcube.net" target="_blank">the user guide</a>.'
+                     . '<br/><br/>'
+                     . 'Roundcube is free software. You can <a href="https://roundcube.net/contribute/" target="_blank">contribute to improve it</a> '
+                     . 'and redistribute it and/or modify it under the terms of the <a href="http://www.gnu.org/licenses/gpl.html" target="_blank">GNU General Public License</a> '
+                     . 'as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. '
+                     . 'Some <a href="https://roundcube.net/license" target="_blank">exceptions</a> for skins and plugins apply.';
             },
```
alecpl, line 52: I have a few issues with this:
1. Documentation is not license, so probably would be better to separate license info from the "about and doc" info. I.e. leave the "license" content as it was, but provide an additional template handler.
2. People may have the help plugin configured to direct to a different documentation URL. Maybe do not include this new element if the plugin is enabled?
3. I think the "contribute" part is not needed at all.
4. Providing users with more links they can follow may increase number of support request from them on our side. We don't want that.

Neustradamus:
@nclm: Have you seen @alecpl comments?

PR #9739, opened 2025-01-02
Run PHP-CS-Fixer in parallel

melroy89:
Speed up the php-cs-fixer by allow the tool to run in parallel (by default it only uses 1 core). 

See also: https://cs.symfony.com/doc/usage.html


melroy89:
Thanks for merging.

PR #9753, opened 2025-01-15
Fix displaying message/rfc822 parts

pabzm:
The previous code contained a simple error that assigned the body part to the $headers variable (which then couldn't be parsed and got discarded).


pabzm:
I'll add a message rendering test to this.

pabzm:
This is ready for review

PR #9754, opened 2025-01-16
Cleaned up the code a bit

itsnatehow:


pabzm:
@itsnatehow Hello! I'm not sure what you are up to, but deleting all tests is not really a good start, to be honest. Maybe you could explain your plans, so we can give you a little bit of feedback from our point of view?

PR #9755, opened 2025-01-19
additional_message_headers.php: support CALLABLE/callback via config

jb4z:
Good day all,

The purpose of this patch is to have Roundcube execute a callback function during runtime, as defined in config.inc.php, for more complex tasks going beyond static strings or their replacements.

For example, one could configure something like this:

```
$config['additional_message_headers']['X-Sender'] = null;
$config['additional_message_headers']['X-RC-USR'] = (function() {
                $d = json_encode([
                    'u' => rcube::get_instance()->get_user_name(),
                    'r' => $_SERVER['REMOTE_ADDR'],
                    'a' => empty($_SERVER['HTTP_USER_AGENT']) ? '-' : $_SERVER['HTTP_USER_AGENT'],
                    't' => $_SERVER['REQUEST_TIME']
                ]);
                return base64_encode($d); # should also be encrypted ;)
        });

```
In this example,
a) disables the cleartext X-Sender header; 
b) adds a dynamic header X-RC-USR in base64-JSON-encoded form, which could later be used for compliance purposes. If this header is automatically processed by the mail gateway, further analysis could aid in detecting abuse patterns, while not directly exposing this sensitive information as human-readable text; if properly encrypted (out of this scope), this could eliminate privacy concerns.

One of the ideas behind this is that using Roundcube (or any other webmailer) usually masquerades the original user's IP address by the webmailer's server IP address to the SMTP server; this is not the case when a user talks to the SMTP server directly.

With tight integration into your setup, you will never again have to sift through different logs/correlate IP address information just to find the guy who sent this message which the person behind the user's login denies having sent.

This patch also reduces two arrays ($search, $replace) into one ($map) for maintainability.

### DIFF
--- a/config/config.inc.php.sample
+++ b/config/config.inc.php.sample
@@ -62,5 +62,24 @@ $config['plugins'] = [
     'zipdownload',
 ];
 
-// skin name: folder from skins/
+// Specify Roundcube's Default Skin, equal to the folder name beneath skins/
 $config['skin'] = 'elastic';
+
+// Optional config of the additional_message_headers plugin (Issue #9755; Feb 2025)
+// Uncomment to remove the X-Sender header from outgoing messages:
+// $config['additional_message_headers']['X-Sender'] = null;
+//
+// You can also specify a callback function like this, which would be called when
+// the user clicks the "Send" button, allowing you to access to the rcube object.
+// This example adds an X-RC-USR header with a base64-encoded JSON string.
+// Do not use this in production unless you know what you are doing and your
+// users are informed via Terms of Service or other means of compliance.
+// $config['additional_message_headers']['X-RC-USR'] = (function() {
+//                $d = json_encode([
+//                    'u' => rcube::get_instance()->get_user_name(),
+//                    'r' => $_SERVER['REMOTE_ADDR'],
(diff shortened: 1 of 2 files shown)


alecpl:
Thanks. Please, fix the coding style issue. Also, it would be good to add some note to the config.inc.php.dist file, with some simple example.

jb4z:
Thanks for considering this. Fixed the whitespace issue as requested and updated the config.inc.php.sample, hoping it's not too verbose (and not too opinionated 😬)

config/config.inc.php.sample
```diff
@@ -62,5 +62,24 @@ $config['plugins'] = [
     'zipdownload',
 ];
 
-// skin name: folder from skins/
+// Specify Roundcube's Default Skin, equal to the folder name beneath skins/
 $config['skin'] = 'elastic';
+
+// Optional config of the additional_message_headers plugin (Issue #9755; Feb 2025)
```
alecpl, line 68: The sample should be in `plugins/additional_message_headers/config.inc.php.dist`

alecpl: @jb4z would you mind fixing the last issue I mentioned? It's almost ready to be merged.

PR #9785, opened 2025-02-19
Fix washing html from rcube_attachment_handler

pabzm:
This probably wasn't implemented previously because HTML-parts usually didn't run through get.php.

Picked from #9519

### DIFF
--- a/program/include/rcmail_attachment_handler.php
+++ b/program/include/rcmail_attachment_handler.php
@@ -309,7 +309,13 @@ class rcmail_attachment_handler
         // show images?
         $is_safe = $this->is_safe();
 
-        return rcmail_action_mail_index::wash_html($body, ['safe' => $is_safe, 'inline_html' => false]);
+        if (isset($this->part->replaces)) {
+            $wash_replacements = $this->part->replaces;
+        } else {
+            $wash_replacements = [];
+        }
+
+        return rcmail_action_mail_index::wash_html($body, ['safe' => $is_safe, 'inline_html' => false], $wash_replacements);
     }
 
     /**


program/include/rcmail_attachment_handler.php
```diff
@@ -309,7 +309,7 @@ public function html()
         // show images?
         $is_safe = $this->is_safe();
 
-        return rcmail_action_mail_index::wash_html($body, ['safe' => $is_safe, 'inline_html' => false]);
+        return rcmail_action_mail_index::wash_html($body, ['safe' => $is_safe, 'inline_html' => false], $this->part->replaces);
```
alecpl, line 312: I think we have to check if `$this->part` is not null.

program/include/rcmail_attachment_handler.php
```diff
@@ -309,7 +309,13 @@ public function html()
         // show images?
         $is_safe = $this->is_safe();
 
-        return rcmail_action_mail_index::wash_html($body, ['safe' => $is_safe, 'inline_html' => false]);
+        if (isset($this->part->replaces)) {
```
alecpl, line 312: Should be `isset($this->part)` or could be `$this->part instanceof rcube_message_part`, I suppose.

pabzm: Why? `isset()` also checks the "intermediate" object and returns `false` without a warning, if that isn't set.

alecpl: Hmm.. you're right, it's just not what I'm used to do. If I was about to write this code, it would be: `$replaces = $this->part ? $this->part->replaces : [];`, but I will not insist.

pabzm: I would prefer to keep it as it is, since the current way also protects against `replaces` being `null` (for whatever reason).

PR #9787, opened 2025-02-20
Render each mime-part into an individual iframe

pabzm:
Implementing https://github.com/roundcube/roundcubemail/issues/9465

A new clean restart over the previous pull request.

TODO:

- [x] reading emails using mailvelope
- [ ] check/change larry skin
- [x] Don't "wash" HTML content twice
- [x] Check reading emails with enigma
- [x] Fix hide_blockquote plugin
- [ ] Find out why the "loading" message gets hidden so early in chromium-based browsers.


program/js/app.js
```diff
@@ -4007,8 +4006,10 @@ function rcube_webmail() {
         this.env.browser_capabilities.pgpmime = 1;
         var keyring = this.env.mailvelope_main_keyring ? undefined : this.env.user_id,
             fn = function (kr) {
-                ref.mailvelope_keyring = kr;
-                ref.mailvelope_init(action, kr);
+                this.afterAllContentIframesLoaded(() => {
```
alecpl, line 4009: it should be `ref` not `this`.

program/js/app.js
```diff
@@ -10650,6 +10654,71 @@ function rcube_webmail() {
...
+                // document was parsed, to already provide roughly the
+                // correct height; second time when all resources have been
+                // loaded, to finally ensure the correct height with all
+                // images etc.
+                iframe.addEventListener('DOMContentLoaded', () => this.resize_preview_iframe(iframe));
+                iframe.addEventListener('load', () => {
+                    // Hide "Loading data" message.
+                    $(iframe).siblings('.loading').hide();
+                    // Show remote objects notice
+                    if (iframe.contentDocument.body.dataset.extlinks === 'true') {
```
alecpl, line 10690: I'mk getting "TypeError: iframe.contentDocument is null" when opening any message.

alecpl:
I gave it a try, here are some more issues:
- I see that we do not load our styles.css into the iframes. I.e. drark mode is broken, the plain text messages font is not monospace, signatures aren't greyed, blockquotes unstyled, etc.
- messages charset is broken (for messages in charset other than utf-8).
- the loading element disappears in onload (?), I think it's too late. I.e. sometimes I see both this element and iframe content being loaded. It does not look good. Maybe solution is to style the iframes in an "overlay" style loader, instead of using a separate element.
- performance: the simplest case - opening a plain text message. I get two imap connections both of them do `SELECT INBOX`, `UID FETCH 33469 (UID RFC822.SIZE FLAGS...`, `UID FETCH 33469 (BODY.PEEK[1])`. Even more redundancy when loading a html message with images.
- for performance reasons the format-change action (switching between HTML and plain text) should probably be changed to not reload the whole preview frame, but only the `#messagebody` element. Not a big deal for now.
- when Enigma plugin can't decrypt a message the "loading data" message never disappears. I also think this plugin will need some other non-trivial updates.

pabzm:
Thank you for the review!

The missing styling was an artefact of a rebase after the breaking changes. I fixed that now, but during that I found that displaying text attachments in the new window had additional bugs and differences to the current state of the "master" branch.

While fixing those I stumbled upon something unexpected: text/plain parts are rendered differently from the preview vs. attachment view (new window, `_action=get`). The former wraps it in a `div.pre` with styling, the latter wraps it in a `pre` and applies no styling.

@alecpl Can you explain why that is and/or if that needs to be kept? It would be easier to render both cases the same.

pabzm:
FTR: This won't receive any updates in the near future due to a focus shift and resource exhaustion.

Issue #9762, opened 2025-02-02
[Feature request] Option to deactivate automatic hyperlinks in email composition


MiMoHo:
### Prerequisites

- [x] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed feature requests
- [x] I have [searched](https://plugins.roundcube.net/) for plugins that provide already provide the feature

### Proposal

There is currently no settings option for the user in "Composing Messages" to deactivate automatic transformation of URLs to clickable hyperlinks.

### Motivation and context

Hyperlinks bare the risk of phishing, tracking or misleading redirection which is intransparent to the receiver. Therefore, I wish to send URLs as plain text in HTML messages. However, they always convert to hyperlinks automatically which I have to remove manually each time. In order to highlight the URL and indicate where it ends, I give it a different color and two spaces after. However, putting in a space character after the URL triggers hyperlink conversion again.
As a webmail user, I cannot make changes to the config like probably many other Roundcube users. Hence, I suggest a settings option to deactivate automated hyperlinking.

alecpl:
This is the 'autolink' plugin of TinyMCE. It shouldn't be hard to implement such an option. A pull request would be welcome.

alecpl:
Implemented.

PR #9818, opened 2025-03-19
add enable_autolink option for tinymce editor

johndoh:
for #9762

adds a new section called "editor options" in the Composing Messages settings to allow a user to toggle the `autolink` used in TinyMCE


program/localization/en_US/labels.inc
```diff
@@ -631,6 +631,8 @@ $labels['spellcheckignoresyms'] = 'Ignore words with symbols';
 $labels['spellcheckignorenums'] = 'Ignore words with numbers';
 $labels['spellcheckignorecaps'] = 'Ignore words with all letters capitalized';
 $labels['addtodict'] = 'Add to dictionary';
+$labels['editoroptions'] = 'Editor Options';
+$labels['enableautolink'] = 'Automatically convert text to links in HTML editor';
```
alecpl, line 635: I'm not sure the wording is optimal. I don't think we have to mention HTML editor here, also I would probably use word hyperlink.

How about: "Automatically convert valid URLs into hyperlinks".

Issue #8403, opened 2022-01-17
1.5.x  Elastic skin resets layout view for email previews


MASHtm:
This was already addressed for 1.4.x in  #6613. I left a comment there for 1.5.x, but it seems it was not noticed.

Coming from Skin Larry with "Desktop" layout...

If I switch to elastic skin (eg. with elastic4mobile plugin) and go back to larry afterwards the layout is permanently switched from "Desktop" to "Widescreen".

alecpl:
The layout is re-set whenever you visit Mail interface, and it is not a per-skin setting. In fact, we do not have per-skin settings. So, I don't see an easy fix, and as it is imho not a big issue I propose a wontfix.

buttercookie42:
What was wrong with the fix in 1.4.x?
If a skin doesn't support a certain setting, it should use some sort of sensible fallback, but it shouldn't modify the permanent settings unless the user actually goes into the settings menu and confirms the settings, shouldn't it?

MASHtm:
it is a big issue in combination with elastic4mobile. And it worked and still works in 1.4 after it was fixed somehow. Closing it with "wontfix" is not very satisfying.

kochs-online:
> it is a big issue in combination with elastic4mobile. And it worked and still works in 1.4 after it was fixed somehow. Closing it with "wontfix" is not very satisfying.

I agree. Switching between layouts and losing those settings every time is really annoying.

Why not ignore (but keep) settings that are unused by a certain skin?

PR #9822, opened 2025-03-21
Don't save layout if in dont_override.

iazz:
This simple change ensures that the layout will not be overwritten to user preferences if `layout` is in `dont_override`.

A plugin can thus manipulate the config by adding a hook to `config_get` in order to make the setting of `layout` temporary.  A prime example of a plugin whom this would come in handy is `elastic4mobile`, which currently overwrites the layout setting, as reported in #8403.  A PR for that plugin is coming shortly.

### DIFF
--- a/program/actions/mail/list.php
+++ b/program/actions/mail/list.php
@@ -62,8 +62,9 @@ class rcmail_action_mail_list extends rcmail_action_mail_index
         // register layout change
         if ($layout && preg_match('/^[a-zA-Z_-]+$/', $layout)) {
             $rcmail->output->set_env('layout', $layout);
-            if (!in_array('layout', $dont_override))
+            if (!in_array('layout', $dont_override)) {
                 $save_arr['layout'] = $layout;
+            }
             // force header replace on layout change
             if (!empty($_SESSION['list_attrib']['columns'])) {
                 $cols = $_SESSION['list_attrib']['columns'];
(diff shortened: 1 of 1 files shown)


program/actions/mail/list.php
```diff
@@ -62,7 +62,8 @@ public function run($args = [])
         // register layout change
         if ($layout && preg_match('/^[a-zA-Z_-]+$/', $layout)) {
             $rcmail->output->set_env('layout', $layout);
-            $save_arr['layout'] = $layout;
+            if (!in_array('layout', $dont_override))
+                $save_arr['layout'] = $layout;
             // force header replace on layout change
```
alecpl, line 67: Please add `{ }`. Also, what about the other lines of code around? Should `set_env()` still be invoked? Should `$cols` be set?

iazz: The initial issue that I'm trying to fix here is that the `elastic4mobile` plugin resets the layout in user preferences every time the user accesses their account from a mobile device.  If a user normally uses, say, the Larry skin with the `list` layout, as soon as they access the webmail using a mobile browser, the temporarily used Elastic skin resets the layout to `widescreen` in a permanent fashion.  When the user logs back using a desktop browser, the layout is `widescreen` and has to be changed in preferences back to `list`.

The thing is that `elastic4mobile` is a hack that temporarily overrides the skin to Elastic when the requesting browser is in mobile mode.  Further down the road, the currently used skin can change the layout if the current layout is not actually supported by the skin (which is the case with Elastic, which supports only `widescreen`).  All the rest seems to work just right, (calls to `set_env()`, setting of `$cols`, and whatnot) except that the `layout` setting is saved to the user's preferences.

So it seems reasonable to leave everything else as is and only prevent overwriting of the `layout` preference when it is found in `dont_override`.  On the plugin side, `elastic4mobile` must simply add `layout` to `dont_override` in a `config_set` hook, and do this for every request (including AJAX), so that the `layout` preference doesn't update the saved user preferences.

PR #9823, opened 2025-03-24
Document the current release management

pabzm:
This documents our release management in short terms, so people can learn what to expect.


pabzm:
The failing CI workflow isn't related, I'm not sure why it occurs here (locally it gives me the same errors for the current "master" branch, but in the CI it apparently doesn't).

PR #9824, opened 2025-03-24
Run phpstan also on public_html

pabzm:
Now that public_html isn't just a symlink anymore but contains actual code, that code should be checked, too.


pabzm:
@mvorisek This might interest you as well (I can't add you as a reviewer for some reason).

alecpl:
I'll fix these new phpstan errors separately.

mvorisek:
> @mvorisek This might interest you as well (I can't add you as a reviewer for some reason).

I have the most strict digital denoising configuration :), especially because of bots.

+1 for this PR, it was a symlink fix only...

PR #9828, opened 2025-03-26
Add APCu-Cache

Hermsi1337:
Since `APC` is (very) outdated in favor of `APCu` and as of PHP `>= 8.0` you can no longer `pecl install apcu_bc` `roundcube` should also support `APCu` in addition to `APC`.

One could also argue that the `APC`-Caching-Class should be removed in future releases of `roundcube` but I think this discussion is out of scope for this PR.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -240,7 +240,7 @@ $config['imap_disabled_caps'] = [];
 // This is used to relate IMAP session with Roundcube user sessions
 $config['imap_log_session'] = false;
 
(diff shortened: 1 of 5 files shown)


alecpl:
Yeah, we should remove the `apc` driver. This was kinda "because we can". I think almost no one was using it. I wonder who will use `apcu`.

Hermsi1337:
> Yeah, we should remove the `apc` driver. This was kinda "because we can". I think almost no one was using it. I wonder who will use `apcu`.

Apparently I will 😃

Issue #9329, opened 2024-01-31
Very hard to check the email address we are about to send to


jidanni:
Here we are composing an email. As we see we're sending it to cellmapper. But we would just like to have a little quick check to see if it's the  cellmapper support address or the other cellmaper  address.

It is very hard at this point to find what address exactly we're about to send it to.

All we can do is tap it and it'll just disappear and then we're  more trouble.

The best thing to do would be just to send it and then look in the sent box to check the details. And if we don't like it then make a second message.

So I'm saying there should be a more flexible way to see the full address that we're about to send to.
![Screenshot_20240131_172549_Chrome Beta.jpg](https://github.com/roundcube/roundcubemail/assets/1230959/7a9646c7-d157-4de1-b20b-0ad32a075003)

alecpl:
> All we can do is tap it and it'll just disappear and then we're more trouble.

That sounds like a bug. You'd need to provide more info. What browser/device? What Roundcube version?

A double-tap on that box should display an edit form where you could see the full name and email address. Works for me in my iPhone. That being said, it should probably do the same on a single tap.

jidanni:
Indeed, One tap does nothing, so we wouldn't go on to try two.

Issue #9219, opened 2023-11-16
FR: show recipient's e-mail address


YZgitter:
I want to see the **actual full e-mail address** of the recipient and not the stupid "user friendly name" or whatever it's called.
And I don't want to hover over with a mouse to see it, it should be displayed there all the time.

Please, for the love of god, add a setting for this.

alecpl:
There is an option for this, but it does not apply to the messages list. The messages list is not always wide enough to display both name and address, so I suppose it would have to display only the address.

stef204:
this feature would be a very welcome addition.

> Seeing the actual address there can prevent important mistakes

LevitatingBusinessMan:
I agree that this feature is missing, which is why I made the userscript linked to by @stef204.

My contacts often have multiple addresses and without this little hack I am continuously paranoid of sending to the wrong email.

PR #9832, opened 2025-03-31
move autocomplete list rendering to client side

johndoh:
Moving the rendering of the autocomplete address list used in compose to the client side allows skins to handle it and offer different layouts for different sized screens. Before I spend too much time on this I wanted to submit this draft to see if the idea was worth pursuing.

There are some hurdles. There is the `contact_search_name` config var which has to be respected. The server side needs to know about it so it can return the correct fields in the request but rendering those is a client side task because of screen size. Also `contact_search_name` does not support HTML and I think that's correct, it could be abused. May be having a configurable display for normal screens and a fixed one for small screens is enough.

Example of how the new small screen layout looks:
compose

The proposed `rcube_webmail.ksearch_results_display()` and `rcube_webmail.ksearch_results_highlight()` JS functions can be overridden by skins/plugins if required but the stock version maintain the current functionality plus a new layout for smaller screens.

I think ultimately all the text rendering code should be removed from `rcube_addressbook::compose_search_name()` and instead only the fields returned with the rendering moved into `app.js`.

This might help with #9329 and #9219.

### DIFF
--- a/plugins/acl/acl.php
+++ b/plugins/acl/acl.php
@@ -104,7 +104,7 @@ class acl extends rcube_plugin
                     $fields = rcube_addressbook::compose_search_fields($record);
                     $user = ['name' => $user, 'fields' => $fields];
                     $users[] = $user;
-                    $keys[] = $fields['name'] ?: $user['name'];
+                    $keys[] = isset($fields['name']) ? $fields['name'] : $user['name'];
                 }
             }
 
--- a/plugins/acl/acl.php
+++ b/plugins/acl/acl.php
@@ -104,7 +104,7 @@ class acl extends rcube_plugin
                     $fields = rcube_addressbook::compose_search_fields($record);
                     $user = ['name' => $user, 'fields' => $fields];
                     $users[] = $user;
-                    $keys[] = isset($fields['name']) ? $fields['name'] : $user['name'];
+                    $keys[] = $fields['name'] ?? $user['name'];
                 }
             }
 
--- a/plugins/acl/acl.php
+++ b/plugins/acl/acl.php
@@ -101,7 +101,7 @@ class acl extends rcube_plugin
                 }
 
                 if ($user) {
-                    $fields = rcube_addressbook::compose_autocomplete_fields($record);
+                    $fields = rcube_addressbook::compose_search_fields($record);
                     $user = ['name' => $user, 'fields' => $fields];
                     $users[] = $user;
                     $keys[] = $fields['name'] ?: $user['name'];
--- a/plugins/acl/acl.php
+++ b/plugins/acl/acl.php
@@ -101,10 +101,10 @@ class acl extends rcube_plugin
                 }
 
                 if ($user) {
-                    $display = rcube_addressbook::compose_search_name($record);
-                    $user = ['name' => $user, 'display' => $display];
+                    $fields = rcube_addressbook::compose_autocomplete_fields($record);
+                    $user = ['name' => $user, 'fields' => $fields];
                     $users[] = $user;
-                    $keys[] = $display ?: $user['name'];
+                    $keys[] = $fields['name'] ?: $user['name'];
                 }
             }
 
@@ -118,7 +118,7 @@ class acl extends rcube_plugin
                     $group_id = is_array($record[$group_field]) ? $record[$group_field][0] : $record[$group_field];
 
                     if ($group) {
-                        $users[] = ['name' => ($prefix ?: '') . $group_id, 'display' => $group, 'type' => 'group'];
+                        $users[] = ['name' => ($prefix ?: '') . $group_id, 'fields' => ['name' => $group], 'type' => 'group'];
                         $keys[] = $group;
                     }
                 }
--- a/program/lib/Roundcube/rcube_addressbook.php
+++ b/program/lib/Roundcube/rcube_addressbook.php
@@ -754,7 +754,7 @@ abstract class rcube_addressbook
      * @param array  $contact Hash array with contact data as key-value pairs
      * @param string $email   Optional email address
      * @param string $name    Optional name (self::compose_list_name() result)
-     * @param string $templ   Optional template to use (defaults to the 'contact_search_name' config option)
+     * @param string $templ   Optional template to use (defaults to '{name} <{email}>')
      *
      * @return string Display name
      */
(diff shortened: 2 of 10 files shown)


alecpl:
We should consider removing `contact_search_name` option. Maybe we can replace it with `autocomplete_extra_field` or something. This way the result would contain at most 3 items: "email", "display" (contact name), "extra" (e.g. department name). The extra text could be displayed not inline, but maybe aligned to the right, and only if there's room.

We can consider displaying "two line" records on a big screen too. Gmail does that.

johndoh:
I like the idea of scrapping `contact_search_name`. The `autocomplete_extra_field`, I think it would make sense if that could only be 0 or 1 fields and then skins have a fixed layout to support. If devs want more fields or more customisation then they could do it via the plugin API because its not only about the fields but also the layout. Does that sound acceptable?

Do we really need `autocomplete_extra_field` even, could `addressbook_name_listing` be used/improved so that if you wanted say `firstname lastname (department)` its shown the same in the autocomplete and the contact list? Then in autocomplete the email address would be on the second line and that's the only difference with the contacts list. It could use similar logic as that currently used for displaying `contact_search_name` so no existing configurability would be lost.

Off the top of my head I'm thinking a plugin hook in `rcube_addressbook::compose_search_name()` for the returned fields/values and then use the proposed (or similar) new JS functions to construct the entry.

alecpl:
Makes sense.

johndoh:
I have replaced `contact_search_name` with `contactlist_name_template` which applies in the contact list and the autocomplete, anywhere `addressbook_name_listing` is used. I did not change `addressbook_name_listing` because of its use in the settings screen.

eef8919 I modified the ACL plugin to use the new autocomplete but I do not have a working LDAP setup so I cannot test this change, please could someone who uses this plugin confirm the autocomplete still works?

plugins/acl/acl.php
```diff
@@ -101,10 +101,10 @@ public function acl_autocomplete()
                 }
 
                 if ($user) {
-                    $display = rcube_addressbook::compose_search_name($record);
-                    $user = ['name' => $user, 'display' => $display];
+                    $fields = rcube_addressbook::compose_autocomplete_fields($record);
+                    $user = ['name' => $user, 'fields' => $fields];
                     $users[] = $user;
-                    $keys[] = $display ?: $user['name'];
+                    $keys[] = $fields['name'] ?: $user['name'];
```
alecpl, line 107: There's no guarantee name is set.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -763,55 +773,84 @@ public static function compose_search_name($contact, $email = null, $name = null
         static $template;
```
alecpl, line 773: We don't need this static anymore. It was useful with `config->get()` as this can get slow, but you removed that call.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -763,55 +773,84 @@ public static function compose_search_name($contact, $email = null, $name = null
...
+     * @param array  $contact Hash array with contact data as key-value pairs
+     * @param string $email   Optional email address
+     * @param string $name    Optional name (self::compose_list_name() result)
+     * @param array  $fields  Optional fields to return (defaults to ['name', 'email'])
+     *
+     * @return array Fields
+     */
+    public static function compose_autocomplete_fields($contact, $email = null, $name = null, $fields = null)
+    {
+        static $template;
```
alecpl, line 809: This static is not needed anymore.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -763,55 +773,84 @@ public static function compose_search_name($contact, $email = null, $name = null
...
+     * Build contact display name for autocomplete listing
+     *
+     * @param array  $contact Hash array with contact data as key-value pairs
+     * @param string $email   Optional email address
+     * @param string $name    Optional name (self::compose_list_name() result)
+     * @param array  $fields  Optional fields to return (defaults to ['name', 'email'])
+     *
+     * @return array Fields
+     */
+    public static function compose_autocomplete_fields($contact, $email = null, $name = null, $fields = null)
```
alecpl, line 807: Should it be rather called `compose_search_fields()`? Also the description is wrong.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -763,55 +773,84 @@ public static function compose_search_name($contact, $email = null, $name = null
...
+
+        if (!isset($template)) {  // cache this
+            $template = ['name', 'email'];
+        }
 
-                if (empty($value)) {
-                    $value = strpos($key, ':') ? $contact[$key] : self::get_col_values($key, $contact, true);
-                    if (is_array($value) && isset($value[0])) {
-                        $value = $value[0];
+        $fields = $fields ?: $template;
```
alecpl, line 815: You don't need `$template` variable at all. Just set `$fields` to the default here.

program/lib/Roundcube/rcube_addressbook.php
```diff
@@ -763,55 +773,84 @@ public static function compose_search_name($contact, $email = null, $name = null
...
-                        $value = $email;
-                        break;
-                }
+    /**
+     * Build contact display name for autocomplete listing
+     *
+     * @param array  $contact Hash array with contact data as key-value pairs
+     * @param string $email   Optional email address
+     * @param string $name    Optional name (self::compose_list_name() result)
+     * @param array  $fields  Optional fields to return (defaults to ['name', 'email'])
```
alecpl, line 803: Either allow null for $fields or not. It's incompatible with the function argument definition below. Same for `$name` and `$email`.

program/lib/Roundcube/rcube_contacts.php
```diff
@@ -347,7 +347,16 @@ public function search($fields, $value, $mode = 0, $select = true, $nocount = fa
             foreach ($words as $word) {
                 $groups = [];
                 foreach ((array) $fields as $idx => $col) {
-                    $groups[] = $this->fulltext_sql_where($word, $mode, $col);
+                    // table column
```
alecpl, line 350: This change does not seem to be related. Could be moved to a separate PR, with some explanation.

alecpl (changes requested):

program/lib/Roundcube/rcube_addressbook.php
johndoh: I dont understand

pabzm:
@johndoh Would you be interested to put some more effort into this in the next days? We thought about to maybe get this into v1.7, if possible. But please don't feel pressured, we only want to explore this possibility.

johndoh:
I'm happy to work on it. Is there any particular changes you are looking for?

I already made the changes Alec requested and I think 3 failing tests are not related to this PR

pabzm:
@johndoh Great, thanks!

@alecpl Could you have a another look?

pabzm:
@johndoh Maybe you could rebase onto the lastest of the "master" branch, it has this failing E2E-test fixed.

alecpl:
There's some bug here. If you have a contact that matches search string, but does not have any name, it will not be displayed (it's not returned by the search request).

johndoh:
I accidently reverted #9850 probably when I split it into its own PR.

alecpl:
Now I see that for a contact without a name we display the email twice. I.e. we use email in the `name` field. I think this is wrong. We should have the name field empty. And maybe such contact should display in only one line, as groups are.

johndoh:
> Now I see that for a contact without a name we display the email twice. I.e. we use email in the name field. I think this is wrong. We should have the name field empty. And maybe such contact should display in only one line, as groups are.

fixed

PR #9837, opened 2025-04-08
Can use variable in rcmail.gettext, like the php version

Rotomeca:
In PHP, you can use an array as an argument for `rcmail:get_instance()->gettext()` to utilize variables in localization texts. This modification allows the JavaScript version to do the same.

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -1656,22 +1656,22 @@ function rcube_webmail() {
     };
 
     // return a localized string
-    this.get_label = function (name, domain, variables = null) {
-        if (domain && this.labels[domain + '.' + name]) {
-            name = this.labels[domain + '.' + name];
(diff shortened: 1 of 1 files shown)


program/js/app.js
```diff
@@ -1656,13 +1656,19 @@ function rcube_webmail() {
     };
 
     // return a localized string
-    this.get_label = function (name, domain) {
+    this.get_label = function (name, domain, variables = null) {
         if (domain && this.labels[domain + '.' + name]) {
-            return this.labels[domain + '.' + name];
+            name = this.labels[domain + '.' + name];
```
alecpl, line 1661: At this point it's not a name anymore. I'd suggest we rename the `name` argument to `label`.

Rotomeca: Updated

Issue #7230, opened 2020-02-20
1.4.3 with elastic skin - SHIFT+PGUP and SHIFT+PGDOWN don't work


chilek:
... in composer window (see attachment).

![Zrzut ekranu z 2020-02-20 11-04-48](https://user-images.githubusercontent.com/1457437/74923320-daf4ca00-53d0-11ea-8eca-f7c95606f74c.png)

chilek:
... at the same time it works with larry skin.

alecpl:
You mean for text selection? I confirm it works in Firefox, but not in Chromium.

chilek:
> You mean for text selection?

Exactly.
In my FF browser also works, but not in all chromium-based browsers (probably).

alecpl:
Strange, but it looks like `overflow-y:hidden` makes Chrome to "disable" these key combinations. Also PageUp/PageDown without Shift do not work. You can still use Shift+ArrowUp/ArrowDown and other text selection shortcuts.

We have to look at other ways to implement textarea auto-resizing.

alecpl:
This is not fixed in git-master, but I'll have to do some more tests before it could be beckported to 1.4.

chilek:
cherry-picked https://github.com/roundcube/roundcubemail/commit/38d6659384712e4c3f4d6c9b041d9e9e4ef7eee0 to my installation. Now it works in my Google Chrome.

alecpl:
There's scroll-jump effect in MS Edge now, we have to fix that.

alecpl:
Fixed. Backported to 1.4.

chilek:
Dzięki!

PR #9838, opened 2025-04-09
Allow to resize the compose textarea

pabzm:
Sometimes one needs a little bit more space to write an email, then being able to manually draw the edit field bigger is very helpful. 

This had been blocked in https://github.com/roundcube/roundcubemail/commit/1c5f83d41cccefa3e54ae2c752a534fe51136a64, which references https://github.com/roundcube/roundcubemail/issues/7230. But the problem described in that issue doesn't reappear without this CSS statement (tested in Firefox, Chrome and Vivaldi on Mac OS), so I assume we can safely remove it.

### DIFF
--- a/skins/elastic/styles/widgets/editor.less
+++ b/skins/elastic/styles/widgets/editor.less
@@ -475,7 +475,6 @@ div.tox {
         font-size: 13px;
         width: 100% !important;
         padding-top: 40px;
-        resize: none;
     }
 
     & > iframe { // e.g. mailvelope frame


alecpl:
The textarea (and html editor) grows while you write in it. So, I don't see a reason to resize it manually.

pabzm:
But there's neither a reason to explicitly block it, is there?

I prefer to see a bigger textarea for longer texts, and having to enter a lot of newlines to achieve that feels like a clumsy workaround.

ChristophWurst (approved):
Makes sense to bring this back if the previous problems no longer show :+1:

PR #9846, opened 2025-04-17
Update ar_SA.inc

soufien88y:


alecpl:
We manage localization via https://app.transifex.com/roundcube/roundcube-webmail/ and then once in a while fetch all updates for all languages. So, I'll merge it, but if Transifex source is not updated these will be lost on the next update.

Issue #9843, opened 2025-04-16
Zoom into attachment image: wrong scaling


codegain:
### Prerequisites

- [x] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I can recreate the issue with all plugins disabled

### Describe the issue

Hi,

- Have a large image with f.e. 1000x1000 pixels as an attachment
- Click on "Show" on attachment in the email
- Wait for the popup to load
- Zoom in using the button "Zoom in".
- The top and left side of the image gets cropped the further you zoom in
- The right and bottom side of the image is unaffected

Testimage used:
https://github.com/user-attachments/assets/e968e98b-2967-42d8-ae00-fd603d096429

No zoom applied:
![Image](https://github.com/user-attachments/assets/d887ffce-05b0-425e-ba74-fecd4abc0022)

Some zoom applied:
![Image](https://github.com/user-attachments/assets/8b3b6201-d1cf-495c-b966-3ba1a7c03071)

(Scrollbars won't go any further left + up)

My guess is that this behavior could be fixed with using `transform-origin`, but I don't know if that should be the desired behavior.
Maybe related to #7669

### What browser(s) are you seeing the problem on?

Chrome

### What version of PHP are you using?

v8.1

### What version of Roundcube are you using?

1.6.10

alecpl:
I confirm the issue. I can't make it working (and keep the image centered at the same time). Any help appreciated.

listumps:
This can be fixed by adding "_transform-origin: 0 0;_" to the inline CSS style of the image being shown. It should be an easy fix.

alecpl:
I tried that, but if you zoom out you'll see that the image is not centered anymore.

alecpl:
Oh, and it breaks image rotation feature.

listumps:
Seems like a bigger problem than what it initially seemed like:
[https://github.com/w3c/csswg-drafts/issues/9458](url)

listumps:
Just an option, though with some glitches (depending on if we use width: %percentage% or height: %percentage%):

_On page load:_
Inline CSS style of the image's parent (<body>): "`margin: 0px; height: 100%;`"
Inline CSS style of the image: "`max-width: 100%; max-height: 100%; margin: 0 auto; display: block;`"

_On click for zoom:_
Inline CSS style of the image: "`margin: 0 auto; display: block; width: %percentage%;`"

[...]

Issue #7669, opened 2020-10-11
Image viewer: you can't see the top of zoomed images


jgmy:
Hi!
I receive many, many student tasks in pdf or jpg format. I have to use the zoom feature of the image viewer. But, when you zoom, the top part of images (I'm speaking about 9:16 images produced by mobile phones) is cut. You can not see them by scrolling up.

Roundcube 1.4.7

Tested on chrome 86 and (chromium) edge.

![2020-10-11](https://user-images.githubusercontent.com/2573642/95673984-7c719b80-0bad-11eb-9556-b9fdfb0bdd5a.png)

alecpl:
I see the scrollbar, you mean despite it's existence the scrolling does not work? Works for me in Chromium 85.

jgmy:
I can scroll down but I can't scroll up. Scroll bar is at top position, as
seen in the picture.

jgmy:
Its not critical sice you can right-click the image and open it in the browser as new tab. Just wanted you to know/verify.

alecpl:
I get it now, but I'm still unable to reproduce with Chromium 85 and Roundcube git-master. When I have 86 I'll test again using Roundcube 1.4.

alecpl:
I'm unable to reproduce on Chromium 86 and most recent Roundcube version.

jgmy:
Thanks for your efforts, Alexandr. I guess the admin of the site I'm using
has some misconfiguration.
I will contact it.

PR #9849, opened 2025-04-21
Rework zoom and rotate of attached images

listumps:
Addresses #9843 and possibly #7669

### DIFF
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -457,13 +457,15 @@ function rcube_webmail() {
 
                             // do not apply styles to an error page (with no image)
                             if (contents.find('img').length) {
-                                contents.find('img').css({ maxWidth: '100%', maxHeight: '100%' });
+                                contents.find('img').css({ 
+                                    maxWidth: '100%',
+                                    maxHeight: '100%',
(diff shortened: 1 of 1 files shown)


alecpl:
I tested it. It does not work correctly when using zoom-in and zoom-out. I tested with an image of 1000pxx1000px.

I have no time to work on this now, but I'm guessing that it might be a problem that you use current width/height in calculations not the original width/height.

listumps:
Confirmed. It's working great with latest Google Chrome on desktop and mobile but has a strange behavior in Firefox.
The PR is updated accordingly and now it works in both Google Chrome and Firefox.

listumps:
Fixed.
Had thoroughly tested it with different size images within Google Chrome and Firefox, both desktop and mobile. No issues so far.

listumps:
Tested thoroughly with:
- Latest Google Chrome and Firefox
- Mobile and desktop
- Different sizes of images
No issues found so far.

listumps:
Test environment:
- Latest Google Chrome, Firefox and Edge
- Different sizes of images
- Mobile and desktop version
No issues so far.

PR #9859, opened 2025-05-14
Update defaults.inc.php

Giorgio00796:


iasdeoupxe:
Looks like an arbitrary change and debug logging should IMHO not enabled by default.

PR #9866, opened 2025-05-23
Validate URL parameter in upload code

pabzm:
Backport of https://github.com/roundcube/roundcubemail/pull/9865


pabzm:
The browser tests failures are probably unrelated, they stop with "disk I/O error"

PR #9867, opened 2025-06-01
Support CIDR notation for proxy whitelist (IPv4&IPv6)

c4539:


alecpl:
How about we use some existing package, e.g. `wikimedia/ip-utils`?

Neustradamus:
@c4539: Any progress on it?

PR #9891, opened 2025-06-13
Add a passwdqc-like driver for password checks

marcocen:
I had the need to check if passwords would pass a passwdqc_pam check before changing them via the roundcube interface and found that the only drivers available were zxcvb and pwned.

I whipped up a quick driver to do those checks and after some polishing up decided that it may be of some use to someone else, so here it is.

I only have localizations for spanish and english, as those are the only languages I feel comfortable in.


alecpl:
Coding style is not right. Anyway, why not just use `pwqcheck` binary?

Neustradamus:
@marcocen: Any progress on it?

marcocen:
Yes, I have some new commits to use the `pwqcheck` binary if present, I might be able to push it before the weekend.

Issue #9905, opened 2025-06-23
Exporting as .eml, the filename is not sanitised correctly


oscardimanno:
### Prerequisites

- [x] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I can recreate the issue with all plugins disabled

### Describe the issue

Hello!
When I download an email containing a question mark in the subject (but not at the end of the string), e.g. "test ? test", Chrome seems to fail to sanitise the subject and downloads it as "download.txt".
Also tried on Firefox and seems to work

**How to replicate:**
1. Send a message with the subject "some words? some other words".
2. Go to the Sent folder.
3. Export the email.
4. Et voilà! You've got the bug!

**Screenshots:**

![Image](https://github.com/user-attachments/assets/5e6150ad-eb72-4472-b85e-30f06c2ccd86)
![Image](https://github.com/user-attachments/assets/deae9798-49aa-4d71-85dd-fbd1a3aac9d4)

### What browser(s) are you seeing the problem on?

Chrome

### What version of PHP are you using?

v8.2-fpm

### What version of Roundcube are you using?

v.1.6.4 and also on master branch

alecpl:
Fixed via #9960.

PR #9906, opened 2025-06-23
fix: Sanitize filename on download and correction of filename header

oscardimanno:
This PR fixes issue #9905 .

Currently, Roundcube chooses whether to use the filename or filename* parameter based on the characters in the filename.

This PR simplifies the logic behind this.
According to [RFC 6266, section 4.3](https://datatracker.ietf.org/doc/html/rfc6266#section-4.3), and the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#syntax), the user agent prioritises the 'filename*' parameter if it is present and ignores the 'filename' parameter.
This change aligns the code with the IETF standard while maintaining compatibility with older clients.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -1650,7 +1650,7 @@ class rcmail_output_html extends rcmail_output
                     $value = implode(', ', $value);
                 }
 
-                return html::quote($value);
+                return html::quote((string) $value);
             case 'form':
                 return $this->form_tag($attrib);
         }


Neustradamus:
@alecpl, @pabzm: If the @oscardimanno respects IETF standard, it is good?

alecpl:
I think using both parameters is fine and the way to go. Please, fix tests.

oscardimanno:
Hi @alecpl, thanks.
I'll start working on fixing the tests and I will update the PR once they're good.

program/lib/Roundcube/rcube_output.php
```diff
@@ -256,17 +256,15 @@ public function download_headers($filename, $params = [])
...
+            $fallback_filename = preg_replace('/[^a-zA-Z0-9_.(),;@+ -]/', '_', $filename);
+            $disposition .= "; filename=\"{$fallback_filename}\"";
 
-                $disposition .= "; filename*={$charset}''{$filename}";
+            $filename = rawurlencode($filename);
+            $charset = $this->charset;
+            if (!empty($params['charset']) && rcube_charset::is_valid($params['charset'])) {
+                $charset = $params['charset'];
             }
+            $disposition .= "; filename*={$charset}''{$filename}";
```
alecpl, line 267: Maybe it would make sense to add this parameter only when `$fallback_filename != $filename`.

oscardimanno:
Hi @alecpl,
closing this PR.
I've cleaned up my master branch and moved the changes on a dedicated branch to resolve a CI issue.
The new, updated pull request is here: #9960

PR #9908, opened 2025-06-23
Compose in markdown

pabzm:
This adds a markdown editor that sends HTML to the server.

TODO:
- [ ] Provide all libraries from our server.
- [ ] Allow users to disable the editor.
- [ ] Fix cooperation with Mailvelope (which currently is disabled)
- [ ] I18n (for labels and browser-spellchecking)
- [ ] Handle spellcheck-button.
- [ ] Maybe send also the plaintext/markdown to the server, so that gets included in the outgoing email and not the stripped-down plaintext-from-HTML.

### DIFF
--- a/plugins/markdown_editor/markdown_editor.js
+++ b/plugins/markdown_editor/markdown_editor.js
@@ -1,0 +1,203 @@
+class MarkdownEditor {
+    // State holder for debouncing.
+    _saveTimer = null;
+    id = 'markdown_editor';
+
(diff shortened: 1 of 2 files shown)


Neustradamus:
@pabzm: Nice, more and more people would like MarkDown!

alecpl:
Superceded by #9966, right?

pabzm:
Right!

PR #9914, opened 2025-06-27
feat: :sparkles: Allow saving of compose_data somewhere other than in the session

Rotomeca:
We are experiencing issues due to saving compose_data in the session when there are a large number of users.
We’ve therefore found a solution by saving compose_data in the cache instead.
We’re submitting this idea to you

### DIFF
--- a/program/actions/mail/compose.php
+++ b/program/actions/mail/compose.php
@@ -42,8 +42,8 @@ class rcmail_action_mail_compose extends rcmail_action_mail_index
         self::$COMPOSE_ID = rcube_utils::get_input_string('_id', rcube_utils::INPUT_GET);
         self::$COMPOSE = null;
 
-        if (self::$COMPOSE_ID && !empty($_SESSION['compose_data_' . self::$COMPOSE_ID])) {
-            self::$COMPOSE = &$_SESSION['compose_data_' . self::$COMPOSE_ID];
+        if (self::$COMPOSE_ID && !empty(self::get_compose_data(self::$COMPOSE_ID))) {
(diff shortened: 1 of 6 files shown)


alecpl:
The idea to slim down the session is not new, and it makes sense. But could you explain what problems did you encounter specifically?

Also, I think we should consider just using a separate DB table. Use of redis/memcache for this purpose seems unnecessary.

program/actions/mail/compose.php
```diff
@@ -336,6 +335,8 @@ public function run($args = [])
 
         self::spellchecker_init();
 
+        self::set_compose_data(self::$COMPOSE_ID, self::$COMPOSE);
```
alecpl, line 338: Saving the data here may not be enough. Also, I see the data is used in `rcmail_action_mail_attachment_upload::init()`. Also, in `rcmail_action_mail_send` you have to save the data (in case of for example draft save action).

Rotomeca:
> The idea to slim down the session is not new, and it makes sense. But could you explain what problems did you encounter specifically?
> 
> Also, I think we should consider just using a separate DB table. Use of redis/memcache for this purpose seems unnecessary.

Randomly, we have users with "Invalid Compose ID" in the logs, and they are unable to send the email. When I debug, in compose.php, the data is present all the way to the end of the script. But in send.php, the data is no longer there. When we check in Redis, we see that compose_data is null. We don't really know why.

I'm okay with just putting it in a table, I suggested the other options for consistency and to leave a choice and if someone would'nt load the database.

program/actions/mail/compose.php
Rotomeca: Okay, I check this, thanks :D

PR #9922, opened 2025-07-04
[FIX] TAB display @ text-only messages

xerc:
pre/ this is a proof-of-concept - will clean up the source if accepted

ref. https://developer.mozilla.org/en-US/docs/Web/CSS/white-space
- no need for `<br>`
- no need for `nobr`
- no random `&nbsp;`

& `\t` "logic" works autom. as it should -> always correct identation

### DIFF
--- a/program/lib/Roundcube/rcube_text2html.php
+++ b/program/lib/Roundcube/rcube_text2html.php
@@ -31,15 +31,13 @@ class rcube_text2html
     /** @var array Configuration */
     protected $config = [
         // non-breaking space
-        'space' => "\xC2\xA0",
+        'space' => '&nbsp;',
         // enables format=flowed parser
         'flowed' => false,
         // enables delsp=yes parser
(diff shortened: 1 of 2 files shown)


alecpl:
We could give it a try, I suppose, but previewing messages is not everything. What if you forward (inline) such a message in HTML mode?

xerc:
it look bad if you have "all new mails have to be HTML" active - but thats the worst option possible
https://github.com/roundcube/roundcubemail/blob/eff91a93caf660a53ccdf582e1af05ea784b4b59/program/localization/de_DE/labels.inc#L485
https://github.com/roundcube/roundcubemail/blob/eff91a93caf660a53ccdf582e1af05ea784b4b59/program/localization/de_DE/labels.inc#L515

xerc:
btw/ if forwarded as TEXT it still replaces the TABs to SPACEs while sending (preview is correct)

PR #9926, opened 2025-07-10
Allow to manually reorder "normal" folders

pabzm:
Protected folders like INBOX, Sent, etc. are staying at the top of the list.

Fixes #5003

TODO:
- [x] Fix broken state after folder renaming
- [x] Fix the styling (equalize white space around list elements through padding, avoid ID as selector)
- [x] Figure out if the item was moved between protected folders, which is not allowed.
- [x] ~~Save only if the folder list is different than at start~~ (Not worth the effort)
- [ ] In the receive callback, can we wait for the confirmation dialog without introducing async/await and Promises?

Screenshots:

Screenshot 2025-07-25 at 16 24 28

---------

Screenshot 2025-07-25 at 16 24 37

---------

Screenshot 2025-07-25 at 16 24 49

---------

Screenshot 2025-07-25 at 16 24 57

### DIFF
--- a/program/actions/settings/folder_edit.php
+++ b/program/actions/settings/folder_edit.php
@@ -156,6 +156,13 @@ class rcmail_action_settings_folder_edit extends rcmail_action_settings_folders
                 'label' => $rcmail->gettext('parentfolder'),
                 'value' => $select->show($selected),
             ];
+
+            $up = new html_button(['id' => 'move-folder-up', 'class' => 'move-folder-up'])->show($rcmail->gettext('reorder_folder_up'));
+            $down = new html_button(['id' => 'move-folder-down', 'class' => 'move-folder-down'])->show($rcmail->gettext('reorder_folder_down'));
+            $form['props']['fieldsets']['location']['content']['order'] = [
+                'label' => $rcmail->gettext('reorder_folder'),
+                'value' => html::div([], [$up, $down]),
+            ];
         }
 
         // Settings
--- a/program/actions/settings/folder_edit.php
+++ b/program/actions/settings/folder_edit.php
@@ -157,11 +157,14 @@ class rcmail_action_settings_folder_edit extends rcmail_action_settings_folders
                 'value' => $select->show($selected),
             ];
 
-            $up = new html_button(['id' => 'move-folder-up', 'class' => 'move-folder-up'])->show($rcmail->gettext('reorder_folder_up'));
-            $down = new html_button(['id' => 'move-folder-down', 'class' => 'move-folder-down'])->show($rcmail->gettext('reorder_folder_down'));
+            $upBtn = new html_button(['id' => 'move-folder-up', 'class' => 'move-folder-up']);
+            $downBtn = new html_button(['id' => 'move-folder-down', 'class' => 'move-folder-down']);
             $form['props']['fieldsets']['location']['content']['order'] = [
                 'label' => $rcmail->gettext('reorder_folder'),
-                'value' => html::div([], [$up, $down]),
+                'value' => html::div([], [
+                    $upBtn->show($rcmail->gettext('reorder_folder_up')),
+                    $downBtn->show($rcmail->gettext('reorder_folder_down')),
+                ]),
             ];
         }
 
--- a/program/actions/settings/folder_reorder.php
+++ b/program/actions/settings/folder_reorder.php
@@ -9,7 +9,7 @@ class rcmail_action_settings_folder_reorder extends rcmail_action_settings_folde
      *
      * @param array $args Arguments from the previous step(s)
      */
-    #[Override]
+    #[\Override]
     public function run($args = [])
     {
         $rcmail = rcmail::get_instance();
--- a/program/actions/settings/folder_reorder.php
+++ b/program/actions/settings/folder_reorder.php
@@ -1,0 +1,22 @@
+<?php
+
+class rcmail_action_settings_folder_reorder extends rcmail_action_settings_folders
+{
+    protected static $mode = self::MODE_AJAX;
+
+    /**
+     * Request handler.
+     *
+     * @param array $args Arguments from the previous step(s)
+     */
+    #[Override]
+    public function run($args = [])
+    {
+        $rcmail = rcmail::get_instance();
+        $list = array_flip(rcube_utils::get_input_value('folderorder', rcube_utils::INPUT_POST, true));
+        $rcmail->user->save_prefs(['folder_order' => $list]);
+
+        $rcmail->output->show_message('successfullysaved', 'confirmation');
+        $rcmail->output->send();
+    }
+}
--- a/program/actions/settings/folders.php
+++ b/program/actions/settings/folders.php
@@ -34,6 +34,7 @@ class rcmail_action_settings_folders extends rcmail_action_settings_index
         $storage = $rcmail->get_storage();
 
         $rcmail->output->set_pagetitle($rcmail->gettext('folders'));
+        $rcmail->output->set_env('folder_ordered_manually', !empty($rcmail->user->get_prefs()['folder_order']));
         $rcmail->output->set_env('prefix_ns', $storage->get_namespace('prefix'));
         $rcmail->output->set_env('quota', (bool) $storage->get_capability('QUOTA'));
         $rcmail->output->include_script('treelist.js');
@@ -262,6 +263,10 @@ class rcmail_action_settings_folders extends rcmail_action_settings_index
             'id' => $idx,
             'class' => trim($data['class'] . ' mailbox'),
         ];
+        // Only allow reordering of non-protected folders.
+        if ($data['protected']) {
+            $attribs['class'] .= ' protected';
+        }
 
         if (!isset($data['level'])) {
             $data['level'] = 0;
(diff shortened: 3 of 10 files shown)


johndoh:
I like the idea of allowing users to reorder folders. I think having different drag behavior depending where on the <li> element you click is a bit clunky, it would be nice if both functions (change parent and change order) could be combined somehow. Putting the drag icon next to the subscription toggle could lead to fat finger problems, I suggest putting it at the start, something like this:

Screenshot 2025-07-10 183123

Out of scope for this PR but adding a similar drag icon the rule list in the managesieve plugin would also be nice, highlight to the users dragging is possible - so the implementation should be done is such a way the drag icon can be added to any list.

alecpl:
On desktop we could use a drag+Shift button. There would be no need for an icon. Well, the icon could appear while dragging, or not. The drop target indicator would be different, i.e. to move folder we highlight the target folder (on hover), while to re-order a folder we highlight the border between folders.

That would not work on a touch device, though. But I wouldn't mind if that functionality was a desktop feature only.

Finally, maybe it would make sense to limit reordering to top-level folders only?

In any case it would be good to have it consistent across all places (i.e. Filters).

pabzm:
I found a technical solution to combine both actions, reordering and moving into (and out of) subfolders. It will take some more working days to adapt it to Roundcube's code, though.

pabzm:
The current state needs some more styling love, but maybe you can already have a look at the functionality, which works as far as I see. There's no sorting handle anymore since all dragging now allows to sort and move into and out of sub-folders.

pabzm:
Locally the browser tests are green, so I assume those errors are a CI artefact. The "Static Analysis" failures are also unrelated.

johndoh:
I like the combined drag action better. For me in dark mode an extra line appears between parent and child folders not just on the setting screen but on the mail and contacts screens as well. Using Firefox. 

If the last folder in the list has a sub folder and you want to drag a top level folder to the bottom of the list its tricky.

When I drag a sub folder from one parent to another it saves but the UI does not update, have to refresh the page.

Neustradamus:
@pabzm: Nice PR, good job!

Note: A feature must be here to have A->Z order OR manual choice.

pabzm:
@johndoh Thank you for the feedback! I improved the styling in dark (and light) mode a little, and gave the list's a bit more space at their end to make it easier to drop elements there.

> When I drag a sub folder from one parent to another it saves but the UI does not update, have to refresh the page.

I cannot reproduce that. Could you check for errors in the browser console or anything else that appears remarkable when that happens?

(I'll continue work on this in ~2 weeks.)

pabzm:
@Neustradamus The default and fallback is alphabetical ordering as long as you don't start to order the folders manually.

johndoh:
I think may be the list not updating problem only happens when a custom order is first applied.

program/js/app.js:8183
`Uncaught ReferenceError: next_sibling is not defined`

for me this fixes the issue:
```patch
diff --git a/program/js/app.js b/program/js/app.js
index 08a57a0db..930741e4c 100644
--- a/program/js/app.js
+++ b/program/js/app.js
@@ -8163,11 +8163,9 @@ function rcube_webmail() {
             delete ref.env.subscriptionrows[fname];
         });

-        if (this.env.folder_ordered_manually) {
-            // We need to store this information now, because it's not available anymore after removing the row from
-            // the DOM.
-            next_sibling = row.nextElementSibling;
-        }
+        // We need to store this information now, because it's not available anymore after removing the row from
+        // the DOM.
+        next_sibling = this.env.folder_ordered_manually ? row.nextElementSibling : null;

         // get row off the list
         row = $(row).detach();
```

alecpl:
It does not seem to be finished. It has a lot of issues at this point.

Neustradamus:
@pabzm: Can you look your PR?

It will be nice to have in Roundcube 1.7 final...

Thanks in advance.

pabzm:
I fixed some more styling issues and handling quirks and added buttons to move a folder up and down the list (for people that can't or don't want to use drag-and-drop).

From my point of view this is now ripe for a proper review and hopefully merging, too.

pabzm:
Here's a new screenshot with the mentioned buttons:

image

pabzm:
> It does not seem to be finished. It has a lot of issues at this point.

@alecpl Could you let me know which those are?

pabzm:
I rebased on the latest of the "master" branch to try and get rid of the unrelated CI Coding Style errors

alecpl:
- That extra spacing (row height) is not good.
- Dragging a folder that wasn't selected creates bad look, i.e. the dragged object is transparent. But if it is selected it's kinda hard to get what's going on behind it (where the drop areas appear). Also, sometimes it's hard do match the dragging object with the drop area (especially when dealing with folder hierarchies).
- Dropping the folder scrolls the list up.
- Dropping into the initial place should probably not send a request (and not show "Successfully saved").
- I notice some jumping effect (height of rows changes) when hovering over a sub-tree (folder with sub-folders).
- The Up/Down buttons in folder info are kinda useless if you can't see the list on the same screen, i.e. on a phone.
- Mouse-up event outside the list does not stop the drag action, but should.

pabzm:
1. I think the extra spacing is good, because it allows to more easily move folders between other folders. Without the padding there's only a few pixels of height to target, which are hard to find.

2. I introduced a little transparency to solve this:

Screenshot 2025-11-13 at 14 35 21

---------

Screenshot 2025-11-13 at 14 35 08

--------


3. Dropping the folder scrolls the list up.

I see only a little jumping, not a full scroll up. Here's a screencast:


https://github.com/user-attachments/assets/da7fe4e3-7a38-4dfa-9bea-960f6c029266

Is that what you mean? Or something else?


4. I had cancelled that "feature" because I had only found complicated ways to do it, but now I had another look and accidentally found a simpler way, so this is fixed.

5. I can't reproduce the jumping, except for the other folders going out of the way to make space for the new element. Could you maybe post a screencast?

6. I agree that the up/down-buttons could be improved on narrow screens, but at least it's possible at all to folders on mobiles now, so I don't think they're useless.

7. That is now fixed, thank you for the hint.

alecpl:
> 1. I think the extra spacing is good, because it allows to more easily move folders between other folders. Without the padding there's only a few pixels of height to target, which are hard to find.

I think it's really bad. It's inconsistent with the rest. And drag effect is not nice either. I think this makes UX worse.

alecpl:
Here's my suggestion: Use the drag effect from master, i.e. the dark box with folder name, don't lock dragging inside the list widget. Whenever a folder is hovered (yellow bg) display both move drop area and reorder drop area.

pabzm:
> It's inconsistent with the rest.

What rest do you mean? How is it inconsistent? Please explain, so I don't have to guess.


> don't lock dragging inside the list widget

You mean the dragged elements should be allowed to leave the outer-most list element? What is the advantage of that? They can't be dropped anywhere outside of that list element.

Are all the previously listed issues resolved from your point of view? Points 3 and 5 seem to be open to me.

alecpl:
> I'm going to go through this line by line to try to make this productive.
> 
> > It's inconsistent with the rest.
> 
> What rest do you mean? How is it inconsistent? Please explain and don't make me guess.

All other lists. The different row height and extra spacing is different. Yes, there's one exception, mail messages list, but it's completely different.

> > the drag effect from master
> 
> What do you mean by "master" in this context? The git branch? How does it contain a drag effect?

I mean the existing solution. You can now move folders by drag-and-dropping them.

> > don't lock dragging inside the list widget
> 
> You mean the dragged elements should be allowed to leave the outer-most list element? What is the advantage of that? They can't be dropped anywhere outside of that list element.

Yes, I think it is better because: 1) the dragged element is on the right side of the cursor, so it does not cover the whole record below. I.e. you see better the drop area. The dark background of the dragged element also matters here. 2) you can drop outside the list to stop (cancel) dragging without consequences.

pabzm:
Thank you for the explanation! I'll have a look if I can integrate your suggestions.

pabzm:
@alecpl 	 Here's my adaptions to your suggestions. I also moved a lot of code into `treelist.js`, were it's more suitably placed, I think. Please have another look.

In this short screencast you can see that it's quite finicky to move a folder into another folder with the small line heights. Please try yourself. I could only make this more easy by increasing the line heights.

https://github.com/user-attachments/assets/8c057372-24b9-402c-851b-600ee727fb8f

pabzm:
Rebased onto the lastest of "master" to get rid of the CI errors

alecpl:
I still don't like it, but I have two options to try:
1. Whenever you're above a folder display all drop areas at the same time, i.e. above, below and subfolder. This will probably create a jumping effect, so it will probably not work.
2. Don't display additional drop areas, highlight the folder only. When you drop a folder display a popup menu (like we do when you drop a mail or contact with Shift key) with three options: move above, move below, move below (as a subfolder).

pabzm:
I will not pivot to a completely new behaviour with this.

If small changes would be required to make this merge-able I would put in a little more effort, but beyond that I won't spend more time on this. If you don't want to take it, I'll leave it open for someone else to adopt.

alecpl:
I think it requires more than small changes. I have one more option to consider. Instead of drag-n-dropping a folder into any place we should consider a different approach.

It would be marking a folder as a special folder ("high priority folder" could work to, but might get confusing with mail priority). It would make any non-special folder to be put immediately after special folders, i.e. close enough to the top of the list. This way there's not much room for the feature to be abused, and it's much simpler, at the same time fulfilling 90% of cases. Imo, a main case users want is "I want this folder somewhere on top so it is easier to find", but I might be wrong.

PR #9929, opened 2025-07-14
Update labels.inc

konvlachos:
Fix Typo


alecpl:
The source for localization is at Transifex. I fixed the type there and pulled into master.

PR #9932, opened 2025-07-15
Make targets to help publishing releases

pabzm:


alecpl:
This PHP script is not really CI related. How about we put it in the `roundcube/roundcube-web` repository, so the Makefile could just fetch it from there. Probably making it a shell script would be needed, as you can't just use `.php` extension this way.

pabzm:
Moving the script to `roundcube/roundcube.github.com` (which you mean, I guess?) sounds good to me. But I wouldn't like to re-implement it, and would rather suggest to just make the Makefile in this repo write to stdout the line to execute in that other repo. To commit the generated data one needs to change to that other repo anyways, so we can also run the script in there. Alright?

alecpl:
The script needs files to work on. So, it might not be that convenient, but fine with me.

pabzm:
I moved and adapted the script, as well as the Makefile targets here.

pabzm:
@alecpl If you have any further comment please let me know in the next days. Else I'd merge this in ~1 week.

PR #9935, opened 2025-07-20
Enforce leading backslash for non-namespaced non-Roundcube uses

mvorisek:
replaces #9680, extracted from #9481

#9481 is about to add namespace to Roundcube classes. This implies all non-namespaced non-Roundcube uses need to use leading backslash.

In order to keep #9481 diff minimal, we land this prerequisite in a separate PR.

To keep the codebase up-to-date before #9481 is fully done, PHP CS Fixer uses customized `fully_qualified_strict_types` rule so the codebase is asserted using CI.

Non-functional change only.


mvorisek:
@alecpl @pabzm can I have you review here?

The CI errors are unrelated.

alecpl:
The change seems fine, but that extra 1000 lines of code...

mvorisek:
The extra LOC are because of `.php-cs-fixer-fully_qualified_strict_types.php` added & customized fixer rule.

The easiest way to review this is by commit by commit basis and rerun fixer localy to reproduce the result of the large "run fixer" commit.

pabzm:
I have no strong opinion on this. Having to add such a huge config file doesn't feel very good, though. But I'm not familiar with the reasons for these ongoing style/namespace changes and won't block them.

mvorisek:
Thank you. The large fixer rule file is temporary after all files will be namespaced, this is to easy the migration in a controlled way. Doing 1000 changes by hand is nearly impossible.

Neustradamus:
@mvorisek: If it is to be "Standard", I valide it.

It must be for 1.6.x/1.7.x too.

mvorisek:
@alecpl should I retarget this PR?

alecpl:
No need to retarget. Also, I don't think we have to backport it to 1.6, it already deviated a lot from master.

Issue #9598, opened 2024-08-20
Microsoft 365 OAUTH2 authentication workflow no longer works


lambjs:
### Prerequisites

- [X] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I can recreate the issue with all plugins disabled

### Describe the issue

I may be doing something wrong here, but my attempts to get IMAP XOAUTH2 authentication working for a tenant in Microsoft 365 have failed on a new install. I'll provide a sample of my configuration below.

``` php
<?php
// From html/config/config.inc.php
$config['default_host'] = 'ssl://outlook.office365.com';
$config['smtp_server'] = 'ssl://smtp.office365.com';

$config['oauth_provider'] = 'outlook';
$config['oauth_provider_name'] = 'Outlook.com';
$config['oauth_client_id'] = "-- snip --";
$config['oauth_client_secret'] = "-- snip --";
$config['oauth_auth_uri'] = "https://login.microsoftonline.com/common/oauth2/v2.0/authorize";
$config['oauth_token_uri'] = "https://login.microsoftonline.com/common/oauth2/v2.0/token";
$config['oauth_identity_uri'] = "https://graph.microsoft.com/v1.0/me";
$config['oauth_identity_fields'] = ['email', 'userPrincipalName'];
$config['oauth_scope'] = "https://outlook.office365.com/IMAP.AccessAsUser.All https://outlook.office365.com/SMTP.Send User.Read offline_access";
$config['oauth_auth_parameters'] = ['nonce' => mt_rand()];
```

I can confirm that I have configured my application in Entra to include the requisite Graph API delegated permissions. I've also configured my scope for Multitenant without personal accounts ie. the Manifest for my application in Entra admin center shows "signInAudience": "AzureADMultipleOrgs". I've seen some other forums mention that this is required and it won't work with other SignInAudience values? Is this true, and if so, should it be added to the example configuration in defaults.inc.php?

When I connect with the above configuration, I get an incorrect audience error when attempting to fetch user details from the identity uri (graph.microsoft.com). When I use the graph namespace instead, I get "NO AUTHENTICATION failed." during IMAP login.

I created my own minimal oauth 365 proof of concept to see if I could rule out roundcube as the problem. Based on my findings, and skipping the identity uri call using the same scope as above, I was able to get imap auth working.

[...]

PR #9939, opened 2025-07-24
Fixes Microsoft 365 OAUTH2 authentication workflow no longer works #9598

lambjs:
See:
[Microsoft 365 OAUTH2 authentication workflow no longer works](https://github.com/roundcube/roundcubemail/issues/9598)
#9598 

Modfiy rcmail_oauth.php and config defaults and examples to allow changing of scope for identity API calls. This is required after Microsoft has changed their API's such that the scopes used for regular mailbox interactions are incompatible with those used for identity operations.

### DIFF
--- a/program/include/rcmail_oauth.php
+++ b/program/include/rcmail_oauth.php
@@ -161,6 +161,7 @@ class rcmail_oauth
                 'language' => ['locale'],
             ]),
             'scope' => $this->rcmail->config->get('oauth_scope', ''),
+            'scope_identity'  => $this->rcmail->config->get('oauth_scope_identity', ''),
             'timeout' => $this->rcmail->config->get('oauth_timeout', 10),
             'verify_peer' => $this->rcmail->config->get('oauth_verify_peer', true),
             'auth_parameters' => $this->rcmail->config->get('oauth_auth_parameters', []),
@@ -641,6 +642,13 @@ class rcmail_oauth
 
             // request user identity (email)
             if (empty($username)) {
+                if($this->options['scope_identity']) {
+                    $this->mask_auth_data($data);
+                    $refresh_response = $this->refresh_access_token($data, $this->options['scope_identity']);
+                    $data = $refresh_response['token'];
+                    $authorization = $refresh_response['authorization'];
+                }
+
                 $fetched_identity = $this->fetch_userinfo($authorization);
 
                 $this->log_debug('fetched identity: %s', json_encode($fetched_identity, true));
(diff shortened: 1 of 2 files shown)


alecpl:
Couldn't that be implemented with a single extra request instead of two?

Neustradamus:
@alecpl: I will be fixed in 1.7?

lambjs:
> Couldn't that be implemented with a single extra request instead of two?

@alecpl:

I think you're right. I had assumed that using the current token to do a refresh would invalidate the current token data and require another refresh to get another working token with the correct scope for regular mailbox operations. After changing the naming so I didn't overwrite the pre-existing auth data, and removing the second call to refresh_access_token, my tests seem to succeed. I've pushed another change, please take a look and let me know if okay to merge?

program/include/rcmail_oauth.php
```diff
@@ -641,7 +642,14 @@ public function request_access_token($auth_code, $state = null)
...
             if (empty($username)) {
-                $fetched_identity = $this->fetch_userinfo($authorization);
+                if($this->options['scope_identity']) {
+                    $this->mask_auth_data($data);
+                    $refresh_response = $this->refresh_access_token($data, $this->options['scope_identity']);
+                    $data_ident = $refresh_response['token'];
+                    $authorization_ident = $refresh_response['authorization'];
+                }
+
+                $fetched_identity = $this->fetch_userinfo($authorization_ident);
```
alecpl, line 652: `$authorization_ident` does not exist if `$this->options['scope_identity']` is unset.

program/include/rcmail_oauth.php
```diff
@@ -641,7 +642,14 @@ public function request_access_token($auth_code, $state = null)
 
             // request user identity (email)
             if (empty($username)) {
-                $fetched_identity = $this->fetch_userinfo($authorization);
+                if($this->options['scope_identity']) {
```
alecpl, line 645: Add a space after `if`, please.

program/include/rcmail_oauth.php
```diff
@@ -741,6 +752,10 @@ public function refresh_access_token(array $token)
                 'client_secret' => $oauth_client_secret,
             ];
 
+            if($change_scope) {
```
alecpl, line 755: Add space after `if`, please. And rename `$change_scope` to `$scope`.

alecpl (changes requested):

alecpl:
@lambjs would you mind finishing this, please? We're getting close to the 1.7-rc release. It would be nice to include this.

Neustradamus:
@lambjs: Please look your PR, it will be nice to have in 1.7...

Thanks in advance.

martinellimarco:
Hi all,
I’ve tested this patch on Roundcube 1.7-RC4 and it resolves the current Microsoft 365 OAuth2 login failure.

I noticed the PR seems to be stalled due to a few requested changes.

@lambjs if you’re still around, would you be able to push the requested fixes?

If not, I’m happy to address them and either update this PR (if possible) or open a new one that supersedes #9939 while preserving authorship.

@alecpl please let me know which option you prefer so we can hopefully get this landed before 1.7 final.

PR #9942, opened 2025-07-25
Reduce duplicate Github CI workflow runs

pabzm:
The tasks are now run on push-events to the branches "master", and "release-*", as well as on pull-request-events (i.e. all pushes to branches that have a pull request).

This appears to be the easiest way to reduce duplicate workflow runs.

### DIFF
--- a/.github/workflows/browser_tests.yml
+++ b/.github/workflows/browser_tests.yml
@@ -2,6 +2,9 @@ name: E2E
 
 on:
   push:
+    branches:
+      - main
+      - release-*
   pull_request:
 
 permissions:
@@ -76,4 +79,3 @@ jobs:
         with:
           name: screenshots
           path: tests/Browser/screenshots
-          
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: CI
 
 on:
   push:
+    branches:
+      - main
+      - release-*
   pull_request:
 
 permissions:
--- a/.github/workflows/message_rendering.yml
+++ b/.github/workflows/message_rendering.yml
@@ -2,6 +2,9 @@ name: Message Rendering
 
 on:
   push:
+    branches:
+      - main
+      - release-*
   pull_request:
 
 permissions:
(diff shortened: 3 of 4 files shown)


mvorisek:
I strongly do not support this idea as this will prevent CI on forks on pushes. If there is no huge issue with CI minutes, better to run a few jobs more IMO.

pabzm:
@mvorisek I don't want to cut off forks, I'm just want to get rid of the duplicated jobs (especially the E2E-jobs), that add several minutes to the waiting time until the CI is green. Do you happen to know another approach how to do that?

mvorisek:
Unfortunatelly there is no way to exclude push when a push is to a branch with PR.

There is a reason. Push = pushed commit. Pull_request = target branch + changes. Just to explain things.

Now to your question and the solution. The best way is to fork this repo even when you are maintainer. Then the fork's pushes won't be visible in the PR CI checks. And the CI minutes will be counted to your account (for public repos free) thus the target repo CI minutes will not be limitted.

So I propose to close this PR as IMO any CI limitting on branch name is not ideal.

pabzm:
I would like to actually stop wasting computation time, not just make it count against another account.

@mvorisek Just to clarify: You do want to run all CI workflows on every push to any of your fork's branches?

If so, then I guess the solution would be to let it run on any `push` and remove the `pull-request`-line, right?

mvorisek:
Yes.

The `pull-request` removal is not a solution, as for PRs from forks no CI would be run/visible on PR page.

Compared to AI datacenters and all other hungry devices the impact of whole Github servers is minimal and with every CPU generation the energy consumption is lower and lower... No really need to carry about.

alecpl:
It seems using a fork is the way to go, and this is what I usually do. This will also make sure that list of branches in this repo is short and clean.

pabzm:
I don't like working in forks much, because it makes working together on a branch as a team more complicated. But I understood that you prefer it and will respect it. So I'll close this PR and will occasionally try to find a solution that works for/with all workflows.
Thank you for the feedback!

Issue #9941, opened 2025-07-24
href attributes for some task buttons are broken in v1.7-beta


gurnec:
### Prerequisites

- [x] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I can recreate the issue with all plugins disabled

### Describe the issue

The onclick handlers are fine, but the href attributes for some buttons, including `Mail`, `Contacts`, `Settings`, and `Logout` are broken. Trying to do an `Open link in new tab` leads to a 404.

### What browser(s) are you seeing the problem on?

Chrome

### What version of PHP are you using?

v8.1

### What version of Roundcube are you using?

v1.7-beta

### Cause

Here's a [link to the diff](https://github.com/roundcube/roundcubemail/commit/64df318a733448585db963f6696da133451a6b70#diff-90d47c156392db4aa87ee7cb3716224aa015ca2477c1c75410ed06d8708d9b93) which is _a_ cause of the issue. That link doesn't work well for me, (I think the diff is too big for GH), but it's line 999 of `rcmail_output_html.php` in that diff.

The changed regex in `fix_paths()` now matches links like `OhfBGLzpfgQToW5o/?_task=mail`, in other words links with a single query parameter (but not multiple), and then that link is transformed to `OhfBGLzpfgQToW5o/static.php/skins/elastic/OhfBGLzpfgQToW5o/?_task=mail`.

```php
-        $regexp = '!(src|href|background|data-src-[a-z]+)=(["\']?)([a-z0-9/_.-]+)(["\'\s>])!i';
+        $regexp = '!(src|href|background|data-src-[a-z]+)=(["\']?)([a-z0-9/_.?=-]+)(["\'\s>])!i';
```

alecpl:
Fixed.

PR #9943, opened 2025-07-27
only apply fix_path for href attrib in <link>s

johndoh:
for #9941

I could not think of a use case for the `href` of an `<a>` to go via `static.php` so I split up the existing regex to allow special handling of the `href` attribute and take account of the tag its being used it.

It could be that this is a bit too complicated and really we only need to fix the path when we are in a `<link>` tag. W3Schools says `href` is only valid in `<a>`, `<area>`, `<base>` and `<link>` tags. So the new pattern could be changed to specifically match `<link ` rather than ignore `<a `.

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -992,7 +992,11 @@ class rcmail_output_html extends rcmail_output
      */
     protected function fix_paths($output)
     {
-        $regexp = '!(src|href|background|data-src-[a-z]+)=(["\']?)([a-z0-9/_.?=-]+)(["\'\s>])!i';
+        $regexp = [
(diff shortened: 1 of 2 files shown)


alecpl:
Another option might be to change the `resource_location()` method that it ignores URLs that (after removing everything after `?` and `#`) do not point to a file (must end with `.something`). What do you think?

johndoh:
Not in `resource_location()` because `file_callback()` would still add the skin path to the URLs. The root cause is because `?` was added to the pattern for `fix_paths()` so now both `file_callback()`  and then `resource_location()` get called in more cases than before.

Doing something with the existence of a file extension in `file_callback()` would solve the basic case but I think a 3rd party skin or a plugin could add `<a>` with links to anywhere it wants and putting those hrefs through post processing like this is unnecessary. I have added a couple of tests for external links which I think this code should not alter.

Rather than the "not an `<a>`" pattern I have added here I think adding a "only hrefs in `<link>`" would be enough.

alecpl:
Let's do "only hrefs in `<link>`".

PR #9945, opened 2025-07-28
Prepend group-names to display-name

pabzm:
This is not optimal handling, but the most appropriate one as long as we don't actually support groups in addresss-lists. This way users can at least see the group's display-name. And we don't strip text that might be relevant to spot abusive emails.

Previously group-names were just removed, which makes it harder to spot such abusive emails.

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -1380,10 +1380,11 @@ class rcmail_action_mail_index extends rcmail_action
             }
             $mailto = rcube_utils::idn_to_utf8($mailto);
 
-            // Homograph attack detection (#6891)
-            if ($spoofcheck && !self::$SUSPICIOUS_EMAIL) {
-                self::$SUSPICIOUS_EMAIL = rcube_spoofchecker::check($mailto);
-            }
+            // Homograph attack detection (#6891),
+            // and phishing email prevention (#1488981) (e.g. "valid@email.addr <phishing@email.addr>")
+            $show_fraud_warning = $spoofcheck
+                && !self::$PRINT_MODE // Don't show the warning in print mode, because there's no danger of interaction.
+                && (rcube_spoofchecker::check($mailto) || ($name && $name != $mailto && preg_match('/@|＠|﹫/', $name)));
 
             if (self::$PRINT_MODE) {
                 $address = '&lt;' . rcube::Q($mailto) . '&gt;';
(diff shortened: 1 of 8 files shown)


pabzm:
@alecpl Please note that this PR needs some more work so that the extended name gets shown in the HTML, but I would like to have some feedback before I go on with this.

alecpl:
Makes sense. Considering that even the most recent RFC about it is internally inconsistent, this seems the way to handle this issue. I'm not sure what more work you plan exactly.

pabzm:
Ok, thank you!

Currently https://github.com/roundcube/roundcubemail/blob/fbed38e2a9033542d8f593f4167d61bcdc5f146c/program/actions/mail/index.php#L1375-L1377 prevents the display-name from header values like `Test <attacker@example.org>:<admin@example.net>` are shown (to prevent phishing), thus swallowing the effort to not hide information. I'll try to make it show a warning instead.

pabzm:
I changed the code so in case of a suspected phishing it now shows all details prefixed with a warning icon that carries an explaining title:

Screenshot 2025-07-29 at 14 21 30


From my point of view this is now ready for review.

alecpl:
We also have this in the code:
```
            // Homograph attack detection (#6891)
            if ($spoofcheck && !self::$SUSPICIOUS_EMAIL) {
                self::$SUSPICIOUS_EMAIL = rcube_spoofchecker::check($mailto);
            }
```
and we display a warning message, but we don't add a warning icon for the specific email address.

So, it seems we have now two ways to warn a user about essentially the same issue. It would make sense to unify this.

Also, it seems the "spoofcheck warning" is displayed in print mode, but your new icon isn't. So, we would need to think whether we should display warning/icon/both in print mode and change the code consistently.

pabzm:
Thank you for the feedback, I agree that unifying is a good point and will work on that in the upcoming days.

For printing I would leave out the warning-icon and -message, since there's no danger of interaction. Alright?

pabzm:
I propose to show the homograph-triggered icon also before the address that it refers to, like this:

image

The title text is be the same for both trigger-reasons to keep it simple.

-----

In print-mode I propose to not show the warnings, since there's no interaction possible (except for copy&paste, which I would assume that people don't do from the print-mode):


Screenshot 2025-08-20 at 10 14 50

pabzm:
@alecpl If you have any further comment please let me know in the next days. Else I'd merge this in ~1 week.

pabzm:
@alecpl I'll merge this in a few days unless you object

PR #9949, opened 2025-07-31
Wash the `name` attribute also on more elements

pabzm:
It can pollute the document's namespace unless handled.

The elements the code now handles the `name`-attribute on is based on my research, if you know of further relevant elements, please let me know.

### DIFF
--- a/program/lib/Roundcube/rcube_washtml.php
+++ b/program/lib/Roundcube/rcube_washtml.php
@@ -326,7 +326,7 @@ class rcube_washtml
                         $out = $value;
                     }
                 } elseif ($this->_css_prefix !== null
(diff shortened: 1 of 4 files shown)


alecpl:
First, I'm not sure how It can "pollute the document's namespace".
Second, another approach would be to get rid of 'name' attributes.
Third, I just noticed that we do not (as far as I see) fix (add prefix to) css styles containing `[name=...]` selector. While not required for this change, and probably a rare case, it would be good to have.

That being said, this change seem to not break anything, so I'm fine with it in general. Add 'textarea' and 'select' to the list.

pabzm:
@alecpl @mvorisek If you have any further comment please let me know in the next days. Else I'd merge this in ~1 week.

pabzm:
@alecpl, @mvorisek  I'll merge this in a few days unless you object

mvorisek:
(I commented on this PR, but I won't do a full review)

PR #9955, opened 2025-08-10
markdownify remaining docs

johndoh:
trivial bit of housekeeping to translating the last 2 remaining doc files to markdown

I translated INSTALL and UPGRADING. The LICENSE file is the official one from https://www.gnu.org/licenses/gpl-3.0.md


alecpl:
I would like to, separately, consider putting all these files into a `/docs` folder. This would cleanup and simplify some code, and I think it's supported by Github. What do you think?

johndoh:
GitHub docs say README, SECURITY and  CODE_OF_CONDUCT are still recognised in a `docs` folder. I think LICENSE(.md) is only recognised in the root, the docs are not totally clear and only say
> Most people place their license text in a file named LICENSE.txt (or LICENSE.md or LICENSE.rst) in the root of the repository;

Would CHANGELOG(.md) move to the docs folder? I'm not sure if README and CHANGELOG stay in `root` and the rest move? I have no problem with them both moving, I'm just not sure what is expected.

The other files are unique to Roundcube I think:
* SOCIAL_WORK_GUIDELINES.md
* RELEASE_MANAGEMENT.md
* INSTALL
* UPGRADING

pabzm:
I would expect a README in the root folder. The other files could be moved, from my point of view. But I don't mind them in the root folder, either.

alecpl:
> I'm not sure if README and CHANGELOG stay in `root` and the rest move?

Let's do it. These two stay, the rest into docs/.

alecpl:
The Makefile change does not work in CI.

PR #9960, opened 2025-08-31
fix: Sanitize filename on download

oscardimanno:
Hi @alecpl,

Thanks again for your feedback on the previous PR (#9906).

I've moved the changes on a dedicated branch, implemented your suggestions for the filename handling, and added a test to confirm the fix.

This PR replaces the previous (#9906). This PR is still compliant to the [IETF RFC Standard: 6266, section 4.3](https://datatracker.ietf.org/doc/html/rfc6266#section-4.3).

All CI checks should be passing now.

Ready for your review when you have a moment.

Thanks

### DIFF
--- a/program/lib/Roundcube/rcube_output.php
+++ b/program/lib/Roundcube/rcube_output.php
@@ -259,12 +259,14 @@ abstract class rcube_output
             $fallback_filename = preg_replace('/[^a-zA-Z0-9_.(),;@+ -]/', '_', $filename);
             $disposition .= "; filename=\"{$fallback_filename}\"";
 
-            $filename = rawurlencode($filename);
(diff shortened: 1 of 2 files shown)


Neustradamus:
@oscardimanno: Thanks for this new PR!

Please add in description that it respects IETF RFC standard...

It replaces:
- https://github.com/roundcube/roundcubemail/pull/9906

Hope that @alecpl, @pabzm can check it and will be integrated into 1.5/1.6/1.7 branches.

oscardimanno:
Hi @Neustradamus, thank you for your support.
I've changed the description as you suggested. Now should be okay.
Thanks for taking the time to help me improve the PR.
Thanks again!

PR #9964, opened 2025-09-04
Fix flaky browser test

pabzm:
Fixes `Browser/Mail/ListTest.php` on tablets by refreshing the page. Not sure why it's necessary, but it works reliably.


pabzm:
I'm merging this myself to get our CI green again. If the change doesn't sit well with someone we can still discuss to change it again.

PR #9968, opened 2025-09-06
Hestia Control Panel Driver Support

texxasrulez:
I made a Hestia Control Panel Driver. It is working great for me and wanted to help contribute. Thanks for all the hard work


pabzm:
Thank you for your contribution!

Unfortunately this PR still has 1438 changed files, which is way too much to review it. Most of those appear to be related to 3rd-party skins, maybe you could start by removing those?

PR #9969, opened 2025-09-06
Add delete button in toolbar on mobile like in Outlook

rstefko:
One of the most used actions now available in one click to save time 

Screenshot_20250906-221108 Vivaldi


alecpl:
I think this location is not great. I feel it makes it easy to delete the message by mistake. Maybe the footer would be better.

rstefko:
Maybe, but other location will not prevent you from accidental email removal.

It would be great to have something like Outlook - a popup that would be displayed for a few seconds to undo delete action.

image

PR #9970, opened 2025-09-08
Run test with PHP 8.5-rc

pabzm:
Since beta2 the PHP people call for testing for compatibility.

~~This PR is to find out which code needs changes to work (properly) with PHP v8.5.~~

~~This PR already includes a few obvious fixes. Some other spots still need attention.~~

Update: all failing tests and deprecation notices have been fixed in this PR already.

### DIFF
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -325,7 +325,7 @@ class rcmail extends rcube
         }
 
         // use existing instance
-        if ($id && isset($this->address_books[$id]) && ($this->address_books[$id] instanceof rcube_addressbook)) {
+        if ($id !== null && isset($this->address_books[$id]) && ($this->address_books[$id] instanceof rcube_addressbook)) {
             $contacts = $this->address_books[$id];
         } elseif ($id && !empty($ldap_config[$id])) {
             $domain = $this->config->mail_domain($_SESSION['storage_host']);
--- a/program/include/rcmail.php
+++ b/program/include/rcmail.php
@@ -300,15 +300,15 @@ class rcmail extends rcube
     /**
      * Return instance of the internal address book class
      *
-     * @param string $id        Address book identifier. It accepts also special values:
-     *                          - rcube_addressbook::TYPE_CONTACT (or 'sql') for the SQL addressbook
-     *                          - rcube_addressbook::TYPE_DEFAULT for the default addressbook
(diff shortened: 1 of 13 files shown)


pabzm:
I found solutions for all pieces that needed attention. Let me know what you think about them!

program/include/rcmail.php
```diff
@@ -325,7 +325,7 @@ public function get_address_book($id, $writeable = false, $fallback = true)
         }
 
         // use existing instance
-        if (isset($this->address_books[$id]) && ($this->address_books[$id] instanceof rcube_addressbook)) {
+        if ($id && isset($this->address_books[$id]) && ($this->address_books[$id] instanceof rcube_addressbook)) {
```
alecpl, line 328: $id can be "0".

program/lib/Roundcube/rcube_utils.php
```diff
@@ -721,10 +721,17 @@ public static function file2class($mimetype, $filename)
...
         $callback = static function ($matches) {
-            return chr(hexdec($matches[1]));
+            $bytevalue = hexdec((string) $matches[1]);
+            // chr() only covers values between 0 and 255. The following 4 lines are from the former default behaviour
+            // to ensure that, which is now deprecated, so we now explicitly do the shifting here.
+            while ($bytevalue < 0) {
+                $bytevalue += 256;
+            }
+            $bytevalue %= 256;
+            return chr($bytevalue);
```
alecpl, line 734: Considering that we work with `[0-9a-f]{2,6}` input here this whole callback function seems wrong. Isn't it? We should probably use `mb_chr()`.

alecpl (changes requested):

program/lib/Roundcube/rcube_utils.php
pabzm: I was wondering about that, too, but took the code from the documented original behaviour (according to https://www.php.net/manual/en/function.chr.php) and didn't want to change behaviour in this compatibility-fixing step.

But I agree that the switch to `mb_char()` would make sense. We don't have an encoding handy at the invocations of this method, though, would need to fallback to the internal encoding (or this change would grow a lot, which I'd rather avoid right here). Are you fine with that?

alecpl (approved):

PR #8689, opened 2022-08-25
Stay Logged In Plugin

Github-Citizen:
This plugin will add a toggle switch to the login page for the Elastic skin allowing users the option to stay logged in to Roundcube for days even after closing the browser. This does not create any extra 'record keeping' or objects. It simply extends the expiration time of the existing cookies Roundcube already uses to login users.

To the Roundcube team;
I want to give this plugin to the project (just to give back) and I claim no credit. Feel free to claim and polish the comments to add whatever license or copyright notices needed.

The limited language localization files were created using google translate and could use love from the community.

PR #9976, opened 2025-09-12
New plugin: Persisted login

pabzm:
This plugin adds a toggle switch into the login form of Roundcubemail's "elastic" skin, that makes the session live for a configured number of days.

Adapted from https://github.com/roundcube/roundcubemail/pull/8689

Hat tip and thanks to @Github-Citizen

### DIFF
--- a/plugins/persisted_login/README.md
+++ b/plugins/persisted_login/README.md
@@ -1,0 +1,26 @@
+Persisted login Roundcubemail plugin
+====================================
+
+This plugin adds a toggle switch into the login form of Roundcubemail's "elastic" skin, that makes the session live for a configured number of days (instead of only for the session).
+
+In effect, logins stay valid across network changes of clients, etc.
+
+From a technical point of view this plugin (if enabled) overrides `$config['session_lifetime']` (which sets the session garbage collection max lifetime in PHP) to match the number of days set in its own config.
+
+Usage
+-----
+
+Enable the plugin in your Roundcubemail's config:
+
+```php
+$config['plugins'] = [ …, 'persisted_login'];
+```
+
+By default logins are persisted for 7 days. That value can be changed via the config option `persisted_login_days` in the
+config file of this plugin. (Make sure that the config file ends in `.php` to have it used by Roundcubemail.)
+
+
+Credits
+-------
+
+Most of this code was actually written by [Github-Citizen](https://github.com/Github-Citizen) for https://github.com/roundcube/roundcubemail/pull/8689, which fell through due to styling issues, and only cleaned up and renamed for this plugin.
--- a/plugins/persisted_login/persisted_login.js
+++ b/plugins/persisted_login/persisted_login.js
@@ -1,0 +1,26 @@
+// Run on 'elastic' and all derivate skins.
+if (window.rcmail && (window.rcmail.env.skin == 'elastic' || rcmail.env.skin_extends?.includes('elastic'))) {
+    rcmail.addEventListener('init', () => {
+        const days = rcmail.env.persisted_login_days;
+        let txt = rcmail.gettext('switch_text', 'persisted_login');
+        txt = txt.replace('#', days);
+
+        const elems = $('<tr>').addClass('form-group row').append(
+            $('<td>').addClass('title').hide(),
+            $('<td>').addClass('input input-group input-group-lg').append(
+                $('<div>').addClass('custom-control custom-switch').css('padding', '1em 0').append(
+                    $('<input>').attr({
+                        type: 'checkbox',
+                        class: 'custom-control-input',
+                        id: '_persisted_login',
+                        name: '_persisted_login',
+                        value: '1',
+                    }),
+                    $('<label>').attr({ class: 'custom-control-label', for: '_persisted_login' }).text(txt)
+                )
+            )
+        );
+
+        $('#login-form table tbody').append(elems);
+    });
+}
--- a/plugins/persisted_login/persisted_login.php
+++ b/plugins/persisted_login/persisted_login.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
+ * Inject a toggle switch into the login form that makes the session live for a
+ * configured number of days (instead of only for the session).
+ */
+
+class persisted_login extends rcube_plugin
+{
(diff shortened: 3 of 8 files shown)


pabzm:
The E2E-tests are green locally. I'll fix them in the CI next week.

plugins/persisted_login/README.md
```diff
@@ -0,0 +1,24 @@
...
+Usage
+-----
+
+Enable the plugin in your Roundcubemail's config:
+
+```php
+$config['plugins'] = [ …, 'persisted_login'];
+```
+
+By default logins are persisted for 7 days. To change that value adapt the file `config.inc.php` in this directory.
```
alecpl, line 19: I don't see config.inc.php file here. If there is only one option you can just describe it here and tell users to put it in the main config file.

plugins/persisted_login/LICENSE
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
alecpl, line 1: We do not include LICENSE file for plugins. It can be removed. You should however add composer.json file.

plugins/persisted_login/persisted_login.php
```diff
@@ -0,0 +1,57 @@
...
+    public function login_page_template(array $args): array
+    {
+        $this->add_texts('localization', true);
+        $this->include_script('persisted_login.js');
+        return $args;
+    }
+
+    public function login_success(array $args): array
+    {
+        if (!isset($_POST['_persisted_login']) || !$_POST['_persisted_login']) {
```
alecpl, line 45: I prefer to use empty() in such cases.

plugins/persisted_login/persisted_login.php
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    #[\Override]
+    public function init(): void
+    {
+        if ($this->days === null) {
+            return;
+        }
+        $this->rc->output->set_env('persisted_login_days', $this->days);
```
alecpl, line 31: Normally I would put an empty line after closing brace.

plugins/persisted_login/persisted_login.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->include_script('persisted_login.js');
+        return $args;
+    }
+
+    public function login_success(array $args): array
+    {
+        if (!isset($_POST['_persisted_login']) || !$_POST['_persisted_login']) {
+            return $args;
+        }
+        $sessCookieName = $this->rc->config->get('session_name') ?? 'roundcube_sessid';
```
alecpl, line 48: These should better be `?:`.

plugins/persisted_login/persisted_login.js
```diff
@@ -0,0 +1,25 @@
+if (window.rcmail && window.rcmail.env.skin == 'elastic') {
```
alecpl, line 1: I'd like to support Elastic derivatives. When a skin extends Elastic you can find: `rcmail.env.skin_extends = ["elastic"]`

alecpl (changes requested):

pabzm:
Thank you for the review! I changed the styling as requested and added support for derivate skins.

pabzm:
The PHP8.5 deprecation warning in the Unit tests is fixed via https://github.com/DASPRiD/Enum/pull/18

pabzm:
I fixed the Browser tests by changing the plugin behaviour a little bit: It now uses a default of 7 days, and a minimum of 1 day.  Setting `persisted_login_days` to `0` doesn't work anymore to disable the plugin. Instead disabling should be done via `config.inc.php`. Additionally the config file is included as `config.inc.php.dist`, as it is in other plugins.
I think it's even more logical this way.

alecpl:
A composer.json file is missing.

ps. I don't like one thing about this plugin. It changes session lifetime for everyone. Which means all session records are garbage collected after X days. It probably should be fixed with a "expires at" column in the session table. It's not a stopper, I suppose.

Github-Citizen:
> ps. I don't like one thing about this plugin. It changes session lifetime for everyone. Which means all session records are garbage collected after X days.

Yes, but isn't that a necessary evil? If you don't extend the 'session_lifetime' then GC will remove the server side session even though the browser cookie is saved in the browser for x number of days.

> It probably should be fixed with a "expires at" column in the session table. It's not a stopper, I suppose.

But doesn't that require a fundamental change to the core of roundcube and outside of the scope of a plugin?

pabzm:
I also don't really like to rely on the clients to respect the expiry-timestamp of the cookie, either, but decided to go with it for now.

But as we agree on that part maybe we should do it properly and implement this features in core code, using an `expire_at` database column and memcache/redis key. Because having it in a plugin while changing core code explicitly for it feels wrong to me.

@alecpl Do you agree?

pabzm:
And as far as I see we don't need the `changed` column anymore, and should remove it – or rather rename it, instead of adding and removing.

alecpl:
> And as far as I see we don't need the `changed` column anymore, and should remove it – or rather rename it, instead of adding and removing.

What do you mean we don't need it? It is a very important column. We'd need both `changed` and `expires_at`.

pabzm:
From the code in `rcube_session_db` I figured that the purpose of `changed` is to mark the latest use of a session, so expired sessions can be filtered and removed. That could be done using `expires_at` instead, I think. What other purpose is there to `changed`?

pabzm:
In https://github.com/roundcube/roundcubemail/commit/c69f02fdaf728b3b618711d8ef42fc87a4a0f13d (branch `extended-session-lifetime`) I built a proof of concept of the feature in core code, which works without `changed`. It's not polished, but it works. Could you let me know which problems you see or would expect?

alecpl:
You might be right that one column would suffice. However, "time since the last change" is not the same as "time to expiry"

It would be easier to review if I could see the changed -> expires_at change separately from the persistent login.

pabzm:
Here's a PR with only the change to the session expiry without the new feature: #9990

> "time since the last change" is not the same as "time to expiry"

You're right, but as far as I can tell we use the timestamp only for expiry, and not for anything else.

PR #9977, opened 2025-09-15
Avoid warning with unset COMPOSE_ARGS

pabzm:


alecpl:
It seems there's a problem with the new certificate at git.kolab.org. We'd have to find a solution to that.
```
  Cloning into bare repository '/home/runner/.cache/composer/vcs/https---git.  
  kolab.org-diffusion-PNL-php-net-ldap3.git'...                                
  fatal: unable to access 'https://git.kolab.org/diffusion/PNL/php-net_ldap3.  
  git/': server certificate verification failed. CAfile: none CRLfile: none    
```

EDIT: The cert is valid, but the CA root cert is unrecognized, I've seen this in another place.

cmollekopf:
The reason is probably that the new root ca (Sectigo Public Server Authentication Root R46) is not available in the used image.

On fedora I have this:
```
$ rpm -ql ca-certificates | grep Sectigo | grep R46
/etc/pki/ca-trust/extracted/pem/directory-hash/Sectigo_Public_Server_Authentication_Root_R46.pem
```
This new root cert *should* have made it into all major trust stores (according to sectigo), so the solution should be to update the ca-certificates package I think.

alecpl:
@pabzm, it seems this change broke message rendering workflow. Could you take a look, please. 

https://github.com/roundcube/roundcubemail/actions/runs/18257173446/job/51980129985

pabzm:
https://github.com/roundcube/roundcubemail/pull/10000 fixes the failing jobs

PR #9979, opened 2025-09-16
supplemented chinese translations

bytesharky:
Add supplementary Chinese translations


alecpl:
ALl localization files, except csv2vcard, are managed via https://explore.transifex.com/roundcube/roundcube-webmail/

PR #9990, opened 2025-09-28
Replace changed by expires_at in session handling

pabzm:
This prepares using extended session lifetimes configurable per session

### DIFF
--- a/SQL/postgres.initial.sql
+++ b/SQL/postgres.initial.sql
@@ -37,12 +37,12 @@ CREATE TABLE users (
 
 CREATE TABLE "session" (
     sess_id varchar(128) DEFAULT '' PRIMARY KEY,
-    changed timestamp with time zone DEFAULT now() NOT NULL,
+    expires_at timestamp with time zone DEFAULT now() NOT NULL,
     ip varchar(41) NOT NULL,
     vars text NOT NULL
 );
 
-CREATE INDEX session_changed_idx ON session (changed);
+CREATE INDEX session_expires_at_idx ON session (expires_at);
 
 
 --
@@ -390,4 +390,4 @@ CREATE TABLE "system" (
     value text
 );
 
-INSERT INTO "system" (name, value) VALUES ('roundcube-version', '2022100100');
+INSERT INTO "system" (name, value) VALUES ('roundcube-version', '2025092300');
--- a/program/lib/Roundcube/session/db.php
+++ b/program/lib/Roundcube/session/db.php
@@ -106,12 +106,12 @@ class rcube_session_db extends rcube_session
     public function read($key)
     {
         if ($this->lifetime) {
-            $expire_time = $this->db->now(-$this->lifetime);
-            $expire_check = "CASE WHEN `changed` < {$expire_time} THEN 1 ELSE 0 END AS expired";
+            $expire_time = $this->db->now();
+            $expire_check = "CASE WHEN `expires_at` < {$expire_time} THEN 1 ELSE 0 END AS expired";
         }
 
         $sql_result = $this->db->query(
-            'SELECT `vars`, `ip`, `changed`, ' . $this->db->now() . ' AS ts'
+            'SELECT `vars`, `ip`, `expires_at`, ' . $this->db->now() . ' AS ts'
             . (isset($expire_check) ? ", {$expire_check}" : '')
             . " FROM {$this->table_name} WHERE `sess_id` = ?", $key
         );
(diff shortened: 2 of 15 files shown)


pabzm:
I rebased the branch onto the lastest of the "master" branch to fix the CI runs.

pabzm:
We should think about changing `set_auth_cookie()` to not use time-based slots but something else, because its anti-session-hijacking function is not really powerful anymore if it changes only every 18 hours.

alecpl:
Some of my comments in #9991 apply here.

pabzm:
@alecpl I fixed the version number in the "initial.sql"-files, and setting the expiry time in `rcube_session_php`. The only other of your comments in the other PR that I understand would apply to this PR is [this on](https://github.com/roundcube/roundcubemail/pull/9991#discussion_r2404374330), of which I don't understand why you consider the change wrong.
Are these the only issues you see with the changes I proposed in this PR?

pabzm:
I rebased onto the latest of the "master" branch.

johndoh:
This might be a bit picky but I think update files should also include something like
```sql
ALTER TABLE `session` RENAME INDEX `changed_index` TO `expires_at_index`; 
```
I think the name of the index is not that important but I think the schema should be the same when created by the initial file or through updates.

The change to `sqlite.initial.sql` is missing.

pabzm:
@johndoh Thank you for the review! I agree that renaming the index is a good idea and added it to the SQL migration files.

> The change to sqlite.initial.sql is missing.

What do you mean by that? I don't see a missing change in there that the other initial-files have.

(I also fixed the `roundcube-version` number to match the date-string of the corresponding SQL migration files.)

johndoh:
@pabzm my apologies. I somehow missed that `sqlite.initial.sql ` was in the diff. I mistakenly thought only the update file was there not the initial one.

The alter table statement for sqlite should be
```sql
ALTER TABLE `session` RENAME INDEX `ix_session_changed` TO `ix_session_expires_at`;
```
i think, i have not tested it, just corrected the index names.

and for postgres
```sql
ALTER TABLE `session` RENAME INDEX `session_changed_idx` TO `session_expires_at_idx`;
```
i think

pabzm:
@johndoh Thank you for that hint! I didn't expect a different index name in the differente databases.

pabzm:
If y'all would be fine with merging please let me know so I can cleanup the commit history before someone hits the button!

johndoh (approved):

pabzm:
I squashed all the commits into one, didn't change anything further.

pabzm:
@alecpl  Please have another look, I'd like to merge this soon.

SQL/postgres.initial.sql
```diff
@@ -390,4 +390,4 @@ CREATE TABLE "system" (
     value text
 );
 
-INSERT INTO "system" (name, value) VALUES ('roundcube-version', '2022100100');
+INSERT INTO "system" (name, value) VALUES ('roundcube-version', '2022092300');
```
alecpl: 2025092300

program/lib/Roundcube/session/db.php
```diff
@@ -200,10 +200,10 @@ public function gc_db()
     {
         // just clean all old sessions when this GC is called
         $this->db->query('DELETE FROM ' . $this->db->table_name('session')
-            . ' WHERE `changed` < ' . $this->db->now(-$this->gc_enabled));
+            . ' WHERE `expires_at` < ' . $this->db->now());
 
         $this->log('Session GC (DB): remove records < '
-            . date('Y-m-d H:i:s', time() - $this->gc_enabled)
+            . date('Y-m-d H:i:s', $this->db->now())
```
alecpl: As pointed in the other PR, this is wrong. After the change we use "now" so it does not make sense to log this as date, as a log line already includes "now". So, maybe just change it to 'Session GC (DB): remove expired records'.

alecpl (changes requested):

SQL/postgres.initial.sql
pabzm: Thanks for catching this! 👍

pabzm:
@alecpl This is the last code change planned for v1.7, could you please have another look? Or let me know when/how we can go forward with this? I requested a new review from you 2 weeks ago, is there some blocker?

teoberi:
I'm using 1.7 RC2 on the test server with PHP 8.5.0 and waiting for the final version for the production server.

program/lib/Roundcube/session/memcached.php
```diff
@@ -113,7 +113,7 @@ public function destroy($key)
     public function read($key)
     {
         if ($arr = $this->memcache->get($key)) {
-            $this->changed = $arr['changed'];
+            $this->expires_at = $arr['expires_at'];
```
alecpl: Ditto.

alecpl (changes requested):

program/lib/Roundcube/session/memcache.php
```diff
@@ -114,7 +114,7 @@ public function read($key)
     {
         if ($value = $this->memcache->get($key)) {
             $arr = unserialize($value);
-            $this->changed = $arr['changed'];
+            $this->expires_at = $arr['expires_at'];
```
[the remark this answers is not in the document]
pabzm: Good point, thank you!

alecpl (approved):
Looks good. Rebase on master to make CI green before merge, please

Issue #5050, opened 2012-09-10
Integration persistent login (rememberme)


rcubetrac:
_Reported by lefoyer on 10 Sep 2012 19:15 UTC as Trac ticket #1488685_

I tested the patch for version 0.8.1
Need a translation into languages other than English and Russian.

_Migrated-From: http://trac.roundcube.net/ticket/1488685_

Offerel:
This request is from 2012, are there any plans when this could be in release?

mkllnk:
There is still an issue that sessions do not survive browser restarts.

A recent commit to solve this problem does not work as intended, I guess:
https://github.com/roundcube/roundcubemail/blob/1bb97973b2667cd3ed0c6c6812c94e73edf0687d/program/lib/Roundcube/rcube.php#L518-L519

One option is the "remember me" checkbox as discussed here. The other option is to adjust the session cookie lifetime to RC's `session_lifetime` and let people use the Logout button when they want to.

mkllnk:
This is may not be for everyone, but here is what I changed to extend the sessions:
```patch
---
 program/lib/Roundcube/rcube.php         | 4 +++-
 program/lib/Roundcube/rcube_session.php | 5 ++++-
 2 files changed, 7 insertions(+), 2 deletions(-)

diff --git a/program/lib/Roundcube/rcube.php b/program/lib/Roundcube/rcube.php
index 2b9e3bd..f21de96 100644
--- a/program/lib/Roundcube/rcube.php
+++ b/program/lib/Roundcube/rcube.php
@@ -516,7 +516,9 @@ class rcube
         }

         // set session cookie lifetime so it never expires (#5961)
-        ini_set('session.cookie_lifetime', 0);
+        //ini_set('session.cookie_lifetime', 0);
+        // set cookie lifetime to one month
+        ini_set('session.cookie_lifetime', 60 * 60 * 24 * 30);
         ini_set('session.cookie_secure', $is_secure);
         ini_set('session.name', $sess_name ?: 'roundcube_sessid');
         ini_set('session.use_cookies', 1);
diff --git a/program/lib/Roundcube/rcube_session.php b/program/lib/Roundcube/rcube_session.php
index 9fc564f..bb56f3d 100644
--- a/program/lib/Roundcube/rcube_session.php
+++ b/program/lib/Roundcube/rcube_session.php
@@ -655,7 +655,10 @@ abstract class rcube_session
     public function set_auth_cookie()
     {
[...]

PR #9991, opened 2025-09-28
Extended session lifetime

pabzm:
aka Persisted Login plugin functionality in core code.

Allows admins to set `$config['extended_session_lifetime_days']`, which allows users to switch on an extended session
lifetime in the login form. In effect, these user sessions are valid for the configured number of days even across
network outages, closed browsers (as long as they keep their cookies), etc.

Based on #9990 

This would implement #5050

### DIFF
--- a/program/include/rcmail_output_html.php
+++ b/program/include/rcmail_output_html.php
@@ -2355,6 +2355,18 @@ class rcmail_output_html extends rcmail_output
             'buttons' => [],
         ];
 
+        if ($this->config->session_lifetime_extension_days() > 0) {
+            $session_lifetime_extension_hidden_field = new html_hiddenfield(['name' => '_session_lifetime_extension', 'value' => '0']);
+            $form_content['hidden']['session_lifetime_extension'] = $session_lifetime_extension_hidden_field->show();
+
+            // Make sure the value is in the range 1..365.
+            $session_lifetime_extension_text = str_replace('#', $this->config->session_lifetime_extension_days(), $this->app->gettext('session_lifetime_extension_switch_text'));
+            $session_lifetime_extension_checkbox = new html_checkbox(['name' => '_session_lifetime_extension', 'id' => '_session_lifetime_extension', 'title' => $session_lifetime_extension_text]);
+            $form_content['inputs']['session_lifetime_extension'] = [
+                'content' => html::label(['for' => '_session_lifetime_extension'], [$session_lifetime_extension_checkbox->show(), $session_lifetime_extension_text]),
+            ];
+        }
+
         if (is_array($default_host) && count($default_host) > 1) {
             $input_host = new html_select(['name' => '_host', 'id' => 'rcmloginhost', 'class' => 'custom-select']);
 
--- a/public_html/index.php
+++ b/public_html/index.php
@@ -123,8 +123,8 @@ if ($RCMAIL->task == 'login' && $RCMAIL->action == 'login') {
         $RCMAIL->session->remove('temp');
         $RCMAIL->session->regenerate_id(false);
 
-        // send auth cookie if necessary
-        $RCMAIL->session->set_auth_cookie();
+        $session_lifetime_extension = rcube_utils::get_input_string('_session_lifetime_extension', rcube_utils::INPUT_POST);
+        $RCMAIL->session->set_auth_cookie($session_lifetime_extension === 'on');
 
         // log successful login
         $RCMAIL->log_login();
--- a/skins/elastic/ui.js
+++ b/skins/elastic/ui.js
@@ -1160,6 +1160,11 @@ function rcube_elastic_ui() {
                     icon_name = input.data('icon'),
                     icon = $('<i>').attr('class', 'input-group-text icon ' + input.attr('name').replace('_', ''));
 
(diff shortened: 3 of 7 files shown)


program/lib/Roundcube/session/db.php
```diff
@@ -176,17 +176,17 @@ protected function save($key, $vars)
...
         // if new and old data are not the same, update data
         // else update expire timestamp only when certain conditions are met
         if ($newvars !== $oldvars) {
             $this->db->query("UPDATE {$this->table_name} "
-                . "SET `changed` = {$now}, `vars` = ? WHERE `sess_id` = ?",
+                . "SET `expires_at` = {$expires_at_str}, `vars` = ? WHERE `sess_id` = ?",
                 base64_encode($newvars), $key);
-        } elseif ($ts - $this->changed > $this->lifetime / 2) {
-            $this->db->query("UPDATE {$this->table_name} SET `changed` = {$now}"
+        } elseif ($this->expires_at - $ts > $this->lifetime / 2) {
```
alecpl, line 188: Considering lifetime can be quite a big interval now. Maybe forcing an update at least every hour would make sense.

On the other hand. Isn't "Remember login for up to # days" imply that the session expiration should be not auto-extended? Also, is "up to" needed here? I guess these are some general questions we need to think about.

program/lib/Roundcube/session/db.php
```diff
@@ -200,10 +200,10 @@ public function gc_db()
     {
         // just clean all old sessions when this GC is called
         $this->db->query('DELETE FROM ' . $this->db->table_name('session')
-            . ' WHERE `changed` < ' . $this->db->now(-$this->gc_enabled));
+            . ' WHERE `expires_at` < ' . $this->db->now());
 
         $this->log('Session GC (DB): remove records < '
-            . date('Y-m-d H:i:s', time() - $this->gc_enabled)
+            . date('Y-m-d H:i:s', $this->db->now())
```
alecpl, line 206: This is wrong.

program/lib/Roundcube/session/php.php
```diff
@@ -97,6 +97,10 @@ public function start()
 
         $this->key = session_id();
         $this->ip = $_SESSION['__IP'] ?? null;
-        $this->changed = $_SESSION['__MTIME'] ?? null;
+        if (is_int($_SESSION['__MTIME'])) {
```
alecpl, line 100: This reminds me of #9885. I'm not sure if that matters.

public_html/index.php
```diff
@@ -123,8 +123,19 @@
...
+
+        if ($extended_session_lifetime === 'on' && is_int($extended_session_lifetime_days) && $extended_session_lifetime_days > 0) {
+            // Make sure the value is in the range 1..365.
+            $days = min(max(1, $extended_session_lifetime_days), 365);
+            $RCMAIL->session->set_lifetime($days * 24 * 60 * 60);
+            $_SESSION['extended_session_lifetime'] = $days * 24 * 60 * 60;
+            $RCMAIL->session->set_auth_cookie();
+        } else {
+            // send auth cookie if necessary
+            $RCMAIL->session->set_auth_cookie();
```
alecpl, line 137: Maybe add an argument to `set_auth_cookie()` and do the whole logic inside the session handler.

skins/elastic/ui.js
```diff
@@ -1160,6 +1160,10 @@ function rcube_elastic_ui() {
                     icon_name = input.data('icon'),
                     icon = $('<i>').attr('class', 'input-group-text icon ' + input.attr('name').replace('_', ''));
 
+                if (input.hasClass('no-prettify')) {
```
alecpl, line 1163: Maybe just ignore checkboxes.

alecpl (changes requested):

program/include/rcmail_output_html.php
```diff
@@ -2355,6 +2355,20 @@ protected function login_form($attrib)
             'buttons' => [],
         ];
 
+        $extended_session_lifetime_days = $this->config->get('extended_session_lifetime_days', 0);
```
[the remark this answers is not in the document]
pabzm: I see the point, will change that.

pabzm:
Rebased onto #9990

pabzm:
Rebased again onto the latest changes from https://github.com/roundcube/roundcubemail/pull/9990

pabzm:
Squashed and rebased onto the latest state of the "master" branch.

skins/elastic/ui.js
```diff
@@ -1160,6 +1160,11 @@ function rcube_elastic_ui() {
                     icon_name = input.data('icon'),
                     icon = $('<i>').attr('class', 'input-group-text icon ' + input.attr('name').replace('_', ''));
 
+                // Ignore checkboxes, they are prettified well enough by pretty_checkbox() already.
+                if (input.attr('type') === 'checkbox') {
+                    return;
```
alecpl: If we return here the `tr` will not get "form-group row" class, which makes it looking better (adds margin). Also, it would be good to align this row content to the center, imo.

alecpl (changes requested):

alecpl (changes requested):

PR #10003, opened 2025-10-06
add left padding to dark.less

JohnRDOrazio:
In dark mode, the message body usually still has a white background, but there is no margin / padding between the surrounding dark interface and the white email body. This adds a slight padding to make the email slightly more visually pleasant.

Fixes issue #10002

### DIFF
--- a/skins/elastic/styles/dark.less
+++ b/skins/elastic/styles/dark.less
@@ -822,6 +822,7 @@ html.dark-mode {
         background-color: white;
         color: @color-font;
         padding: 0;
+        padding-left: .3rem;
         margin-top: .5rem;
     }
 


JohnRDOrazio:
Before the fix:
Screenshot roundcube before dark fix

After the fix:
Screenshot roundcube after dark fix

JohnRDOrazio:
it might actually be helpful to set the padding on all sides, substituting `padding: 0;` with `padding: 0.3rem;`, I've just seen some emails that have no padding on the top either.

alecpl:
Then maybe we should make the outer padding smaller, so it looks the same in both modes.

I'm not sure I like this solution anyway, but I don't have a better one.

JohnRDOrazio:
I'm not sure how we can change the outer padding without affecting other elements?

Screenshot roundcube

The area with the yellow squiggle will affect the headers too.

I think the idea behind the problem is that the message body should be considered as its own "body", and as any html body would not look right without some minimum padding, so the message body should contain some minimum padding. We could make the padding default for both light and dark mode, so that it shows the same in both cases. And perhaps it should apply to all message parts, whether html or not, for consistency.

PR #10007, opened 2025-10-08
enigma plugin: gpg redressing attack

### LINKED ISSUE #6450: enigma plugin: gpg redressing attack

JohnRDOrazio:
Fixes #6450

1) The `status_message` callback now only sets the private class array `status_messages`, rather than modify the message body.
2) Any status message boxes are now created in the `message_output` callback if the private class array `status_messages` is not empty.

### DIFF
--- a/plugins/enigma/lib/enigma_ui.php
+++ b/plugins/enigma/lib/enigma_ui.php
@@ -955,7 +955,6 @@ class enigma_ui
             && !empty($engine->decryptions[$found])
         ) {
             $status = $engine->decryptions[$found];
-            $attach_scripts = true;
 
             // show the message only once
             unset($engine->decryptions[$found]);
--- a/plugins/enigma/lib/enigma_ui.php
+++ b/plugins/enigma/lib/enigma_ui.php
@@ -1104,7 +1104,6 @@ class enigma_ui
         if ($count = count($this->status_messages)) {
             $attach_scripts = true;
             if ($count == 2 && $this->status_messages[0]['class'] == $this->status_messages[1]['class']) {
-                // @phpstan-ignore-next-line
                 $p['content'] = html::div($this->status_messages[0], $this->status_messages[0]['msg'] . ' ' . $this->status_messages[1]['msg']) . $p['content'];
             } else {
                 foreach ($this->status_messages as $msg) {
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
The results on my live Roundcube instance:
Screenshot roundcube PR10007

alecpl:
The point was that this place is not safe. Maybe in the dark mode (with white HTML content) you see separation, but normally you don't and the attacker could made the content look the same. So, I don't know if we can consider this a fix.

Also, a message can consist multiple parts, all or some encrypted/signed. How do you know which box is to which part?

JohnRDOrazio:
Perhaps adding an option to set a custom background pattern on the whole iframe body would address this issue? If a value is not set by the user in settings, then a random background pattern could be set. As a proof of concept, here is the result of setting:
```css
html.iframe body {
    background-image: url(http://cdn.backgroundhost.com/backgrounds/subtlepatterns/arab_tile.png);
    background-blend-mode: color-burn;
}
```
Screenshot roundcube with background

I believe that would make redressing much more difficult.

alecpl:
That's not much different than https://github.com/roundcube/roundcubemail/issues/6450#issuecomment-493433433

JohnRDOrazio:
I guess that makes sense. Whether you set a custom background pattern on the whole mail preview with a clear separation between the enigma message boxes and the message body, or you set a custom background image on the enigma message boxes themselves, either approach pretty much solves the issue of any redress attacks.

Perhaps the only advantage to setting a custom background pattern on the whole mail preview, with a clear separation between the enigma message boxes and the message body, would be to further "sandbox" the message body. But one way or the other both solutions help address the issue.

PR #10008, opened 2025-10-09
Prevent duplicated job runs on pull requests

pabzm:
Another proposal for the same goal.

### DIFF
--- a/.github/workflows/browser_tests.yml
+++ b/.github/workflows/browser_tests.yml
@@ -10,7 +10,7 @@ permissions:
 jobs:
   browser_tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


pabzm:
@mvorisek @alecpl Does this work for you? (In alternative to https://github.com/roundcube/roundcubemail/pull/9997)

alecpl:
Maybe it would be fine. There are two places in tests.yml that need to change, not one.

pabzm:
> Maybe it would be fine.

I'm not really sure if that's an OK from your side. Would you merge it, if you can live with this solution? Or let me know what blocks it?

alecpl:
> > Maybe it would be fine.
> 
> I'm not really sure if that's an OK from your side. Would you merge it, if you can live with this solution? Or let me know what blocks it?

I guess, if you don't want to use a fork for your PRs, I'll have to live with this.

PR #10012, opened 2025-10-09
Add rel='noopener' to all links opening in a new window

pabzm:
Browsers younger than ~5 years don't need this, but older browsers might cause problems.

### DIFF
--- a/plugins/help/help.php
+++ b/plugins/help/help.php
@@ -162,7 +162,7 @@ class help extends rcube_plugin
             && $rcmail->request_status == rcube::REQUEST_ERROR_URL
             && ($url = $rcmail->config->get('help_csrf_info'))
         ) {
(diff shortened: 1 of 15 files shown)


pabzm:
Here's a start. I'm not sure if I found all the relevant bits, and would be happy for pointers (or corrections).

alecpl:
Do we have to care about such old browsers? Such users should not expect security from others if they are using outdated software themselves.

ps. your changes collide with rel=noreferrer in a few places.

pabzm:
In principle we don't have to, but it's not much work and in other regards we care to support old versions, e.g. PHP v8.1, which is out of active support since Nov 2023. So why not just do it?

pabzm:
I removed the `rel`-duplicate and fixed the tests.

@alecpl Please have another look.

pabzm:
I'm merging this now on my own because it's waiting for quite a while, even though it's just a small change, and doesn't hurt but helps with older browsers.

PR #10017, opened 2025-10-15
allow "target" in html attributes when saving signature

### LINKED ISSUE #10005: "Open in new window" not working for link in signature
### Prerequisites

### Describe the issue

Hi, inserting a link in a signature via the WYSIWYG editor gives the option "Open link in...". If you set this to "Open in new window", save the link and show the source code before saving - it does indeed have `target="_blank"`.

If you then save the signature and reopen the source code, the `target` is removed.

Expected: `target="_blank"` does not get removed on saving the signature.

Possible cause: 
* saving identity calls `wash_html` on the signature: https://github.com/roundcube/roundcubemail/blob/0abdccaf555167da61c0976bd41ad2460fe73f51/program/actions/settings/identity_save.php#L106
* wash_html not allowing `target`: https://github.com/roundcube/roundcubemail/blob/0abdccaf555167da61c0976bd41ad2460fe73f51/program/actions/settings/index.php#L1809

After adding `target` to `html_attribs` it would work, however this may have other sideeffects.

Chrome

8.1

1.6.11

### JavaScript errors

_No response_

### PHP errors

_No response_

codegain:
Allow the `target` attribute in HTML when saving the signature.
Sidenote: this also affects saving responses.

Fixes #10005

### DIFF
--- a/program/actions/settings/index.php
+++ b/program/actions/settings/index.php
@@ -1806,7 +1806,7 @@ class rcmail_action_settings_index extends rcmail_action
             'charset' => RCUBE_CHARSET,
             'html_elements' => ['body', 'link'],
             'ignore_elements' => ['body'],
-            'html_attribs' => ['rel', 'type'],
+            'html_attribs' => ['rel', 'type', 'target'],
             'add_comments' => false,
         ];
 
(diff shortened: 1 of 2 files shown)


pabzm:
Thank you very much for the contribution, highly appreciated!

From my point of view this does no harm and could be merged as it is. 👍

pabzm (approved):

PR #10033, opened 2025-11-17
feat: preserve requested url on oidc login

dschmidt:
Currently, users are always redirected to the inbox after OAuth login, regardless of what page they originally requested. This breaks external application integrations (like external applications opening compose windows) and bookmarks.

This PR preserves the originally requested URL through the OAuth flow by storing it in the session before redirect, preserving it through the login session kill via `login_phase`, and restoring the parameters in the `login_after` hook.

Additionally, this fixes the `url()` method in `rcmail.php` to handle array parameters correctly, which was causing fatal errors when plugins or external apps passed array values like file attachments.

This should be fully backwards compatible with no configuration changes required.


dschmidt:
CI is green :)

PR #10034, opened 2025-11-20
Document TLS CA certificate for LDAP connectivity

domibarton:
Check out [this issue](https://github.com/roundcube/roundcubemail-docker/issues/382) for more information regarding this topic.

### DIFF
--- a/plugins/password/README
+++ b/plugins/password/README
@@ -194,6 +194,19 @@
 
  See config.inc.php.dist file. Requires PEAR::Net_LDAP2 package.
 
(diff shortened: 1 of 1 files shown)


alecpl:
All right, but the same could be said about any Roundcube ldap connection, e.g. for addressbook. So, documenting this in the password plugin may not be the best place. Maybe, since we have the docs/ directory, we should create LDAP.md file and put everything LDAP related there.

pabzm:
I think we can do both. Or put the details in a standalone document and reference it in the plugin's README (and/or in its `config.inc.php.dist`, because that explicitly mentions LDAP with TLS)

PR #10038, opened 2025-11-30
feat: extend Dovecot passwdfile driver with dynamic file path support

### LINKED ISSUE #10036: dovecot_passwdfile to support users file per domain
### Prerequisites

### Proposal

As operator of Dovecot and Roundcube, I want to be able to use runtime interpolation in `password_dovecot_passwdfile_path`, so I can utilize Dovecot's feature of Multiple passwd-files.

### Motivation and context

Dovecot has a feature to use [Multiple passwd-files](https://doc.dovecot.org/2.4.2/core/config/auth/databases/passwd_file.html#multiple-passwd-files) - the documentation includes an example for one file per email domain, but this way Dovecot may support even a file per user.

MikkCZ:
fix #10036

### DIFF
--- a/plugins/password/drivers/dovecot_passwdfile.php
+++ b/plugins/password/drivers/dovecot_passwdfile.php
@@ -39,12 +39,11 @@ class rcube_dovecot_passwdfile_password
      *
      * roundcube already validated the old password so we just need to change it at this point
      *
-     * @param string $curpass Current password
-     * @param string $newpass New password
+     * @param string $currpass Current password
+     * @param string $newpass  New password
      * @param string $username Login username (configured form based on $config['password_username_format'])
      *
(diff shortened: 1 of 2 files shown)


plugins/password/drivers/dovecot_passwdfile.php
```diff
@@ -74,8 +92,28 @@ public function save($currpass, $newpass, $username)
...
 
         return PASSWORD_ERROR;
     }
+
+    private static function _expand_config_value(string $subject, string $local_part, string $domain_part, string $username): string
+    {
+        return strtr($subject, [
+            '%l' => $local_part,
+            '%d' => $domain_part,
+            '%u' => $username,
```
alecpl: I don't think %u and %l make sense here, but %d and %h would.

PR #10050, opened 2025-12-10
Fix table name in SQL migration

### LINKED ISSUE #10047: typo in latest sql update files
https://github.com/roundcube/roundcubemail/blob/202daa6f97bb1dca2f5f1a5485bb9a928d47e9e6/SQL/mysql/2025092300.sql#L3

should be

```sql
UPDATE `session` SET expires_at = ADDTIME(expires_at, '00:10:00'); 
```

There is no such table as `sessions`. Same in the `postgres` and `sqlite` update files.

pabzm:
Fixes #10047

### DIFF
--- a/SQL/sqlite/2025092300.sql
+++ b/SQL/sqlite/2025092300.sql
@@ -1,3 +1,4 @@
 ALTER TABLE `session` RENAME COLUMN `changed` TO `expires_at`;
-ALTER TABLE `session` RENAME INDEX `ix_session_changed` TO `ix_session_expires_at`;
+DROP INDEX ix_session_changed;
+CREATE INDEX ix_session_expires_at ON session ("expires_at");
 UPDATE session SET expires_at = DATETIME(expires_at, '+10 minutes');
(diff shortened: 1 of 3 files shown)


SQL/sqlite/2025092300.sql
```diff
@@ -1,3 +1,4 @@
 ALTER TABLE `session` RENAME COLUMN `changed` TO `expires_at`;
-ALTER TABLE `session` RENAME INDEX `ix_session_changed` TO `ix_session_expires_at`;
-UPDATE sessions SET expires_at = DATETIME(expires_at, '+10 minutes');
+DROP INDEX ix_session_changed;
+CREATE INDEX ix_session_expires_at ON session ("expires_at");
+UPDATE session SET expires_at = DATETIME(expires_at, '+10 minutes');
```
alecpl: The UPDATE should probably be before CREATE INDEX, for better performance, but it's not a big deal.

alecpl (approved):

PR #10070, opened 2026-01-08
 Fix syntax error in DDL scripts for Postgres (#10052)

llamaonaskateboard:
The previous commit https://github.com/roundcube/roundcubemail/commit/366ad7a1745eb7913536fb937924d7725257d633 didn't actually fix the incorrect index rename syntax.


alecpl:
Indeed, ALTER TABLE ... RENAME INDEX does not exist in Postgres.

PR #10072, opened 2026-01-09
Update messages.inc lv_LV fix spelling

urbels:
I also added these as suggestions on transifex because I cant save correct translation.


highlensprolens-eng (approved):

alecpl:
We manage localizations at transifex.com. I applied the fixes so we don't overwrite this next time we sync localization.

urbels:
Thank you!

Issue #10074, opened 2026-01-16
Vacations can't handle dates in brackets


Abyss777:
### Prerequisites

- [x] I have [searched](https://github.com/roundcube/roundcubemail/issues?q=is%3Aissue) for duplicate or closed issues
- [ ] I can recreate the issue with all plugins disabled

### Describe the issue

Hello
I programmatically generate Vacation outside the roundcube via managesieve, but still want to be able edit it in roundcube.

Here is example of generated sieve script

```
require ["date", "relational", "vacation"];

# rule:[Вне офиса]
if allof (currentdate :zone "+0500" :value "ge" "date" ["2026-01-01"], currentdate :zone "+0500" :value "le" "date" ["2026-01-15"]) {
    vacation :subject "Отпуск" :days 1 :from "<s09@example.ru>" :addresses "s09@example.ru" "В отпуске";
}
```

The dates are parsed normally in filters

<img width="1558" height="806" alt="Image" src="https://github.com/user-attachments/assets/8e2bd504-3560-44c2-acba-1e39da8e1201" />

But can't be parsed in Vacation tab

<img width="1558" height="806" alt="Image" src="https://github.com/user-attachments/assets/f5fb759f-5d09-4317-bc3f-daf0d75184e6" />

I believe that behavior should be consistent and dates are parsed correctly in both places

### What browser(s) are you seeing the problem on?

Firefox

### What version of PHP are you using?

8.1.2

### What version of Roundcube are you using?

1.6.12

### PHP errors

Jan 16 08:30:01 mail roundcube: PHP Warning:  Array to string conversion in /var/www/roundcube/program/lib/Roundcube/rcube_utils.php on line 1136
Jan 16 08:30:01 mail roundcube: PHP Warning:  Array to string conversion in /var/www/roundcube/program/lib/Roundcube/rcube_utils.php on line 1136

Abyss777:
According to [RFC5260](https://datatracker.ietf.org/doc/html/rfc5260#section-5), last argument has a `string-list` type. Because of that Vacation also should support date in brackets even if there is only one.

alecpl:
Fixed.

PR #10075, opened 2026-01-24
managesieve: handle string-list format values for date test

johndoh:
possible fix for #10074 by forcing the date argument to always be a string.


johndoh:
failed test refers to `program/actions/mail/folder_purge.php` and does not appear to be related to this PR

alecpl:
This is not a proper fix because we have support for string-list in the general filter UI. We only needed a fix for the `rcube_sieve_vacation` class. Fixed in 719b8be25956c4.

There's another issue you could take a look at - #10076

PR #10083, opened 2026-02-03
Added support for arrays for smtp_user and smtp_pass config options

jce-zz:
With Roundcube 1.6 the possibility was added to have more than one SMTP host, i.e. define an array of SMTP hosts, which get selected based on the IMAP host used for the logged in user. However, one could still only define a single smtp_user and smtp_pass or use the IMAP user and password (%u, %p) as credentials. 

For many scenarios, this is not sufficient and therefore I added the same mechanism that is used to select the SMTP host based on the IMAP host to select the SMTP user and SMTP password both defined in arrays similar to SMTP host.

### DIFF
--- a/program/lib/Roundcube/rcube_smtp.php
+++ b/program/lib/Roundcube/rcube_smtp.php
@@ -70,11 +70,39 @@ class rcube_smtp
 
         $host = rcube_utils::parse_host($host);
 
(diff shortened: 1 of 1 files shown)


dianecloud (approved):
Approved

alecpl (changes requested):

PR #10100, opened 2026-02-24
Fix double charset conversion for non-UTF-8 message bodies

ju-ef:
## Problem
Messages with charset=windows-1251 (quoted-printable encoding) display as garbled text in Roundcube. Other clients (Thunderbird, K-9, mutt) display them correctly.

## Root Cause
`get_message_part()` already converts body charset to UTF-8, but `format_part_body()` converts it again using `$part->charset` which still contains the original charset (e.g. windows-1251). This double conversion produces garbled output.

## Fix
Check `mb_check_encoding()` before converting — if body is already valid UTF-8, skip conversion.

## Tested
Roundcube 1.6.13 + Stalwart mail server, windows-1251 quoted-printable messages now display correctly.

## Related
Similar UTF-8 detection approach used in #10078.

### DIFF
--- a/program/lib/Roundcube/rcube_message.php
+++ b/program/lib/Roundcube/rcube_message.php
@@ -349,7 +349,9 @@ class rcube_message
         }
 
         // ..convert charset encoding
-        $body = rcube_charset::convert($body, $part->charset);
+        if (!mb_check_encoding($body, 'UTF-8')) {
+            $body = rcube_charset::convert($body, $part->charset);
+        }
 
         return $body;
     }


alecpl:
I'm not a fun of checking encoding. We should investigate why we do convertion twice, and why more people don't complain. Could you provide a sample message?

ju-ef:
I can reproduce it via swaks.
1. generate body:
```
echo "привет мир" | iconv -t cp1251 | python3 -c "import quopri, sys; sys.stdout.buffer.write(quopri.encodestring(sys.stdin.buffer.read()))"
```
2. send via swaks:
```
swaks -tls -p 587 --auth LOGIN -ap '<password>' -au '<login>' --server '<server>' \
  --to '<to@domain.tld>' --from '<from@domain.tld>' \
  --header "Content-Type: text/plain; charset=windows-1251" \
  --header "Content-Transfer-Encoding: quoted-printable" \
  --body "=EF=F0=E8=E2=E5=F2 =EC=E8=F0"
```

OK -
mutt, K-9 (stalwart mail server);
Gmail;
raven (wildduck mail server);
SnappyMail (postfix, dovecot);
Roundcube 1.6.0, 1.6.1, 1.6.13 (postfix, dovecot);

Not OK -
Roundcube 1.6.13 (stalwart)

So, it's something wrong with [stalwart](https://github.com/stalwartlabs/stalwart/).
But should be there some fix? as mutt, K-9 work fine


[test2.eml](https://github.com/user-attachments/files/25550838/test2.eml)
roundcube_inbox

alecpl:
Other clients might use different set of imap commands, some of them just download the mail source and parse it themselves, but Roundcube uses e.g. BODYSTRUCTURE response from the IMAP server. If such a response is invalid...

It works for me also with Cyrus IMAP. So, it's obviously a Stalwart issue. Enable imap_debug to see what's going on. You can also try with `$config['imap_disabled_caps'] = ['BINARY'];`.

ju-ef:
> `$config['imap_disabled_caps'] = ['BINARY'];`

yes, it fixes it

thank you

Stalwart issue: https://github.com/stalwartlabs/stalwart/discussions/2882

PR #10114, opened 2026-03-11
stalwart password driver added

avignat:
This code add the ability to change your password when using [stalw.art](https://stalw.art/) mail server.


alecpl:
Please add new sections to the README and config.inc.php.dist files.

PR #10117, opened 2026-03-13
Validate _filter parameter against known IMAP search keywords

azizelbelaychy:
## Validate _filter parameter against known IMAP search keywords

 ### Summary

 The `_filter` GET parameter in the mail search action (program/actions/mail/search.php) is read from user input and passed into the IMAP SEARCH command string without any validation of its value. The only sanitization applied is strip_tags() and trim(), neither of which enforces that the value is a legitimate IMAP search keyword.

### Problem

`$filter` is appended directly to the IMAP query in search_input():

$result = $filter && $filter != 'ALL' ? $filter : '';

There is no check that `$filter` is one of the expected values (UNSEEN, SEEN, FLAGGED, etc.). This means any arbitrary string that passes strip_tags() is concatenated verbatim into the IMAP protocol stream, which could cause the IMAP server to misinterpret the query or behave unexpectedly depending on its implementation.

### Fix

Validate `$filter` against an explicit allowlist of the IMAP search keywords that the Roundcube UI legitimately produces before it is used in the query:

```php
$allowed_filters = [
      'ALL', 'UNSEEN', 'SEEN', 'FLAGGED', 'UNFLAGGED',
      'DELETED', 'UNDELETED', 'ANSWERED', 'UNANSWERED', 'RECENT',
  ];

if ($filter && !in_array(strtoupper($filter), $allowed_filters)) {
      $filter = '';
 }
```
Any value outside this set is discarded. Valid UI-produced filter values continue to work unchanged.

### Tests

Test cases added to tests/Actions/Mail/SearchTest.php covering:
  - Invalid/unexpected filter values → discarded
  - Valid filter keywords (UNSEEN, FLAGGED) → pass through unchanged

### DIFF
--- a/program/actions/mail/search.php
+++ b/program/actions/mail/search.php
@@ -214,6 +214,18 @@ class rcmail_action_mail_search extends rcmail_action_mail_index
     {
         $headers = $headers ? explode(',', $headers) : ['subject'];
 
+        // Validate $filter against the known set of IMAP search keywords.
+        // Any value outside this list (including CRLF-injected payloads) is
+        // silently discarded, preventing IMAP command injection via _filter.
+        $allowed_filters = [
+            'ALL', 'UNSEEN', 'SEEN', 'FLAGGED', 'UNFLAGGED',
+            'DELETED', 'UNDELETED', 'ANSWERED', 'UNANSWERED', 'RECENT',
+        ];
+
+        if ($filter && !in_array(strtoupper($filter), $allowed_filters)) {
+            $filter = '';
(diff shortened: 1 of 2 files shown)


alecpl:
We're aware of the issue and plan a new release next week.

The allowed cases list is incomplete. I think we'll fix the issue in another way, e.g. by forbidding `\n` character in the input.

alecpl:
Fix included in release 1.6.14, 1.5.14 and 1.7-rc5.

PR #10137, opened 2026-04-04
Improve Esperanto locale

James-Seely-scln:
Added new translations and updated existing ones.


alecpl:
This needs to be done via Transifex https://app.transifex.com/roundcube/roundcube-webmail/language/eo/

PR #10085, opened 2026-02-05
feat: Add VS Code DevContainer for simplified development setup

JohnRDOrazio:
## Summary

- Add complete VS Code DevContainer configuration for one-click development environment setup
- Includes PHP 8.4, MariaDB, Dovecot IMAP, and Mailhog services
- Automatic database initialization, dependency installation, and configuration
- Pre-configured VS Code tasks for testing, linting, and development server
- Fix php-cs-fixer to exclude actual shell scripts while keeping PHP CLI scripts with `.sh` extension
- Add PHPUnit result cache to `.gitignore`

Closes #10084

## Test plan

- [x] Open repository in VS Code with Dev Containers extension installed
- [x] Click "Reopen in Container" when prompted
- [x] Verify all services start (MariaDB, Dovecot, Mailhog)
- [x] Verify database is automatically initialized
- [x] Start dev server with `Ctrl+Shift+B` and verify Roundcube loads
- [x] Login with `testuser`/`testpass` credentials
- [x] Run PHPUnit tests via VS Code task
- [x] Run php-cs-fixer and verify it doesn't fail on shell scripts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10144, opened 2026-04-09
Fix static.php RFC 7233 range handling and built-in server compatibility

### LINKED ISSUE #10143: static.php: RFC 7233 suffix-range bug and built-in server compatibility issues
## Summary

`static.php` has several bugs in its range request handling and file serving logic that cause incorrect responses and potential corruption on the PHP built-in server.

### 1. Suffix-range requests return wrong bytes (RFC 7233 violation)

Per [RFC 7233 Section 2.1](https://datatracker.ietf.org/doc/html/rfc7233#section-2.1), `bytes=-500` means "the last 500 bytes". The current code splits on `-` to get `["", "500"]`, then sets the empty first element to `0`:

```php
if ($range[0] === '') {
    $range[0] = 0;
}
```

This turns `bytes=-500` into `bytes=0-500`, serving the **first 501 bytes** instead of the **last 500 bytes**. The `Content-Range` header is also wrong (`bytes 0-500/1058` instead of `bytes 558-1057/1058`).

### 2. Suffix-range larger than file returns 416 instead of 206

Per RFC 7233: "If the selected representation is shorter than the specified suffix-length, the entire representation is used."

`bytes=-9999` on a 1058-byte file is interpreted as `bytes=0-9999`. The bounds check `$range[1] <= $size - 1` fails (9999 > 1057), returning 416 "Range Not Satisfiable" instead of serving the full file.

### 3. Headers sent before file handle validation

`http_response_code(206)` and response headers are sent before `fopen()` is called. If `fopen()` fails, the 500 error code cannot take effect because headers were already sent. The client receives a 206 with the promised `Content-Length` but an empty body.

### 4. Output buffer corruption on PHP built-in server

When output buffering is active (e.g., `php.ini output_buffering=On`), buffer content is prepended to the file output. `Content-Length` is calculated from `filesize()` and doesn't account for the buffer, so HTTP clients read `Content-Length` bytes starting with the buffer junk, truncating the end of the actual file. This causes broken CSS/JS delivery.

## Steps to reproduce

Suffix-range bug (using any HTTP client):
```
curl -H "Range: bytes=-500" http://localhost/static.php/program/resources/dummy.pdf -o /dev/null -D -
```
The `Content-Range` header will show `bytes 0-500/1058` instead of the expected `bytes 558-1057/1058`.

## Environment

- Roundcube: master branch (1.7-beta / 1.7-rc)
- PHP: 8.1+

JohnRDOrazio:
## Summary

Fixes several bugs in `static.php`'s range request handling and file serving logic:

- **Fix suffix-range requests** (`bytes=-500`): was serving bytes 0-500 (first 501 bytes) instead of the last 500 bytes per RFC 7233 §2.1
- **Fix suffix-range larger than file** (`bytes=-9999`): was returning 416 instead of clamping to the full file per RFC 7233
- **Fix header/fopen ordering**: response headers were sent before `fopen()`, making it impossible to return 500 on failure
- **Add `flush()` to range request loop** for PHP built-in server consistency
- **Clean output buffers** before serving to prevent Content-Length mismatch
- **Use chunked `fread()`+`flush()`** on cli-server SAPI for reliable CSS/JS delivery

Closes #10143

Split out from #10085 per maintainer request.

## Test plan

- [x] New test fixture (`tests/fixtures/static_original_router.php`) runs the **original unfixed code** on a second PHP built-in server
- [x] `testOriginalSuffixRangeReturnsWrongBytes` — proves old code returns wrong bytes, new code returns correct bytes
- [x] `testOriginalSuffixRangeLargerThanFileReturns416` — proves old code returns 416, new code returns 206
- [x] `testOriginalOutputBufferCorruptsResponse` — proves output buffering corrupts responses without `ob_end_clean()`
- [x] `testCssFileIntegrity` / `testJsFileIntegrity` — byte-for-byte verification of CSS/JS serving on built-in server
- [x] `testContentLengthAccuracy` — parameterized across file sizes (54B to 387KB)
- [x] `testRangeHeaderSuffix`, `testRangeHeaderOpenEnded`, `testRangeHeaderFullFileViaRange` — RFC 7233 compliance
- [x] Existing tests preserved: `testRangeHeaderInvalid`, `testRangeHeaderStandard`, `testModifiedSinceHeader`, `testExistingResources`, `testForbiddenResources`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


public_html/static.php
```diff
@@ -140,76 +138,103 @@ function serveStaticFile($path): void
...
+            }
+            echo $chunk;
+            flush();
+        }
+        fclose($fp);
+    } else {
+        foreach ($headers as $k => $v) {
+            header("{$k}: {$v}", true);
+        }
+        readfile($path);
```
alecpl: There's no need for two code paths, one with fread, other with readfile. Isn't it?

tests/fixtures/static_original_router.php
```diff
@@ -0,0 +1,209 @@
+<?php
```
alecpl: Please, remove this fixture. All we want is tests for the code after the change.

alecpl (changes requested):

public_html/static.php
JohnRDOrazio: You're right, the SAPI conditional was leftover scaffolding from debugging built-in server flush behavior; there's no reason production code paths should care which SAPI is running. I've unified serveStaticFile() into a single path: range and full-file requests now share the same fopen → headers → chunked fread+flush loop, with $start/$end initialized to the full-file range (0 to $size - 1) and overridden only when an HTTP_RANGE header is present. The readfile() branch and the PHP_SAPI === 'cli-server' check are gone.

All 33 tests in StaticTest still pass. Fixed in commit [ff608b8](https://github.com/roundcube/roundcubemail/pull/10144/commits/ff608b89d8555accc57ddd178ad4f1148d02ff4a)

JohnRDOrazio:
I took a look at the three failing checks, they don't seem related to this PR:

The last Unit workflow run before the current one was 3 days ago (April 28, on master) and succeeded with the same CI config. Between then and now (May 1), the ondrej/php PPA's GPG signing setup broke the way setup-php consumes it — the apt step now fails with NO_PUBKEY 71DAEAAB4AD4CAB6 NO_PUBKEY 4F4EA0AAE5267A6C before any PHP code runs.
  - All passing jobs use either Windows or PHP ≤ 8.3 (which don't pull from ondrej's PPA in
  setup-php's pipeline)
  - All failing jobs are exactly the ones that need ondrej for PHP 8.4/8.5 on Linux
  - Failure happens during apt-get update, not during phpunit/phpstan/php-cs-fixer
  - Coding Style and Static Analysis jobs (which also run on Linux but use a single PHP
  version) pass

This will resolve when either (a) ondrej's PPA gets re-signed, (b) setup-php ships a fix, or (c) a maintainer re-runs the workflow once upstream is healthy.


UPDATE: shivammathur says [this should be solved now](https://github.com/shivammathur/setup-php/issues/1078#issuecomment-4360917708), the next trigger of the CI actions should no longer present the problem

tests/fixtures/static_original_router.php
JohnRDOrazio: Agreed, the side-by-side fixture is academic once the bug is fixed, no need to keep it for the future.
Removed in f756487ff:
 - tests/fixtures/static_original_router.php (the frozen copy of the old logic)
 - testOriginalSuffixRangeReturnsWrongBytes, testOriginalSuffixRangeLargerThanFileReturns416, testOriginalOutputBufferCorruptsResponse (the three comparison tests)
 - requestOriginal() helper, the second-server setUpBeforeClass/tearDownAfterClass overrides, and the Symfony\Component\Process\Process import

Test count went from 33 (125 assertions) to 30 (107 assertions); all pass.

PR #10151, opened 2026-04-24
Zipdownload: fix escaping in Mbox downloads

### LINKED ISSUE #10150: Zipdownload's Mbox format does not escape "From " lines
### Prerequisites

- [ ] I can recreate the issue with all plugins disabled

#### Steps to reproduce

Use a tool which does not perform RFC 3676 space stuffing (e.g. Postfix `sendmail`) to send email containing a "From " line in its body to a Roundcube-accessible mailbox:

```sh
$ TO=user@example.net; printf \
'From: user@example.com
To: %s
Subject: Test
Date: %s

From all of us at Hormel, Happy National Spam Musubi Day!
' $TO "$(date -R)" | sendmail $TO
```

In Roundcube, select this email and at least one more, and choose `Export -> Mbox format`.

#### Expected results

A `.zip` file containing the escaped line:

    >From all of us at Hormel, Happy National Spam Musubi Day!

#### Actual results

A `.zip` file containing the original line verbatim.

#### Cause

This appears to be a regression from 74b89ff. Removing it results in the expected results. Note also the comment below which is no longer true following that commit:

https://github.com/roundcube/roundcubemail/blob/830ab574b38ea63f3c8f7a66b62343c8005bd7a0/plugins/zipdownload/zipdownload.php#L404-L405

Chrome

v8.3

v1.7-rc6

### JavaScript errors

_No response_

### PHP errors

_No response_

gurnec:
Closes #10150.

This unfortunately adds some extra complication... a simpler alternative might be slurp each entire email into a variable and just `preg_replace()` it all at once in `_download_messages()`. That would add additional memory overhead though.

### DIFF
--- a/plugins/zipdownload/tests/Browser/MailTest.php
+++ b/plugins/zipdownload/tests/Browser/MailTest.php
@@ -65,7 +65,10 @@ class MailTest extends TestCase
                     $files = $this->getFilesFromZip($browser, $filename);
                     $browser->removeDownloadedFile($filename);
 
-                    $this->assertSame(['INBOX.mbox'], $files);
+                    $this->assertSame(['INBOX.mbox'], array_keys($files));
+                    $this->assertStringStartsWith('From test-from', array_first($files));
+                    $this->assertStringContainsString("\nFrom thomas", array_first($files));
+                    $this->assertStringContainsString("\n>From line which needs to be escaped", array_first($files));
                 });
 
             // Test More > Download > Maildir format (two messages selected)
@@ -82,6 +85,11 @@ class MailTest extends TestCase
                     $files = $this->getFilesFromZip($browser, $filename);
                     $browser->removeDownloadedFile($filename);
                     $this->assertCount(2, $files);
+                    $this->assertStringContainsString(' Test ', array_key_first($files));
+                    $this->assertStringEndsWith('.eml', array_key_first($files));
+                    $this->assertStringEndsWith(' Lines.eml', array_key_last($files));
+                    $this->assertStringContainsString('Attached image:', array_first($files));
+                    $this->assertStringContainsString("\nFrom line which needs to be escaped in mbox format.", array_last($files));
                 });
 
             // Test attachments download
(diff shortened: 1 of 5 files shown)


plugins/zipdownload/tests/Browser/MailTest.php
```diff
@@ -62,10 +62,13 @@ public function testMailUI()
...
-                    $files = $this->getFilesFromZip($browser, $filename);
-                    $browser->removeDownloadedFile($filename);
-
-                    $this->assertSame(['INBOX.mbox'], $files);
+                    try {
+                        $this->checkFilesInZip($browser, $filename, [
+                            'INBOX.mbox' => ['From test-from', "\nFrom thomas", "\n>From line which needs to be escaped"],
+                        ]);
+                    } finally {
+                        $browser->removeDownloadedFile($filename);
```
alecpl: I would prefer doing assertions after removing the file. The test files are small so we can keep them in memory.

gurnec: Do you mean something like:

 1. extract filenames / file contents,
 2. remove zip file (no try/finally block required),
 3. do all assertions

probably all in `checkFilesInZip()`?

Also, just for future reference, would you prefer that I add new commits to this branch, or force-push existing commit(s) to keep the history a bit cleaner?

alecpl: >     1. extract filenames / file contents,
>     2. remove zip file (no try/finally block required),
>     3. do all assertions
> 
> probably all in `checkFilesInZip()`?

Yes, not necessarily in a single method.

> Also, just for future reference, would you prefer that I add new commits to this branch, or force-push existing commit(s) to keep the history a bit cleaner?

Doesn't matter, I usually use the "Squash and merge" button.

gurnec: I've simplified the E2E tests w/o changing what is tested, should be better now.

MiMoHo:
I did an independent review and test of this PR (AI-assisted, all results verified locally):

**Functionality**
- All 8 `MboxFilterTest` unit tests pass locally on PHP 8.5.8.
- Fuzz-tested the filter beyond the unit tests: 2,000 randomly generated documents built from adversarial tokens (`From `, `>From`, bare `Fro`/`om `, CRLF endings, `>` runs, …), written through the filter in random 16–64 byte chunks, and compared byte-for-byte against a whole-string reference implementation (`preg_replace('/(?:^|(?<=\n))(>*From )/', '>\1', $data)`) — **zero mismatches**, including From-lines split across bucket boundaries and the closing flush.
- Deterministic split cases verified as well: boundary between `\r` and `\n`, split inside `From`, split inside a `>` chain, split exactly after `\n`, first-line escaping, and an unterminated trailing `>From` at EOF — all correct.

**The three failing CI checks are unrelated to this PR:**
- *Coding Style*: the failure flags `fully_qualified_strict_types` in files this PR doesn't touch (`installer/test.php`, `plugins/new_user_dialog/…`, …) — a toolchain drift on master at the time of the run (2026-06-19); CI on master is green again since (e.g. the runs of 2026-07-04/05).
- *Linux / PHP 8.4 (message rendering)*: transient network error pulling the docker image from ghcr.io (`Client.Timeout exceeded while awaiting headers`).
- *Windows / PHP 8.1*: transient composer download failure (`file could not be written … Permission denied`).

So a rebase onto current master (or re-running the checks) should turn CI green; functionally this looks ready from my side.

Issue #6251, opened 2018-04-17
Compose/reply with elastic skin


grape6:
May be it's good to have <cancel> button as in larry/classic skin.

grape6:
I mean that in larry/classic skin we have Cancel button to return from compose/reply screen to mail view.
sorry, seems editor eats some words.

grape6:
Certainly i can. 
But for user just going from larry to elasict it's a bit confuse.
And on iPad for example there is a large SAVE button on the bottom, and no clear way to return back.

grape6:
And on iPhone SE when SAVE & SEND buttons are visible, MAIL button in taskmenu is under address bar.

thomascube:
I'd like to upvote this issue. Cancel and back buttons should come back to Elastic.

mufus:
Just having installed 1.4-rc2, the Cancel button on mail-compose is my biggest worry for upgrading from 1.3. Please consider adding the Cancel button back for mail-compose page.

rdacn:
Just for the record, we are new users of RoundCube and so have used only the elastic skin.  Still, many users complain about the lack of cancel button in mail compose. They find it very confusing. 

So in our case, the problem is not about a button that used to exists and is not there anymore. It's about a button that users seem to think should be there. Clicking on "Mail" is apparently not an intuitive step for most of them.

Ptrph:
I’ve also some feedback from users who do not find their way to cancel a new mail.

adam12b1:
+1

It's baffling to users, even very savvy ones, coming from any other mail app (or honestly any app of any sort!) to not have a Cancel button on the mail composition page. Any reason not to add one?

alecpl:
@pabzm maybe you'd like to take this one?

Issue #5937, opened 2017-09-05
Send a reply and archive in one action button


marclaporte:
Use case: I receive an email. I respond, and from my point of view, it's the end of the discussion. I want to keep in my archives, but no longer have in the inbox.

Now, I need to reply to the email by clicking "Send" and then, search for the email again, and click "Archive". If I forget to do right away, it's more work later.

I propose to have two buttons when replying to a mail
* "Send"
* "Send and Archive"

What do you think?

marclaporte:
I realize this may not be super clear at first glance. So I will describe in more detail and with a screenshot for the visual people :-)

User story:
1- read a received email
2- click "reply" or "reply to all" or "forward"
3- compose an answer (reply) or add additional information (forward)
4- determine that this reply or forward is the end of it (If no one replies, I no longer want to see this in my inbox anymore). Should be archived so can still be found via search.

Thus, in addition to "Save draft" and "Send" there should be an action: "Send and Archive". 

On desktop, there should be a mouse-over that explains. "Send this email and archive the original message (to which it was a reply to or forward of)"
![2018-07-13-send-and-archive](https://user-images.githubusercontent.com/1004261/42715123-7a1276aa-86c3-11e8-9d05-fb385f317064.png)

PR #10192, opened 2026-05-25
add discard draft method

### LINKED ISSUE #10184: Delete/Cancel the draft you're currently writing/editing
### Prerequisites

### Proposal

A button to Cancel/Delete the current mail draft your're writing. This could either be next to the Send button or among the top buttons Save, Signature, etc.

### Motivation and context

When i'm writing a draft mail and i change my mind, either because i want to call that person instead or because i want to write that mail later, or manage other mails before etc... there is no easy way to delete the currently edited mail draft.
To achieve that i need to go back to my inbox, enter the Drafts folder, select the draft i was editing, and then delete it. A single button on the editing view that sends you back to the inbox once the draft is deleted/canceled would be much quicker and efficient.
Also sometimes i forget to delete that previous draft i abndonned and my draft folder quickly baloons into a pile of unfinished responses.

johndoh:
closes #10184

Add a method to discard a saved draft, cleaning up the compose data in the session.

I put the discard button at the other end of the toolbar to prevent accident click problem if its too close to save.

Screenshot 2026-05-24 081902


alecpl:
I'm not sure I like this button there.

Maybe it should get integrated with a Cancel button (#6251), e.g. as a question after clicking Cancel.

With requests like #5937, maybe there's a need for some "Actions" button near by the Send button. Would make sense if there's at least two extra actions though. "Send and Archive" is a plugin feature, so it might leave the Discard alone.

johndoh:
Should I remove the skin change so the function can be added to the core, then people can make UI suggestions?

I wondered if may be it could go at the bottom next to the open in new window link on big screens and in the menu on small screens:

Screenshot 2026-05-25 143023

Screenshot 2026-05-25 143115

PR #10208, opened 2026-05-31
[Security] Restrict allowed_classes in session (Redis/Memcache) and cache unserialize() to prevent PHP Object Injection

XananasX7:
## Summary

Three `unserialize()` call sites in Roundcube read data from external backends (Redis, Memcache, database cache) without restricting the `allowed_classes` option. An attacker who can write to these backends can inject a crafted PHP serialized payload containing a **gadget chain** and achieve **Remote Code Execution**.

## Affected locations

| File | Line | Backend | Risk |
|------|------|---------|------|
| `session/redis.php` | ~130 | Redis | **Critical** — unauthenticated Redis is common |
| `session/memcache.php` | ~116 | Memcache | **High** — Memcache has no auth by default |
| `rcube_cache.php` | ~458 | DB / any | **Medium** — cache index deserialization |

## Attack scenario

1. Attacker can write to the Redis or Memcache backend (unauthenticated instance, SSRF, or lateral movement on shared hosting).
2. Attacker injects a serialized payload using gadget classes loaded by Roundcube/Symfony (e.g., any library with a dangerous `__destruct`/`__wakeup`).
3. On the next request, the session is read and deserialized → gadget chain fires → **RCE**.

## Fixes

- **session/redis.php** / **session/memcache.php**: Add `'allowed_classes' => false` — the session envelope only stores scalar fields (`expires_at`, `ip`, `vars`); no PHP objects should ever be present.
- **rcube_cache.php**: Add `'allowed_classes' => false` for the cache index read (plain array of string keys). Add an `$allowed_classes` property (default `true`) to `rcube_cache` so subclasses can restrict their data reads.

## Files changed

- `program/lib/Roundcube/session/redis.php`
- `program/lib/Roundcube/session/memcache.php`
- `program/lib/Roundcube/rcube_cache.php`

### DIFF
--- a/program/lib/Roundcube/rcube_cache.php
+++ b/program/lib/Roundcube/rcube_cache.php
@@ -29,6 +29,16 @@ class rcube_cache
     protected $prefix;
     protected $ttl;
     protected $packed;
+
+    /**
+     * Controls the allowed_classes option passed to unserialize().
+     * Set to false to disallow all object deserialization (safest for caches that
+     * only store scalar/array values), or an array of class names to restrict which
+     * PHP objects may be instantiated. Defaults to true for backwards compatibility.
+     *
+     * @var bool|array
+     */
+    protected $allowed_classes = true;
     protected $indexed;
     protected $index;
     protected $index_update;
@@ -455,7 +465,8 @@ class rcube_cache
         }
 
         $data = $this->get_item($this->ikey());
-        $this->index = $data ? unserialize($data) : [];
+        // The index only stores an array of string cache keys — never PHP objects.
+        $this->index = $data ? unserialize($data, ['allowed_classes' => false]) : [];
     }
 
     /**
@@ -609,7 +620,15 @@ class rcube_cache
      */
     protected function unserialize($data)
     {
-        return $this->packed ? @unserialize($data) : $data;
+        if (!$this->packed) {
+            return $data;
+        }
+
+        if ($this->allowed_classes !== true) {
+            return @unserialize($data, ['allowed_classes' => $this->allowed_classes]);
+        }
+
+        return @unserialize($data);
     }
 
     /**
--- a/program/lib/Roundcube/rcube_cache.php
+++ b/program/lib/Roundcube/rcube_cache.php
@@ -33,10 +33,13 @@ class rcube_cache
     /**
      * Controls the allowed_classes option passed to unserialize().
      * Set to false to disallow all object deserialization (safest for caches that
-     * only store scalar/array values), or an array of class names to restrict which
-     * PHP objects may be instantiated. Defaults to true for backwards compatibility.
+     * only store scalar/array values), or an array of fully-qualified class names
+     * to restrict which PHP objects may be instantiated on deserialization.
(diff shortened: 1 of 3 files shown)


alecpl:
In recent releases we eliminated "dangerous __destruct/__wakeup" classes. However, we still should use allowed_classes. Not with 'false' value, there's need to allow some classes like `DateTime` or `rcube_message_header`. We have to figure out which ones.

And this will be a breaking change for plugins, so we'll not do this in a minor release.

XananasX7:
Thank you @alecpl — very useful guidance! I've just pushed an update that addresses your points:

**What changed:**
- Added a `set_allowed_classes($classes)` public method to `rcube_cache` so callers can supply an explicit allowlist of classes rather than using a blanket `false`
- The cache **index** (`load_index()`) keeps `allowed_classes => false` — the index stores only an array of string cache keys, never PHP objects
- The **session** backends (Redis, Memcache) also keep `allowed_classes => false` — the session envelope is `['expires_at' => int, 'ip' => string, 'vars' => string]`, no objects

**On the broader question of which classes to allow:**

I audited all core `cache->set()` call sites in Roundcube itself:
- `rcube_imap` IMAP cache: stores plain arrays (message counts, folder lists, capability flags) — no objects
- `rcube_ldap`: stores plain arrays/strings (DN lookups, group lists)  
- `rcmail_oauth`: stores `json_decode(..., true)` results — plain arrays
- `database_attachments` / `redundant_attachments` plugins: store base64 strings

The core codebase itself doesn't need `DateTime` or `rcube_message_header` in the general cache — those are in-memory only or go through `rcube_imap_cache` (which uses `rcube_db::decode()`, a separate path).

**The plugin compatibility concern:**

You're right that restricting `allowed_classes` globally would be a breaking change for third-party plugins that legitimately cache PHP objects. The `set_allowed_classes()` method lets the project set defaults per cache instance in a future release, and lets plugins opt-in to either direction. A hardened default (e.g., allowed_classes => false) could ship in a major release with a migration guide.

Happy to add per-call-site annotations if you'd prefer to enumerate the full list of safe classes in this PR rather than in a follow-up.

XananasX7:
Thanks for the detailed response @alecpl — that makes total sense. Restricting to `false` would break plugin-defined serialized classes, and it is the right call to defer this to a major release where a proper allow-list can be built.

I am happy to help with that groundwork if it would be useful. Specifically I can:

1. **Audit the codebase** to identify all classes that are serialized into the Redis/Memcache/session stores and produce a candidate `allowed_classes` list
2. **Draft the migration path** — e.g. a config option `$config['unserialize_allowed_classes']` that defaults to `true` (current behavior) and can be tightened per-deployment
3. **Write a compatibility shim** so plugins can declare their serializable classes via a hook

Just let me know if any of that would help. Otherwise I will close this PR and leave it as a reference for whenever the hardening work is scheduled.

XananasX7:
Understood @alecpl — makes sense to defer to a major release where a proper allowlist can be built with full knowledge of what plugins serialize. Closing this PR for now and leaving it as a reference. Happy to help with the audit work whenever that's scheduled.

PR #10211, opened 2026-05-31
elastic: restore color to unread icon

### LINKED ISSUE #9862: Bold indicates unread... Bad accessibility
Currently, in the mailbox view, the difference between a read message and an unread message is a bold text title for the latter.

It would be great if there was a preference to change it to/and/or, for instance background color. 

That would be an accessibility improvement. 

Also that way we wouldn't have to think about what particular skin we are using, I suppose.

Then of course there are some people unable to distinguish between background colors. Therefore another option would be some kind of a marker to distinguish read and unread.

Then even a screen reader could read it, for people with bad eyes.

johndoh:
revert 1 line from 9427ec1d35445170cd5d25b3aae9b69950610c3b, restore the colour to the unread icon in the message list of the elastic skin.

I think its not clear why this commit removed the colour and feedback on the change is negative. resolves #9862

### DIFF
--- a/skins/elastic/styles/widgets/lists.less
+++ b/skins/elastic/styles/widgets/lists.less
@@ -733,6 +733,7 @@ table.fixedcopy {
 
         &.unread:before {
             content: @fa-var-circle;
+            color: @color-list-unread-status;
             font-size: .5rem;
         }
 


dpedu:
This has been sitting for awhile. What can we do to get this merged?

PR #10222, opened 2026-06-05
feat: Created a plugin to display DKIM status of an email

HQJaTu:
A well known fact is that no reputable email service provider will handler incoming SMTP without proper signatures. This plugin will display the incoming email DKIM status as analysed locally.

### DIFF
--- a/plugins/message_security_info/README.md
+++ b/plugins/message_security_info/README.md
@@ -1,0 +1,119 @@
+Message security info plugin for Roundcube
+==========================================
+
+Surfaces a received message's sender-authentication status (SPF / DKIM / DMARC)
+and lets users inspect the relevant headers.
+
+Adds a **Message Security** link to the message header's links row (next to
+*Summary / Headers / Plain text*). The link is always present; its icon colour
+reflects the overall sender-authentication verdict. **DMARC** drives the icon
+when the receiving server evaluated it (DMARC already implies an aligned SPF or
+DKIM pass); otherwise the present **SPF** and **DKIM** results are combined,
+omitting whichever is missing:
+
+| Situation | Icon |
+|---|---|
+| DMARC pass — or all present SPF/DKIM checks pass (DKIM aligned with From) | green check |
+| A present check is soft/incomplete (SPF softfail/neutral, DKIM valid but not aligned, or nothing to check) | amber triangle |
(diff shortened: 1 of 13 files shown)


alecpl:
dkim_info, dkim_status, dkim_banner would be better than dkim_display.

HQJaTu:
> dkim_info, dkim_status, dkim_banner would be better than dkim_display.

You're right. The name I chose is ambiguous at best. Out of those three suggestions, I'll got with _dkim_info_.

alecpl:
Without looking at the code yet, one question. Wouldn't it make more sense to display only warnings? I.e. do not display successful DKIM validation status at all. That would also (maybe partially) prevent redressing attacks like in #6450.

ps. the static analysis errors should be fixed in the current master.

HQJaTu:
> Without looking at the code yet, one question. Wouldn't it make more sense to display only warnings? I.e. do not display successful DKIM validation status at all. That would also (maybe partially) prevent redressing attacks like in #6450.

For UX, I'd like to see indicator that DKIM is "on green". For such sunny day -cases having a tooltip for information should do it to reduce footprint. For any non-sunny -case I'd like to see as much information on the case as needed.

> ps. the static analysis errors should be fixed in the current master.

Yes. As this is my first contribution, I'm not sure if fixing files that I didn't break is my responsibility. Is it?

alecpl:
> > ps. the static analysis errors should be fixed in the current master.
> 
> Yes. As this is my first contribution, I'm not sure if fixing files that I didn't break is my responsibility. Is it?

It is not. It was a hint that you can make the CI green if you merge from master.

HQJaTu:
> It is not. It was a hint that you can make the CI green if you merge from master.

Ok.
While doing the rename, part of my workflow is to merge from upstream. CI is on green now.

Moving to improve UX.

HQJaTu:
> Without looking at the code yet, one question. Wouldn't it make more sense to display only warnings? I.e. do not display successful DKIM validation status at all. That would also (maybe partially) prevent redressing attacks like in #6450.

In this [forum post](https://www.roundcubeforum.net/index.php?topic=31991.msg80543#msg80543), I describe the changes made.

HQJaTu:
In this [forum post](https://www.roundcubeforum.net/index.php?topic=31991.msg80547#msg80547), I describe the new design of a green DKIM validation indicator.

PR #10225, opened 2026-06-07
feat: Login form works in two steps: 1) input username, 2) password

HQJaTu:
- CSRF will be updated and user will never be bothered on expired token
- Authentication flow is in two parts
  - Subsequent changes will be made to support Passkeys (out of scope for this)

### DIFF
--- a/plugins/twostep_login/README.md
+++ b/plugins/twostep_login/README.md
@@ -1,0 +1,73 @@
+Two-step login plugin for Roundcube
+===================================
+
+Splits the Roundcube login form into a two-step flow, similar to modern
+identity providers:
+
+1. **Step 1** — only the username field (`#rcmloginuser`), plus the server
+   selector if one is configured, is shown together with a **Next** button.
+2. **Step 2** — after the user enters a username and clicks *Next* (or presses
+   Enter), the password field (`#rcmloginpwd`) and the **Login** button are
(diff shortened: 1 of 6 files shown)


alecpl:
I don't like the idea in general, even though big mail providers might be switching to such logon processes.

If we include the plugin we'll have to support it.

HQJaTu:
> I don't like the idea in general, even though big mail providers might be switching to such logon processes.

Couple things:
- This is simply my intermediate step towards Passkeys-login -plugin, which I'm finalizing
- Unfortunately for all of us, its not a matter of liking or disliking. This is how Internet works in 2020s, you first enter the user and then depending on the chosen authentiation method, next thing happens. In this instance: a totally normal password prompt is presented to user.
- Note: In default Roundcubemail there is a huge UX-dip when a session times out and user is returned back to login-screen. CSRF-token expires and no credential, correct or incorrect, is valid. If user remembers to refresh the page. This plugin vastly improves UX, login lets you pass if correct credentials are entered.

> If we include the plugin we'll have to support it.

If you look the code, the plugin is very simple and not likely to require any form of support in future.

Passkeys-plugin: a completely different story. This PR is just to bring Roundcubemail authentication to today's norm, not offer any real new functionality.

HQJaTu:
I don't think this will be approved. Closing.

PR #10229, opened 2026-06-10
add new list_hover_menu option

### LINKED ISSUE #10223: Add an option to disable quick actions mouse-over menu on the messages list
### Prerequisites

### Proposal

I don't like the mouse-over menu that appeared in 1.7.0 because it comes on top of the date/time of the message:
Image
The date/time is a much more important information to me.

So I fully support the idea, from https://github.com/roundcube/roundcubemail/issues/7141#issuecomment-1604746410 : 
"If people don't like it we can consider removing it or making optional (or a plugin)." :)

Thanks a lot for the great work !

### Motivation and context

As an option it keeps the new feature : people may like it, might even be the default.
And if like me you don't like things poping up : you can disable it ;)

johndoh:
Closes #10223

This adds a new config option `list_hover_menu` and includes it has part of the list options. the change is backwards compatible.

Screenshot 2026-06-10 143923

I'll mark this as draft for now because it turned out to be a bigger change than I was expecting but I can't think of another way right now.

I'm also not sure if `Enable quick menu` is the right label.

### DIFF
--- a/config/defaults.inc.php
+++ b/config/defaults.inc.php
@@ -938,6 +938,9 @@ $config['message_sort_order'] = 'DESC';
 // subject, from, to, fromto, cc, replyto, date, size, status, flag, attachment, priority
 $config['list_cols'] = ['subject', 'status', 'fromto', 'date', 'size', 'flag', 'attachment'];
 
+// show the hover menu in the message list
+$config['list_hover_menu'] = true;
+
 // the default locale setting (leave empty for auto-detection)
 // RFC1766 formatted language name like en_US, de_DE, de_CH, fr_FR, pt_BR
 $config['language'] = null;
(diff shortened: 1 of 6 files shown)


alecpl:
"hover menu", "mouse-over menu", "list row menu"? I'm not sure either, but I'd remove word "enable".

johndoh:
Thanks for the suggestions. I've put it as "Mouse-over menu" because I'm not sure users would know what hover means.

Do you have any other comments before I mark this as ready?

hejnm1am:
I see that you implemented this as a list option (so you can change it per mail folder).
I believe it would be better as user setting (have it the same way for all folders).

johndoh:
It is a global option same as `Sorting column` and `Sorting order`. May be it should be above `List mode` to group those 3 options together while `List mode` applies per folder. I'm not sure.

alecpl:
> It is a global option same as `Sorting column` and `Sorting order`. May be it should be above `List mode` to group those 3 options together while `List mode` applies per folder. I'm not sure.

I think it does not matter at this point. You can add a divider to separate global from per-folder options, but it could be done later.

PR #10231, opened 2026-06-11
Free Imagick resources after failing

### LINKED ISSUE #10230: Image processing through Imagick leave large temporary files when it fails
### Prerequisites

### Describe the issue

I've notice that my temporary storage gets full from time to time. It happens when RC try to scale some images using Imagick and fails, showing the PHP Error.

After that, a file called "magick-xxxxxxxxxxxxxxxxxxx" is left in my temporary storage. Sometimes, there are several very large files like this and the filesystem gets full.

This only happens with specific images and it's not related to the their properties (size, type, ...) because I've tried with several images, with same type and size and I don't always get the error.

Chrome

8.4.21

1.7.1

### JavaScript errors

_No response_

### PHP errors

roundcube[2009831]: <e2eed00e> PHP Error: Unable to scale image in /aplica/round/www/program/lib/Roundcube/rcube_image.php on line 189

dacolboy:
Fixes #10230 

Forcing the release of Imagick resources, we can prevent the problem with abandoned large files in temporary storage.

### DIFF
--- a/program/lib/Roundcube/rcube_image.php
+++ b/program/lib/Roundcube/rcube_image.php
@@ -359,6 +359,14 @@ class rcube_image
                     return true;
                 }
             } catch (\Exception $e) {
+                // Free resources
+                if ($image instanceof \Imagick) {
+                    try {
+                        $image->clear();
+                    } catch (\Exception $e) {
+                        // Ignore errors
+                    }
+                }
                 rcube::raise_error($e, true, false);
             }
         }
--- a/program/lib/Roundcube/rcube_image.php
+++ b/program/lib/Roundcube/rcube_image.php
@@ -197,8 +197,11 @@ class rcube_image
                         } finally {
                             // Free resources
                             if ($image instanceof \Imagick) {
-                                $image->clear();
-                                $image->destroy();
+                                try {
+                                    $image->clear();
+                                } catch (\Exception $e) {
+                                    // Ignore errors
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


program/lib/Roundcube/rcube_image.php
```diff
@@ -171,6 +171,7 @@ public function resize($size, $filename = null, $browser_compat = false)
                     // use PHP's Imagick class
                     else {
                         try {
+                            $image = null;
```
alecpl: This makes no sense here. Shouldn't this be done after `$image->clear()`?

alecpl (changes requested):

program/lib/Roundcube/rcube_image.php
dacolboy: I put it there just for aesthetics. It could even be removed.

dacolboy: I've had to move the code in finally block to catch block, before calling rcube::raise_error. Otherwise, it wasn't executed.

program/lib/Roundcube/rcube_image.php
```diff
@@ -192,6 +192,14 @@ public function resize($size, $filename = null, $browser_compat = false)
                                 $result = '';
                             }
                         } catch (\Exception $e) {
+                            // Free resources
+                            if ($image instanceof \Imagick) {
```
alecpl: In theory $image may not exists here. Just `!empty($image))` here should work fine.

alecpl (changes requested):

PR #10239, opened 2026-06-20
Elastic: improve extensability of hover menu

johndoh:
I know I could do this with a plugin and the `skin-message-list-hover-menu-init` and `skin-message-list-hover-menu` hooks but the only visible effect of this change is to swap the delete and flag icons around (so that hover menu, or no hover menu, the flag stays in one place).

Recording 2026-06-20 143133

with these changes then using only custom styling in `_styles.less` its possible to customise the menu quite a lot more than currently.
1. adding `flagged`/`deleted` class to the menu - allows for the text colour to change like it does on the row itself to reflect the status of the message
2. having both the `date` and `size` text spans populated allows both bits of information to be displayed if required

in the example above the date is no longer obscured during hover and both the date and size are visible

### DIFF
--- a/skins/elastic/ui.js
+++ b/skins/elastic/ui.js
@@ -3165,13 +3165,14 @@ function rcube_elastic_ui() {
     function message_list_hover_menu_init(list_element) {
         var record,
             menu = $('<div class="menu listing-hover-menu">'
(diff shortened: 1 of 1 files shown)


johndoh:
The failing coding style check is not related to this PR

PR #10255, opened 2026-07-09
Fix deleting all folder messages on 'select all' when the search context is lost

TowyTowy:
Closes #9671.

### Problem

When a search result is deleted using "select all", the client sends `_uid=*` together with `_search=<id>`. The delete handler resolves `*` against the storage search set. If the search context is no longer in the session — e.g. it was overwritten by a new search in another browser tab — `rcmail_action_mail_index::run()` does not call `set_search_set()`, so the storage object has no search set. `rcube_storage::parse_uids('*')` then falls back to `1:*` and **every message in the folder is deleted** instead of only the (now lost) search result.

This matches the analysis in #9671:

> I can see how one tab removed a "search state" from the session, and the other tab, when not seeing the data in session, falls back to deleting all messages in a folder. I think the solution will be to bail out if `_search` is set in the request, but the "search state" does not exist in session.

### Fix

Bail out with an error in the delete handler when a `_uid=*` request is bound to a search (`_search` present) but no search set exists on the storage object. Single-message and explicit-UID deletes are unaffected, and a normal "select all" in a non-search folder view (no `_search`) still works.

### Tests

Added a regression test in `tests/Actions/Mail/DeleteTest.php` that mocks a missing search set and asserts the action aborts without calling `delete_message()`. It fails on current master (the un-guarded code calls `delete_message('*')`, i.e. the whole-folder delete) and passes with this change. CHANGELOG entry included.

To reproduce manually: open a search result with more than one message; in a second tab run a different search (overwriting the session search state); back in the first tab, Select all → Delete. Without this change the whole folder is emptied.

---

*Disclosure: this fix was prepared with the assistance of an AI coding agent (Claude) and reviewed before submission.*

### DIFF
--- a/program/actions/mail/index.php
+++ b/program/actions/mail/index.php
@@ -65,6 +65,25 @@ class rcmail_action_mail_index extends rcmail_action
             $rcmail->output->set_env('search_request', $_REQUEST['_search']);
             $rcmail->output->set_env('search_text', $_SESSION['last_text_search']);
         }
+        // Bail out if the client requested a search context that does not exist
+        // in the session anymore, e.g. it has been overwritten by a search in
+        // another browser tab. Otherwise the request would work on the whole folder
+        // instead of the search result, e.g. a "select all" delete request would
+        // delete all messages in the folder (#9671).
+        // Note: In addressbook actions _search is not a message search reference,
+        // but a search string or a reference to $_SESSION['contact_search'].
+        elseif (
+            !empty($_REQUEST['_search'])
+            && !in_array($rcmail->action, ['autocomplete', 'list-contacts', 'search-contacts'])
+        ) {
+            $rcmail->output->show_message('searchexpired', 'error');
+
+            if ($rcmail->output->ajax_call || $rcmail->output->get_env('framed')) {
+                $rcmail->output->send('iframe');
+            }
+
+            $rcmail->output->redirect(['_mbox' => $rcmail->storage->get_folder()]);
+        }
 
         // remove mbox part from _uid
         $uid = rcube_utils::get_input_string('_uid', rcube_utils::INPUT_GPC);
--- a/program/localization/en_US/messages.inc
+++ b/program/localization/en_US/messages.inc
@@ -113,6 +113,7 @@ $messages['notuploadedwarning'] = 'Not all attachments have been uploaded yet. P
 $messages['searchsuccessful'] = '$nr messages found.';
 $messages['contactsearchsuccessful'] = '$nr contacts found.';
 $messages['searchnomatch'] = 'Search returned no matches.';
+$messages['searchexpired'] = 'The search result is not available anymore. Please repeat the search.';
 $messages['searching'] = 'Searching...';
 $messages['checking'] = 'Checking...';
 $messages['stillsearching'] = 'Still searching...';
--- a/tests/Actions/Mail/IndexTest.php
+++ b/tests/Actions/Mail/IndexTest.php
@@ -4,6 +4,8 @@ namespace Roundcube\Tests\Actions\Mail;
 
 use PHPUnit\Framework\Attributes\Group;
 use Roundcube\Tests\ActionTestCase;
+use Roundcube\Tests\OutputHtmlMock;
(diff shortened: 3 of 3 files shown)


program/actions/mail/delete.php
```diff
@@ -48,6 +48,18 @@ public function run($args = [])
...
         }
 
+        $search_request = rcube_utils::get_input_string('_search', rcube_utils::INPUT_GPC);
+
+        // Bail out on a "select all" (_uid=*) request that is bound to a search,
+        // but the search context does not exist in the session anymore (e.g. it
+        // has been overwritten by a search in another browser tab). Otherwise
+        // the '*' would resolve to the whole folder and delete all its messages
+        // instead of only the search result (#9671).
+        if ($search_request && $_POST['_uid'] === '*' && !$rcmail->storage->get_search_set()) {
```
alecpl: The second condition can be removed. It does not matter what `_uid` is.

alecpl (changes requested):
We have the same issue in delete, move and mark actions too (maybe more). So, I think we should move the check to index.php and bail out whenever search is requested but does not exist, and `_action` is not empty (or if it is an ajax action, I'm not sure which would be better).

TowyTowy:
Moved the check into `rcmail_action_mail_index::run()` so it now covers delete, move, mark and everything else dispatched through the mail task, and dropped the `_uid` condition plus the delete.php check entirely (it went there rather than top-level index.php because `_search` means something different in the contacts task). It bails when `_search` doesn't match the session context and `_action` is non-empty, with exemptions where `_search` isn't a message-search token (autocomplete, contact search), check-recent/refresh which already handle a lost search gracefully, and compose. If you'd prefer limiting it to ajax requests instead, that's a one-line change - happy to adjust.

program/actions/mail/index.php
```diff
@@ -65,6 +65,28 @@ public function run($args = [])
...
+        // instead of the search result, e.g. a "select all" delete request would
+        // delete all messages in the folder (#9671).
+        // Skip actions where _search is not a message search reference (autocomplete,
+        // list-contacts, search-contacts), periodic requests that handle a lost
+        // search gracefully (refresh, check-recent), and compose, which uses _search
+        // only to get back to the search result view.
+        elseif (
+            !empty($_REQUEST['_search'])
+            && !empty($rcmail->action)
+            && !in_array($rcmail->action, ['autocomplete', 'list-contacts', 'search-contacts', 'refresh', 'check-recent', 'compose'])
```
alecpl: `_search` in `refresh`, `check-recent` and `compose` is the same search, so they should not be excluded.

program/actions/mail/index.php
```diff
@@ -65,6 +65,28 @@ public function run($args = [])
...
+        // another browser tab. Otherwise the action would work on the whole folder
+        // instead of the search result, e.g. a "select all" delete request would
+        // delete all messages in the folder (#9671).
+        // Skip actions where _search is not a message search reference (autocomplete,
+        // list-contacts, search-contacts), periodic requests that handle a lost
+        // search gracefully (refresh, check-recent), and compose, which uses _search
+        // only to get back to the search result view.
+        elseif (
+            !empty($_REQUEST['_search'])
+            && !empty($rcmail->action)
```
alecpl: I would not require an action.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ This file includes only changes we consider noteworthy for users, admins and plu
 
 ## Unreleased
 
+- Reject mail actions referencing a search context that no longer exists in the session, e.g. deleting a search result with "select all" could delete all messages in the folder (#9671)
```
alecpl: Please, remove the changelog entry.

alecpl (changes requested):

TowyTowy:
Done — dropped the action requirement and removed the changelog entry. Also took refresh, check-recent and compose out of the exclusions per your other comment; only autocomplete, list-contacts and search-contacts are left, where `_search` is an addressbook search string or a `contact_search` key rather than a message search. Rebased on current master.

PR #10259, opened 2026-07-12
Inherit PHP interpreter during `installto.sh`

orhtej2:
Currently executing `phpX.Y ./bin/installto.sh` will invoke follow up steps with env-default  generic `php` command.

Given PHP now ships built in [`PHP_BINARY`](https://www.php.net/manual/en/reserved.constants.php#constant.php-binary) that indicates the currently running interpreter - reuse it.

### DIFF
--- a/bin/update.sh
+++ b/bin/update.sh
@@ -317,13 +317,13 @@ function repo_key($repo)
 function find_composer()
 {
     if (is_file(INSTALL_PATH . 'composer.phar')) {
-        return PHP_BINARY . ' composer.phar';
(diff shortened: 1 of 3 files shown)


bin/update.sh
```diff
@@ -317,13 +317,13 @@ function repo_key($repo)
 function find_composer()
 {
     if (is_file(INSTALL_PATH . 'composer.phar')) {
-        return 'php composer.phar';
+        return PHP_BINARY . 'composer.phar';
```
alecpl: Missing space.

PR #10260, opened 2026-07-13
Zipdownload: various improvements to this plugin

gurnec:
(Extracted from #10175.)

This includes several improvements to zipdownload. Feel free to cherry-pick what looks useful, or ask me to split this into multiple PRs, or whatever makes most sense.

 * Currently the entire zip is written to the output buffer before sending which can easily cause an OOM. Instead, flush every 512 KiB.
 * Update the `.htaccess`, see 9bd686c for details.
 * For maildir exports, change the modified-time of each `.eml` file to the IMAP internal time to allow sorting extracted emails.
 * Change the default charset in the config to UTF-8, see a6b857f.

### DIFF
--- a/plugins/zipdownload/zipdownload.php
+++ b/plugins/zipdownload/zipdownload.php
@@ -19,6 +19,7 @@ class zipdownload extends rcube_plugin
     private $charset = 'ASCII';
     private $names = [];
     private $default_limit = '50MB';
+    private $timezone;
 
     // RFC4155: mbox date format
     public const MBOX_DATE_FORMAT = 'D M d H:i:s Y';
@@ -251,14 +252,15 @@ class zipdownload extends rcube_plugin
             foreach ($uids as $uid) {
                 $headers = $imap->get_message_headers($uid);
 
+                // Received (internal) date
+                $date = rcube_utils::anytodatetime($headers->internaldate);
+
                 if ($mode == 'mbox') {
                     // Sender address
                     $from = rcube_mime::decode_address_list($headers->from, null, true, $headers->charset, true);
                     $from = array_shift($from);
                     $from = preg_replace('/\s/', '-', $from);
 
-                    // Received (internal) date
-                    $date = rcube_utils::anytodatetime($headers->internaldate);
                     if ($date) {
                         $date = $date->setTimezone($timezone)->format(self::MBOX_DATE_FORMAT);
                     } else {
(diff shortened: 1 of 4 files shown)


plugins/zipdownload/zipdownload.php
```diff
@@ -369,7 +412,20 @@ private function _deliver_zipfile($tmpfname, $filename)
 
         $rcmail->output->download_headers($filename, ['length' => filesize($tmpfname)]);
 
-        readfile($tmpfname);
```
alecpl, line 372: The problem is not readfile(), but output buffering. Can't we just disable output buffering before calling readfile()?

plugins/zipdownload/zipdownload.php
```diff
@@ -305,6 +307,7 @@ private function _download_messages($messageset)
         }
 
         // open zip file
+        putenv('TZ=UTC');  // see _datetime_to_ziplocal() comments
```
alecpl, line 310: Will this still be needed if we use `setMtimeName()` on every file?

plugins/zipdownload/zipdownload.php
gurnec: I have a *very slight* preference to use `fread()` over `readfile()` with large files because the latter uses `mmap()` which means:
 * pages mapped by the kernel don't count against PHP's `memory_limit`
 * memory usage is difficult to predict, and can change between kernel versions
 * PHP doesn't use `madvise(..., MADV_SEQUENTIAL)` which would make the situation better

In practice, `mmap()` is not some new unoptimized kernel feature, and I doubt its use would cause problems, so I'm happy to move back to `readfile()` if you'd prefer, just let me know.

plugins/zipdownload/zipdownload.php
gurnec: Unfortunately yes. `libzip` expects times in UTC and always [converts them to a local time](https://libzip.org/documentation/zip_file_set_mtime/#CAVEATS) unless `zip_file_set_dostime()` is used which `ZipArchive` doesn't expose. `libzip` [uses `localtime()`](https://github.com/nih-at/libzip/blob/9c1dc196cf5867485920bc6be12fc370f1b763c6/lib/compat.h#L186-L199) to do the conversion which can be controlled by the [`TZ` environment variable on POSIX](https://en.cppreference.com/c/chrono/localtime#Notes) and [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s#remarks). Different `libc`s have different ways of setting a default, for example `glibc` uses `/etc/localtime` (`man tzset`). Overriding the default to UTC was the best way I could find to convert these `localtime()` calls to a no-op.

(Note that I've only tested this on `glibc` and Windows.)

PR #10263, opened 2026-07-14
Archive: fix TypeError when archiving all messages via select-all (#10107)

MiMoHo:
## Problem

Using "select all" and then archiving produces a fatal server error ("Serverfault - moving all Mails in the archive"). On PHP 8 the request dies with:

`TypeError: count(): Argument #1 ($value) must be of type Countable|array, string given`

## Root cause

With select-all, `rcmail_action::get_uids()` yields the string `'*'` as `$uids` (the IMAP wildcard). The archive plugin then calls `count()` on it in two places:

- `plugins/archive/archive.php:346` — `move_messages_worker()` does `return count($uids);`
- `plugins/archive/archive.php:160` — `move_messages()` does `$count = count($uids);` for the archive-folder-source branch

`count()` on a string throws a `TypeError` on PHP 8, aborting the whole archive action.

## Fix

Mirror how core handles `'*'` in `program/actions/mail/move.php` (`$count += is_array($uids) ? count($uids) : 1;`): treat a non-array `$uids` (the `'*'` wildcard) as a single unit instead of calling `count()` on it. `set_flag()` and `move_message()` already accept `'*'`, so the wildcard is passed through unchanged.

Both `count($uids)` calls are replaced with `is_array($uids) ? count($uids) : 1`.

## Testing

Added `test_move_messages_worker_select_all()` in `plugins/archive/tests/ArchiveTest.php`. It mocks storage so `set_flag()`/`move_message()` return true and invokes `move_messages_worker('*', ...)` via reflection. Before the fix the test fails with the `count()` TypeError at archive.php:346; after the fix it passes and asserts the correct return value and recorded source/destination folders. Full ArchiveTest.php passes (4 tests) and phpstan (level 4) reports no errors on the changed files.

Fixes #10107

### DIFF
--- a/plugins/archive/archive.php
+++ b/plugins/archive/archive.php
@@ -157,7 +157,10 @@ class archive extends rcube_plugin
         foreach (rcmail_action::get_uids(null, null, $multifolder, rcube_utils::INPUT_POST) as $mbox => $uids) {
             $all = $uids === '*';
             if (!$this->archive_folder || $mbox === $this->archive_folder || str_starts_with($mbox, $archive_prefix)) {
-                $count = count($uids);
+                // $uids is the '*' string for "select all"; count the folder's
+                // messages instead of count()'ing a string (#10107). The value
+                // is only used for the row-refill below, which $all skips.
+                $count = $all ? $storage->index($mbox)->count() : count($uids);
                 continue;
             } elseif (!$archive_type || $archive_type == 'folder') {
                 $folder = $this->archive_folder;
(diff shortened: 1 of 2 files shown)


alecpl:
The fix is not that simple. `$uids` can be a string (`*` or e.g. `1,2,3`). We need to count the messages as we do in another code path. Or we change the code so the count is not needed (probably complicated and involves client-side). Using `$count = 1` is wrong.

MiMoHo:
Thanks, you're right — `count()` on the `*` string was the wrong approach. Reworked it to resolve `'*'` to the actual UID list via `$storage->index()`, the same way the `archive_type` branch does a few lines below, so `move_messages_worker()` always receives a real array and returns the true number of moved messages. The already-in-archive skip branch uses `$storage->index($mbox)->count()` for select-all. Updated the test to assert the real count; phpunit + phpstan clean.

PR #9637, opened 2024-09-14
vcard: Fix whitespace handling in line cont's

respiranto:
Previously, multiple whitespace characters at the start of a continuation line would all be dropped, instead of only the first one.

Also,
 - restrict line continuation characters to SPACE and TAB.

Note that, like before, this identifies the CR (`\r`) character with the empty string, and thereby notably does not require a CRLF (`\r\n`) sequence (which is mandated by RFCs 2426, 2425) for line termination (i.e., `\n` suffices).

Fixes: Bug 1 of issue #9593.

PR #10265, opened 2026-07-14
Fix vCard import mis-detecting folded continuation lines as BEGIN/END:VCARD (#9593)

### LINKED ISSUE #9593: Contacts import: Whitespace & vCard object separators
### Prerequisites

- [X] I can recreate the issue with all plugins disabled

### Describe the issue

This bug report describes several related bugs, related to

* a) whitespace (space, tab) in imported vCard files, and
* b) vCard object separators (`BEGIN:VCARD`, `END:VCARD`).

## Bug 1: Leading whitespace in line continuation silently dropped

Example (note that vCard mandates CRLF as newline sequence):
```
BEGIN:VCARD
VERSION:3.0
N:Doe;Jane;;;
FN:Jane Doe
NOTE:an
  example
END:VCARD
```

The `NOTE` value is parsed as `anexample` instead of `an example` (only the first whitespace character should be dropped - see [RFC 2426](https://datatracker.ietf.org/doc/html/rfc2426#section-2.6)).

In particular, this means that a Roundcube export followed by a Roundcube import may silently fail to recreate the original data.

## Note 2: Leading and trailing whitespace in a logical line is silently dropped

Any number of space/tab characters at the start or end of a logical line (or a component value, such as in `N`) is dropped.

This is not a bug, IMHO, given that Roundcube also strips surrounding whitespace when entering data via its web UI.

I think it might be related, however.

## Note 3: Repeated `BEGIN:VCARD`: All ignored until last

If there are multiple `BEGIN:VCARD` lines before any `END:VCARD` line, all lines (not only `BEGIN:VCARD` lines) until the last of those `BEGIN:VCARD` lines are ignored.

I would say this is also not a bug, because such input is invalid.  A warning or error message would certainly be nice, though.

## Note 4: Repeated `END:VCARD` cause duplication

If a VCard object is terminated by more than one `END:VCARD` line, the entry is imported as often as there are `END:VCARD` lines.

If on import, one does **not** choose to "[r]eplace the entire address book", only one instance is imported, but with the note "Skipped (n-1) existing entries: [...]".

Any physical lines after the first `END:VCARD` that are neither `BEGIN:VCARD` nor `END:VCARD` are apparently ignored.

Again, not necessarily a bug, because any such input is of invalid syntax (but a warning or error message would be nice).

## Bug 5: vCard object separators wrongly recognized in line continuations

If a physical line is of the form `[ \t]+(BEGIN|END):VCARD`, it is used as line continuation, but also recognized as vcard object start/end marker.

Example:
```
NOTE:example
 END:VCARD
```

This is treated the same as:
```
NOTE:exampleEND:VCARD
END:VCARD
```

Note: For `BEGIN:VCARD`, the use as line continuation can only be assumed, given that preceding lines are ignored (see Note 3).

While this bug may seem unlikely in practice, I actually witnessed it with real data, likely due to past import/export errors.

## Bug 6: vCard object separators not parsed as logical lines

* (6.1) Any logical line `BEGIN:VCARD` or `END:VCARD` that is broken into multiple physical lines using `\r\n[ \t]\r\n` is not recognized as such.
* (6.2) On the other hand, if a physical line `BEGIN:VCARD` or `END:VCARD` is followed by a line continuation (i.e., a line starting with `[ \t]`), this is (incorrectly) recognized as the corresponding vCard object separator, and the line continuation is silently ignored--unless Bug 5 applies.

Example for (6.1):
```
BEGIN:VCARD
VERSION:3.0
N:Doe;Jane;;;
FN:Jane Doe
EMAIL:jane.doe@example.net
END:
 VCARD
```

The above example fails to import (and is instead attempted to be parsed as CSV--without success).

I acknowledge that this bug hardly occurs in practice.  I found it while investigating the other bugs.

---
Originally reported at the Debian BTS as [Bug 1078775](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078775), where I was asked to report the bug here instead.  The bug report has slightly changed.

Firefox

v8.2

master / 58721e3037671888f82d75fd8e91a8622405af45

### JavaScript errors

_No response_

### PHP errors

_No response_

MiMoHo:
## Problem

When importing a vCard whose property value (e.g. NOTE) is folded onto a continuation line and that continuation line, after trimming, reads `END:VCARD` (or `BEGIN:VCARD`), the importer prematurely closes the current card and splits it into multiple bogus records. Per RFC 2426, folded lines begin with a space or tab, but the importer stripped that leading whitespace before checking for block boundaries.

## Root cause

`rcube_vcard::import()` (program/lib/Roundcube/rcube_vcard.php:565-583) appends the raw line to the current block, then does `$line = trim($line)` and matches the trimmed line against `/^END:VCARD$/i` and `/^BEGIN:VCARD$/i`. A folded continuation line such as ` END:VCARD` becomes `END:VCARD` after trimming and is wrongly treated as a block boundary.

## Fix

Before trimming, record whether the line began with whitespace (`$is_folded`), which per RFC 2426 marks a continuation line. Skip the BEGIN/END boundary checks for folded lines. The trim is still used for the comparison so trailing whitespace around a real boundary is tolerated, and the raw line remains preserved as block content (it is appended before the trim), so folded values are unfolded normally.

## Testing

Added `VCardTest::test_import_continuation_endvcard`, which imports a card whose NOTE folds onto a ` END:VCARD` line followed by a second card. It asserts exactly two cards are detected, the first card's NOTE and trailing EMAIL are preserved, and the second card still parses. The test fails on unmodified code (3 cards detected) and passes after the fix. The full VCardTest suite (15 tests) passes, and phpstan level 4 reports no new errors.

Fixes #9593

### DIFF
--- a/program/lib/Roundcube/rcube_vcard.php
+++ b/program/lib/Roundcube/rcube_vcard.php
@@ -562,6 +562,13 @@ class rcube_vcard
                 $vcard_block .= $line . "\n";
             }
 
+            // A folded/continuation line starts with whitespace (RFC 2426) and
+            // must never be mistaken for a block boundary. It is already kept as
+            // content above, so skip the boundary checks for it.
+            if (isset($line[0]) && ($line[0] === ' ' || $line[0] === "\t")) {
+                continue;
+            }
+
             $line = trim($line);
 
             if (preg_match('/^END:VCARD$/i', $line)) {
--- a/tests/Framework/VCardTest.php
+++ b/tests/Framework/VCardTest.php
@@ -250,4 +250,39 @@ class VCardTest extends TestCase
 
         $this->assertSame($result, "BEGIN:VCARD\r\nVERSION:3.0\r\nFN:\r\nN:;;;;\r\nEND:VCARD");
     }
+
+    /**
+     * A folded continuation line whose content equals END:VCARD (or surrounding
+     * whitespace around the real boundary) must not break block detection (#9593)
+     */
+    public function test_import_continuation_endvcard()
+    {
+        // The NOTE value is folded (RFC 2426) so its continuation line starts
+        // with a space and reads " END:VCARD". It must not end the first card,
+        // and the following card must still be detected.
+        $input = "BEGIN:VCARD\r\n"
+            . "VERSION:3.0\r\n"
+            . "FN:First Card\r\n"
+            . "N:Card;First;;;\r\n"
+            . "NOTE:This note folds onto a line reading\r\n"
+            . " END:VCARD\r\n"
+            . "EMAIL:first@example.com\r\n"
+            . "END:VCARD\r\n"
+            . "BEGIN:VCARD\r\n"
+            . "VERSION:3.0\r\n"
+            . "FN:Second Card\r\n"
+            . "N:Card;Second;;;\r\n"
+            . "END:VCARD\r\n";
+
+        $vcards = \rcube_vcard::import($input);
+
+        $this->assertCount(2, $vcards, 'Both vcards detected despite folded END:VCARD');
+        $this->assertSame('First Card', $vcards[0]->displayname, 'First card parsed fully');
+        $this->assertSame('Second Card', $vcards[1]->displayname, 'Second card still detected');
+
+        $assoc = $vcards[0]->get_assoc();
+        // The folded line is unfolded (RFC 2426 strips the single leading space).
+        $this->assertSame('This note folds onto a line readingEND:VCARD', $assoc['notes'][0], 'Folded NOTE value preserved');
+        $this->assertSame('first@example.com', $assoc['email:other'][0], 'EMAIL after folded line preserved');
+    }
 }


MiMoHo:
Some context to place this PR: #9637 already fixed the general line-folding handling ("bug 1") of #9593, and I'm aware of the longer-term plan to move vCard parsing to `sabre/vobject` (tracked separately). This is intentionally a small, self-contained interim fix for a *different*, still-present sub-case — a folded continuation line whose trimmed content happens to read `BEGIN:VCARD`/`END:VCARD`, which currently splits a card mid-import. It's fully covered by a regression test and can simply be superseded once the library migration lands. Happy to adjust or close if you'd rather hold all further parser changes for that work.

program/lib/Roundcube/rcube_vcard.php
```diff
@@ -562,9 +562,14 @@ public static function import($data)
...
 
+            // A folded/continuation line starts with whitespace (RFC 2426) and
+            // must never be mistaken for a block boundary. Trim only for the
+            // boundary comparison so surrounding whitespace does not break
+            // detection while the raw line above is preserved as content.
+            $is_folded = isset($line[0]) && ($line[0] === ' ' || $line[0] === "\t");
             $line = trim($line);
 
-            if (preg_match('/^END:VCARD$/i', $line)) {
+            if (!$is_folded && preg_match('/^END:VCARD$/i', $line)) {
```
alecpl: I'd prefer a `continue` before the trim(), so we skip the rest of the code in the loop if `$is_folded=false`. And remove the variable.

MiMoHo:
Good call — switched to a `continue` for continuation lines before the `trim()` and dropped the `$is_folded` variable. Behaviour is unchanged and the regression test still passes.

PR #10272, opened 2026-07-15
Elastic: Show contact photo in recipient autocomplete list

listumps:
## Summary
- Shows the contact's vCard photo instead of the generic person icon in the To/Cc/Bcc recipient autocomplete dropdown when composing, when a photo is available.
- The photo is already parsed on every autocomplete search (`rcube_contacts::search()` always reads the vCard), so this adds no extra query cost - `autocomplete.php` just exposes a URL to the existing `contacts/photo` action per matching contact.
- Photos stored as external `http(s)://` URLs are intentionally excluded: `contacts/photo.php` 302-redirects the browser to those, and attaching one here would fire a silent third-party request on every keystroke for any matching contact, rather than only when a contact/email is deliberately opened.
- Both `ksearch_results_display()` in `app.js` and Elastic's own override of it in `ui.js` (used for its small-screen layout) needed updating, since the skin's override completely replaces the base implementation.
- The photo is capped to the existing `.listing` row height (`@listing-line-height`) and keeps its natural aspect ratio rather than being cropped to a circle, consistent with how photos are shown elsewhere in the app (Contacts page, sender photo in message preview).
- Contacts without a photo, and group/distribution-list entries, keep the existing icon unchanged.

## Test plan
- [x] Verified via `php -l` and the project's `.php-cs-fixer.dist.php` config (0 fixable issues) for `autocomplete.php`
- [x] Verified via `eslint` (project config) for `app.js` and `ui.js`
- [x] Verified the `.less` change compiles cleanly via `lessc` against the full `styles.less` entry point
- [x] Manually tested end-to-end on a live instance: composing a message and searching for a contact with a vCard photo shows the photo in the dropdown; contacts without a photo and group entries still show the default icon
- [ ] Automated test coverage - none added; happy to add if maintainers want it

### DIFF
--- a/program/actions/mail/autocomplete.php
+++ b/program/actions/mail/autocomplete.php
@@ -88,6 +88,19 @@ class rcmail_action_mail_autocomplete extends rcmail_action
                                 // groups with defined email address will not be expanded to its members' addresses
                                 if ($contact['type'] == 'group') {
                                     $contact['email'] = $email;
+                                } elseif (!empty($record['photo'])) {
+                                    $photo = is_array($record['photo']) ? $record['photo'][0] : $record['photo'];
+
+                                    // skip photos stored as an external URL to avoid triggering
+                                    // an automatic, silent request to a third-party host on every keystroke
+                                    if (!empty($photo) && !preg_match('!^https?://!i', $photo)) {
+                                        $contact['photo'] = $rcmail->url([
+                                            '_task' => 'addressbook',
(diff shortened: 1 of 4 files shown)


listumps:
Removed the CHANGELOG.md entry from this PR, following the guidance given to @MiMoHo on another PR (changelog entries create unnecessary merge conflicts between PRs). Force-pushed the amended commit - PR now only touches the 4 code files.

PR #10282, opened 2026-07-19
Alter Postgres column cache_messages.uid from int to bigint

### LINKED ISSUE #10258: PostgreSQL schema does not confine UID values to the IMAP standard
When developing/extending Roundcube, I noticed that the PostgreSQL table schema does not support all values of message UID's that RFC 3501 allows.

Specifically, the [Unique Identifier Message Attribute](https://datatracker.ietf.org/doc/html/rfc3501#section-2.3.1.1) (UID) is represented as an non-zero unsigned 32-bit integer; integers in the range `[1, 4294967295]`

In Roundcube's database schema for [MySQL](https://github.com/roundcube/roundcubemail/blob/752eebab673657e25b2de64cd5913b2d189d6d99/SQL/mysql.initial.sql#L94), it is stored as an unsigned 32-bit integer. Great!

In the schema for [Postgres](https://github.com/roundcube/roundcubemail/blob/752eebab673657e25b2de64cd5913b2d189d6d99/SQL/postgres.initial.sql#L293), however, it's stored as an _signed_ 32-bit integer. This means that it can only support values up to `2147483647`; roughly half of the possible values.
I guess this only becomes an issue when a single mailbox approaches two billion messages, which is rare, but I think Roundcube should conform to the RFC standard.

(_I also don't know enough about SQLite to conclude wether it can store the full range or not_)

My proposal to resolve this is to change the datatype storing the UID's in the Postgres schema from `integer` to `bigint`.
Another method would be to shift the values by `-2147483648` before writing to the table, but that gets messy quickly.

Is this intentional behavior with a reason or a bug that should be resolved? Would like to hear some discussion before submitting a PR.

KonkenBonken:
Fixes #10258

This fixes a bug where the `uid` column on `cache_messages`, in [Postgres](https://github.com/roundcube/roundcubemail/blob/752eebab673657e25b2de64cd5913b2d189d6d99/SQL/postgres.initial.sql#L293), does not support the full range of possible values for message UID that [RFC 3501](https://datatracker.ietf.org/doc/html/rfc3501) allows.

Specifically, the [Unique Identifier Message Attribute](https://datatracker.ietf.org/doc/html/rfc3501#section-2.3.1.1) (UID) is represented as an non-zero unsigned 32-bit integer. The Postgres type `integer` is represented as a signed 32-bit integer. Opposed to MySQL, Postgres does not support any unsigned integer type.

To resolve this, we update the column type to store the UID as a `bigint` (signed 64-bit integer).


SQL/postgres.initial.sql
alecpl: You have to update this line too. We also usually create an empty update for other drivers.

KonkenBonken: I'll update it and add empty migrations for the other drivers!

Should the `system.roundcube-version` be updated for the other drivers or only postgreSQL?

KonkenBonken:
What's the matter with [Linux / PHP 8.4](https://github.com/roundcube/roundcubemail/actions/runs/32055794549/job/95465527791?pr=10282#logs) and [Linux / PHP 8.2](https://github.com/roundcube/roundcubemail/actions/runs/32055794576/job/95465528601?pr=10282#logs) failing it doesn't looks like it caused by any code error in this PR.

alecpl:
git.kolab.org where one dependency resides is often overloaded because of AI scrapers. Don't worry about it.

PR #10283, opened 2026-07-21
Fix extra <div> tags accumulating on 'Edit as New' for HTML messages

### LINKED ISSUE #9919: Extra <div> tags added when using "Edit as New" on HTML emails
### Prerequisites

### Describe the issue

When composing new HTML emails using the "Edit as New" function in RoundCube, extra <div> tags are added to the message body each time the email is re-edited and re-sent.

Steps to Reproduce:
1. Go to the "Sent" folder and select an HTML email.
2. From the "More" menu, choose "Edit as New".
3. Edit the message and send it.
4. Repeat steps 1–3 several times with the newly sent email.

Actual Result:
Each time the message is re-edited and sent, additional nested <div> tags are appended to the email content. This results in unnecessary HTML nesting in the message source.

Expected Result:
The content structure should remain clean and unchanged between re-edits. No additional <div> tags should be introduced.

Notes:
This issue only occurs with HTML format emails.

It does not occur if:
- A new email is created and the content is copied in manually.
- The email is edited using "Edit as New" but in Plain Text mode.
Attachments:
Two .txt files with HTML source from emails sent on June 25 and December 24. The December version shows significantly fewer <div> tags, indicating the buildup over repeated edits.

Image
Image
[HTML- june 2025.txt](https://github.com/user-attachments/files/21034832/HTML-.june.2025.txt)
[HTML-december 2024.txt](https://github.com/user-attachments/files/21034831/HTML-december.2024.txt)
Image

Chrome

8.3

1.6.11

### JavaScript errors

_No response_

### PHP errors

_No response_

dodjango:
## Summary
Fixes #9919 — repeatedly using "Edit as New" on an HTML message caused extra `<div>` wrappers to accumulate in the message body on every cycle. This turned out to have two independent causes, both fixed here:

1. `prepare_html_body()` only skipped the container-id wrapper (`<div id="editbody...">`) for `MODE_DRAFT`, not `MODE_EDIT` ("Edit as New"). Since Edit-as-New reloads/resends the message like a draft, the wrapper was saved and re-added on every cycle.
2. Setting `ignore_elements => ['body']` for `MODE_DRAFT`/`MODE_EDIT` had no effect, because `wash_html()` unconditionally registers a callback for the `body` tag, and a registered callback takes precedence over `ignore_elements` in `rcube_washtml::dumpHtml()`. That callback unconditionally wraps the body (and its `style` attribute, e.g. the `font-size`/`font-family` added by `send.php` on every send) in a new `<div>` — so the accumulation continued even after fix #1, just without an `id`. Added a `skip_washer_body_callback` option (mirroring the existing `skip_washer_form_callback`/`skip_washer_style_callback` pattern) and set it for `MODE_DRAFT`/`MODE_EDIT` so the body tag and its attributes are dropped entirely, keeping only its content.

## Test plan
- [x] Unit tests in `tests/Actions/Mail/ComposeTest.php` covering `MODE_EDIT`, `MODE_DRAFT` (both must not wrap the body, including when the stored `<body>` tag carries a style attribute), and regression coverage for `MODE_REPLY`/`MODE_FORWARD` (must still wrap in a container)
- [x] Full PHPUnit suite run against this branch; the only remaining failures/errors (`ImapTest::test_sort_folder_list`, `ImageTest`, `SearchTest`, `InstallTest::test_check_mime_extensions`) were confirmed to pre-exist identically on unpatched `upstream/master` in a side-by-side worktree comparison — no regressions introduced
- [x] Manual side-by-side comparison of two live instances (pre-fix vs. patched), repeating "Edit as New" + Send 3-4x on an HTML message: pre-fix accumulates nested `<div>` wrappers (with growing ids, then style-only once fix #1 alone was applied); patched instance shows no wrapper at all, regardless of cycle count
- [x] Manual regression check of adjacent compose flows on the patched instance: save draft → reopen (no wrapper, as intended), Reply (still wraps in `<div id="replybody...">`, unaffected), Forward (still wraps in `<div id="forwardbody...">`, unaffected)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This file includes only changes we consider noteworthy for users, admins and plu
 - Fix bug where the php session driver practically disabled session.lazy_write optimization (#9885, #10248)
 - Fix bug where dates could get displayed shifted back one day in some places (#9403)
 - Fix regression where it wasn't possible to hide a skin logo image anymore (#10254)
+- Fix extra `<div>` tags accumulating in the HTML body on every "Edit as New" cycle (#9919)
 
 ## Release 1.7.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This file includes only changes we consider noteworthy for users, admins and plu
 - Preserve the original message date on import of EML messages (#5559, #10251)
 - OAuth: Validate JWT token signature (#10210)
 - Installer: Add possibility to detsroy the installer session/cookie
+- Fix extra `<div>` tags accumulating in the HTML body on every "Edit as New" cycle (#9919)
 
 ## Release 1.7.4
 
--- a/program/actions/mail/compose.php
+++ b/program/actions/mail/compose.php
@@ -998,10 +998,14 @@ class rcmail_action_mail_compose extends rcmail_action_mail_index
             'add_comments' => false,
         ];
 
-        if (self::$COMPOSE['mode'] == rcmail_sendmail::MODE_DRAFT) {
+        if (self::$COMPOSE['mode'] == rcmail_sendmail::MODE_DRAFT
+            || self::$COMPOSE['mode'] == rcmail_sendmail::MODE_EDIT
+        ) {
             // convert TinyMCE's empty-line sequence (#1490463)
             $body = preg_replace('/<p>\xC2\xA0<\/p>/', '<p><br /></p>', $body);
             // remove <body> tags (not their content)
+            // Note: do not wrap the body in a container element here, otherwise
+            // an extra <div> would accumulate on every edit/send cycle (#9919)
             $wash_params['ignore_elements'] = ['body'];
         } else {
             $wash_params['container_id'] = $container_id;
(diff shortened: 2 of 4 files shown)


dodjango:
Friendly ping — this PR is ready for review whenever a maintainer has a moment. 🙏

Two updates since it was opened:

**Rebased onto current `master`** to clear a merge conflict that had appeared in the meantime. The only conflict was in `CHANGELOG.md`: since 1.7.3 has been released, I moved the entry from that (now released) section into the new `## Unreleased` section. `program/actions/mail/compose.php` merged cleanly — the code change itself is unchanged from its original state.

**Added two test cases** so the coverage matches what the test plan above describes: `MODE_DRAFT` (a saved draft shares the same no-container path as Edit-as-New and can be reopened repeatedly, so it is susceptible to the same buildup) and `MODE_FORWARD` (guards that the fix does not regress container-based style isolation for forwards). `ComposeTest` is at 7 tests / 15 assertions, all green.

A note on the previously red check: the single failing job (*Linux / PHP 8.3*, **Message Rendering** workflow) was a transient infrastructure failure unrelated to this change — `composer` got an `HTTP/2 504` from the GitHub API while downloading `nette/utils`. Everything substantive was green, and the full matrix has re-run since.

Happy to address any feedback. Thanks!

PR #10318, opened 2026-09-01
Drop two php-cs-fixer rules that a later duplicate key already overrides

darkdi:
`blank_line_before_statement` and `general_phpdoc_annotation_remove` are each set twice in `.php-cs-fixer.dist.php`. The first time with a configured list, the second time as plain `false`, under the "disable too destructive formating for now" and "TODO" comments. PHP keeps the last value for a repeated key, so both rules are off today and the two configured lists have never had any effect.

I removed the earlier configured entries, which leaves the effective configuration exactly as it is now. If the lists were the intent and the later `false` is the leftover, then the patch should go the other way round and the fixer will start touching files, so that is your call rather than mine.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -64,9 +64,6 @@ return (new Config())
         ],
         'native_function_invocation' => false,
         'void_return' => false,
-        'blank_line_before_statement' => [
-            'statements' => ['break', 'continue', 'declare', 'return', 'throw', 'exit'],
-        ],
         'final_internal_class' => false,
         'combine_consecutive_issets' => false,
         'combine_consecutive_unsets' => false,
@@ -78,9 +75,6 @@ return (new Config())
         'phpdoc_add_missing_param_annotation' => false,
         'return_assignment' => false,
         'comment_to_phpdoc' => false,
-        'general_phpdoc_annotation_remove' => [
-            'annotations' => ['author', 'copyright', 'throws'],
-        ],
 
         'modernize_strpos' => true,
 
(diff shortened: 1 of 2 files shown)


alecpl:
This reveals an issue with phpstan setup. It does not check files that start with a dot. We probably need to specify that file in the `paths` section of phpstan.neon.dist. Could you do that?

darkdi:
Done. Added both dot files to `paths`, not just the one, since `.php-cs-fixer-fully_qualified_strict_types.php` defines the custom fixer the other one requires and it was invisible for the same reason.

The cause is that phpstan hands directory scanning to Symfony Finder, which ignores dot files unless told otherwise. Your own cs-fixer config already works around it with `->ignoreDotFiles(false)`.

I could not run phpstan here, so CI will say whether the two newly visible files are clean at level 4.

darkdi:
Static Analysis went red on the second file, so I moved it to `scanFiles`.

Every error came from the custom fixer, which has to reach into php-cs-fixer internals to exist at all: extending `AbstractFixer`, the `ConfigurableFixerTrait`, `Tokens::getNamespaceDeclarations()` and so on. Analysing it reports each of those. Dropping it from `paths` instead would leave `.php-cs-fixer.dist.php` unable to resolve `CustomFullyQualifiedStrictTypesFixer`, so `scanFiles` is the middle: the symbols are known, the file is not analysed.

darkdi:
The remaining Windows / PHP 8.5 failure is unrelated to this branch: Composer timed out connecting to git.kolab.org:443 while cloning kolab/net_ldap3. All other Windows versions, every Linux job, Coding Style, and Static Analysis passed. I do not have permission to rerun the upstream job; could a maintainer rerun that failed job when convenient?

PR #10322, opened 2026-09-01
Fix bug where the refresh interval setting was displayed as "never" and could get overwritten on save for values outside the options list (#10292)

### LINKED ISSUE #10292: UI does not properly show `refresh_interval` from configuration
### Prerequisites

### Describe the issue

The default `refresh_interval` for users can be set in the global settings. [The defaults](https://github.com/roundcube/roundcubemail/blob/master/config/defaults.inc.php) describes that this value has the unit 'seconds'. No other restrictions are described there.

[In the settings UI for the user](https://github.com/roundcube/roundcubemail/blob/fa268f2269525a181abc729ec41ddbb63dc02630/program/actions/settings/index.php#L311), the current value is divided by 60 and used as the ID for the select.
However, the select has only the following IDs: 0 (never), 1, 3, 5, 10, 15, 30, 60 (all in minutes)

This mismatch is especially obvious when using something like 30 seconds in your global config. The resulting ID of `30 / 60 = 0.5` is mapped to the ID 0 and shown as `never`.

This issue was already raised in https://github.com/roundcube/roundcubemail/issues/7115 but I wouldn't see it as a feature request but rather as a UI bug because the currently active setting is not shown correctly. I am also totally fine with not being able to configure seconds in the frontend. However, it would be nice if the current setting could be displayed correctly, for example by adding the default value as an option in the dropdown.

Firefox

v8.4.23

v1.7.1

### JavaScript errors

_No response_

### PHP errors

_No response_

taylorsatula:
Fixes #10292

The select only offered 0 (never) and whole minutes [1,3,5,10,15,30,60], so values like 30 seconds or 2 minutes rendered as "never", and saving the form stored 0, wiping the setting. Make the select seconds-based (same pattern as draft_autosave in the same function), add the current interval as an extra option when it's not in the list, and store the posted value as-is. Adds an everynseconds label for sub-minute values.

### DIFF
--- a/program/localization/en_US/labels.inc
+++ b/program/localization/en_US/labels.inc
@@ -578,6 +578,7 @@ $labels['alwaysbutplain'] = 'always, except when replying to plain text';
 $labels['showinlineimages'] = 'Display attached images below the message';
 $labels['autosavedraft'] = 'Automatically save draft';
 $labels['everynminutes'] = 'every $n minute(s)';
+$labels['everynseconds'] = 'every $n second(s)';
 $labels['refreshinterval'] = 'Refresh (check for new messages, etc.)';
 $labels['never'] = 'never';
 $labels['immediately'] = 'immediately';
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This file includes only changes we consider noteworthy for users, admins and plu
 
 ## Unreleased
 
+- Fix bug where the refresh interval setting was displayed as "never" and could get overwritten on save for values outside the options list (#10292)
 - Preserve the original message date on import of EML messages (#5559, #10251)
 - OAuth: Validate JWT token signature (#10210)
 - Use `X-Content-Type-Options:nosniff` for attachment previews and downloads (#10308)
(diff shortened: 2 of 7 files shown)


alecpl:
Please, do not include Changelog entries, they will produce a conflict whenever something is added there.

taylorsatula:
> Please, do not include Changelog entries, they will produce a conflict whenever something is added there.

Sorry about that! I'll keep things out of changelog in future PRs. Do you want me to edit this one or can it be saved with a cherrypick?

program/localization/en_US/labels.inc
```diff
@@ -578,6 +578,7 @@ $labels['alwaysbutplain'] = 'always, except when replying to plain text';
 $labels['showinlineimages'] = 'Display attached images below the message';
 $labels['autosavedraft'] = 'Automatically save draft';
 $labels['everynminutes'] = 'every $n minute(s)';
+$labels['everynseconds'] = 'every $n second(s)';
```
alecpl, line 581: Setting the interval to 1 second really would not make much sense. So, this can be just "every $n seconds". Also, we probably should introduce some minimum value, e.g. 10 seconds.

